PR #3, opened 2020-07-22
docs: add API reference page using OpenAPI & redoc

rohanpm:
Use the OpenAPI spec produced by FastAPI to generate online docs.
All endpoints will be automatically detected and included here.
Right now, this includes only the single /healthcheck endpoint,
so the docs are almost entirely empty.

The API viewer uses redoc, so it's a standalone HTML file not
embedded within the sphinx-generated docs. (I tried also
sphinxcontrib-redoc and sphinxcontrib-openapi, but they don't
seem to support OpenAPI 3.x properly.)

Note that the openapi JSON file needs to be served by a real
web server in order for redoc to work correctly, so viewing
the docs locally needs something like:

    cd docs/_build/html
    python -m http.server 8001
    # then browse to http://localhost:8001

The layout of the application's modules was tweaked slightly to
better support accessing the application object from a script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,7 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+docs/openapi/*.json
 
 # PyBuilder
 target/
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -94,6 +94,7 @@ html_theme_options = {
     "github_banner": True,
     "description": "Publishing microservice for Red Hat's Content Delivery Network",
     "extra_nav_links": {
+        "API": "api.html",
         "Source": "https://github.com/release-engineering/exodus-gw",
         "Index": "genindex.html",
     },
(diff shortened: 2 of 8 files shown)


rohanpm:
Can be seen live at https://rohanpm.github.io/exodus-gw/api.html

lioramilbaum:
Something is not working for me. The 
```
(exodus-lambda) [lmilbaum@lmilbaum exodus-gw]$ cd docs/_build/html
bash: cd: docs/_build/html: No such file or directory
(exodus-lambda) [lmilbaum@lmilbaum exodus-gw]$ python -m http.server 8001
Serving HTTP on 0.0.0.0 port 8001 (http://0.0.0.0:8001/) ...
127.0.0.1 - - [22/Jul/2020 15:06:14] "GET / HTTP/1.1" 200 -
127.0.0.1 - - [22/Jul/2020 15:06:14] code 404, message File not found
127.0.0.1 - - [22/Jul/2020 15:06:14] "GET /favicon.ico HTTP/1.1" 404 -
```

Screenshot from 2020-07-22 15-08-42

rohanpm:
> Something is not working for me. Is there a preliminary step I should execute?
> 

Yeah, what I wrote in the message is just for viewing the docs assuming you've already generated them somehow.  Usually "tox -e docs" would be the way to do it.

PR #8, opened 2020-08-06
CODEOWNERS: add missing owner

rohanpm:
Ownership of this project was transferred to a new guild.
Ensure all the permanent guild members are marked as owners.


rohanpm:
I know we've seen some Travis quirks recently.  This PR is also in a weird state:

It has two Travis CI pull request checks, one passed and one "Build Created".

It doesn't seem to be a travis-ci .org vs .com issue, as both are for .com, and in fact if you click through to check both - they are both for exactly the same (passed) build.  It seems like this has to be a problem on Travis CI's end.

PR #11, opened 2020-08-25
Adjusting tox to run on RHEL7 CSB

lioramilbaum:


rohanpm (approved):
If this works for you and for CI, good, but I think in general having one test-requirements.txt with locked versions and hashes crossing multiple python versions is prone to issues.  If it works now, it might break later if the set of indirect dependencies changes based on the python version (requiring e.g. a test-requirements-py36.txt, test-requirements-py37.txt).  Of course, we can cross that bridge when we come to it.

PR #13, opened 2020-08-27
API end-point '/<env>/publish' created via FastAPI - request model

lioramilbaum:
/{env}/publish end point.

Run:
curl -X PUT http://localhost:8000/{env}/publish
Expected Result:
{"detail":"Created Publish Id"}

https://projects.engineering.redhat.com/browse/RHELDST-2774

Additional Information:
Changes to .pylintrc file were added due to an import-error. Following suggestion from the following KB:
https://stackoverflow.com/questions/1899436/pylint-unable-to-import-error-how-to-set-pythonpath

### DIFF
--- a/exodus_gw/gateway.py
+++ b/exodus_gw/gateway.py
@@ -1,7 +1,17 @@
 from .app import app
+from .publish import create_publish_id
 
 
 @app.get("/healthcheck")
 def healthcheck():
     """Returns a successful response if the service is running."""
     return {"detail": "exodus-gw is running"}
+
+
+@app.post("/{env}/publish")
+def publish(env: str):
+    """WIP: Returns a new, empty publish id"""
+    if env not in ["dev", "qa", "stage", "prod"]:
+        return {"error": "environment {0} not found".format(env)}
+    create_publish_id()
+    return {"detail": "Created Publish Id"}
--- a/exodus_gw/publish.py
+++ b/exodus_gw/publish.py
@@ -1,0 +1,7 @@
+import uuid
+
+
+def create_publish_id():
+    """Create a new publish id"""
+
+    return uuid.uuid4()
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -8,9 +8,10 @@ attrs==20.2.0 \
     --hash=sha256:26b54ddbbb9ee1d34d5d3668dd37d6cf74990ab23c828c2888dccdceee395594 \
     --hash=sha256:fce7fc47dfc976152e82d53ff92fa0407700c21acd20886a13777a0d20e655dc \
     # via -r test-requirements.in, pytest
-importlib-metadata==1.6.1 \
-    --hash=sha256:15ec6c0fd909e893e3a08b3a7c76ecb149122fb14b7efe1199ddd4c7c57ea958 \
-    # via -r test-requirements.in
+importlib-metadata==1.7.0 \
+    --hash=sha256:90bb658cdbbf6d1735b6341ce708fc7024a3e14e99ffdc5783edea9f9b077f83 \
+    --hash=sha256:dc15b2969b4ce36305c51eebe62d418ac7791e9a157911d58bfb1f9ccd8e2070 \
+    # via -r test-requirements.in, pluggy, pytest
 iniconfig==1.0.1 \
     --hash=sha256:80cf40c597eb564e86346103f609d74efce0f6b4d4f30ec8ce9e2c26411ba437 \
     --hash=sha256:e5f92f89355a67de0595932a6c6c02ab4afddc6fcdc0bfc5becd0d60884d3f69 \
@@ -24,9 +25,11 @@ more-itertools==8.5.0 \
     --hash=sha256:9b30f12df9393f0d28af9210ff8efe48d10c94f73e5daf886f10c4b0b0b4f03c \
     # via -r test-requirements.in, pytest
 packaging==20.4 \
+    --hash=sha256:4357f74f47b9c12db93624a82154e9b120fa8293699949152b22065d556079f8 \
     --hash=sha256:998416ba6962ae7fbd6596850b80e17859a5753ba17c32284f67bfff33784181 \
     # via -r test-requirements.in, pytest
 pluggy==0.13.1 \
+    --hash=sha256:15b2acde666561e1298d71b523007ed7364de07029219b604cf808bfa1c765b0 \
     --hash=sha256:966c145cd83c96502c3c3868f50408687b38434af77734af1e9ca461a4081d2d \
     # via -r test-requirements.in, pytest
 py==1.9.0 \
@@ -34,26 +37,34 @@ py==1.9.0 \
     --hash=sha256:9ca6883ce56b4e8da7e79ac18787889fa5206c79dcc67fb065376cd2fe03f342 \
     # via -r test-requirements.in, pytest
 pyparsing==2.4.7 \
+    --hash=sha256:c203ec8783bf771a155b207279b9bccb8dea02d8f0c9e5f8ead507bc3246ecc1 \
     --hash=sha256:ef9d7589ef3c200abe66653d3f1ab1033c3c419ae9b9bdb1240a85b024efc88b \
     # via -r test-requirements.in, packaging
 pytest-asyncio==0.14.0 \
     --hash=sha256:2eae1e34f6c68fc0a9dc12d4bea190483843ff4708d24277c41568d6b6044f1d \
     --hash=sha256:9882c0c6b24429449f5f969a5158b528f39bde47dc32e85b9f0403965017e700 \
     # via -r test-requirements.in
+pytest-mock==3.3.1 \
+    --hash=sha256:024e405ad382646318c4281948aadf6fe1135632bea9cc67366ea0c4098ef5f2 \
+    --hash=sha256:a4d6d37329e4a893e77d9ffa89e838dd2b45d5dc099984cf03c703ac8411bb82 \
+    # via -r test-requirements.in
 pytest==6.0.1 \
     --hash=sha256:85228d75db9f45e06e57ef9bf4429267f81ac7c0d742cc9ed63d09886a9fe6f4 \
     --hash=sha256:8b6007800c53fdacd5a5c192203f4e531eb2a1540ad9c752e052ec0f7143dbad \
-    # via -r test-requirements.in, pytest-asyncio
+    # via -r test-requirements.in, pytest-asyncio, pytest-mock
 six==1.15.0 \
+    --hash=sha256:30639c035cdb23534cd4aa2dd52c3bf48f06e5f4a941509c8bafd8ce11080259 \
     --hash=sha256:8b74bedcbbbaca38ff6d7491d76f2b06b3592611af620f8426e82dddb04a5ced \
     # via -r test-requirements.in, packaging
 toml==0.10.1 \
     --hash=sha256:926b612be1e5ce0634a2ca03470f95169cf16f939018233a670519cb4ac58b0f \
     --hash=sha256:bda89d5935c2eac546d648028b9901107a595863cb36bae0c73ac804a9b4ce88 \
     # via pytest
-wcwidth==0.2.4 \
-    --hash=sha256:79375666b9954d4a1a10739315816324c3e73110af9d0e102d906fdb0aec009f \
+wcwidth==0.2.5 \
+    --hash=sha256:beb4802a9cebb9144e99086eff703a642a13d6a0052920003a230f3294bbe784 \
+    --hash=sha256:c4d647b99872929fdb7bdcaa4fbe7f01413ed3d98077df798530e5b04f116c83 \
     # via -r test-requirements.in
 zipp==3.1.0 \
     --hash=sha256:aa36550ff0c0b7ef7fa639055d797116ee891440eac1a56f378e2d3179e0320b \
+    --hash=sha256:c599e4d75c98f6798c509911d08a22e6c021d074469042177c8c86fb92eefd96 \
     # via -r test-requirements.in, importlib-metadata
--- a/.pylintrc
+++ b/.pylintrc
@@ -101,3 +101,4 @@ disable=print-statement,
         # we use black code style and don't need additional checks
         bad-continuation,
         line-too-long,
+
--- a/test-requirements.in
+++ b/test-requirements.in
@@ -11,4 +11,5 @@ pytest
 six
 wcwidth
 zipp
+pytest-mock
 
--- a/tests/test_gateway.py
+++ b/tests/test_gateway.py
@@ -1,5 +1,30 @@
+import pytest
+
 from exodus_gw import gateway
 
 
 def test_healthcheck():
     assert gateway.healthcheck() == {"detail": "exodus-gw is running"}
+
+
+@pytest.mark.parametrize(
+    "env,expected",
+    [
+        ("dev", {"detail": "Created Publish Id"}),
+        ("qa", {"detail": "Created Publish Id"}),
+        (
+            "stage",
+            {"detail": "Created Publish Id"},
+        ),
+        (
+            "prod",
+            {"detail": "Created Publish Id"},
+        ),
+        (
+            "env_doesnt_exist",
+            {"error": "environment env_doesnt_exist not found"},
+        ),
+    ],
+)
+def test_publish(env, expected):
+    assert gateway.publish(env) == expected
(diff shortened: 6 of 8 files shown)


nathanegillett:
What's the intended scope of this PR?
What you have looks okay but I'm not seeing where the API would connect to or create publish objects in the database. Is that coming later?

lioramilbaum:
The scope of this PR is to act as the MVP of the feature.  I am breaking the overall story implementation to baby steps/PRs, such that each PR is effectively re-viewable by the whole team.

exodus_gw/gateway.py
```diff
@@ -1,7 +1,23 @@
...
 
 
 @app.get("/healthcheck")
 def healthcheck():
     """Returns a successful response if the service is running."""
     return {"detail": "exodus-gw is running"}
+
+
+class PublishObject(BaseModel):
+    artifacts: list
```
rohanpm, line 14: We will need to define the model for the artifacts.

exodus_gw/gateway.py
```diff
@@ -1,7 +1,23 @@
...
+
+
+@app.put("/{env}/publish")
+def publish(env: str, publishobject: PublishObject):
+    """Publish artifacts"""
+    if env not in ["dev", "qa", "stage", "prod"]:
```
rohanpm, line 20: I think these shouldn't be hardcoded - we'll need settings to define the details of each environment anyway, so those settings would also define the list of supported environments.  Maybe that should be done first?

exodus_gw/gateway.py
```diff
@@ -1,7 +1,23 @@
...
 def healthcheck():
     """Returns a successful response if the service is running."""
     return {"detail": "exodus-gw is running"}
+
+
+class PublishObject(BaseModel):
+    artifacts: list
+
+
+@app.put("/{env}/publish")
```
rohanpm, line 17: I don't think this endpoint is right, you need to add artifacts onto *a specific publish object* and not just send them to the service in general.

I would expect something more like:

```
@app.put("/{env}/publish/{publish_id}")
```

...meaning "add artifacts to the publish of the given ID".

It's also possible that env shouldn't be here because the service could already keep track of the environment on each publish object - not sure currently whether this is a good idea.  It might still make sense to keep env there anyway for a more consistent API.

exodus_gw/publish.py
```diff
@@ -0,0 +1,19 @@
+def publish_artifacts(env, artifacts):
+    """Publish all artifacts"""
+    for artifact in artifacts:
+        publish_artifact(env, artifact)
+    return {
+        "detail": "Publishing artifacts {0} in environment {1}".format(
```
rohanpm, line 6: I think we should keep these detail messages very terse, like just "publishing" in all successful cases and not have additional details on what was done.  Probably more detailed messages should be used only for errors.

The reasoning is that the messages are meant to be human-oriented, but if you put more detail in them there is a risk that someone starts parsing the messages to extract some details instead of using the API properly, essentially locking the format.

exodus_gw/gateway.py
```diff
@@ -1,7 +1,23 @@
...
+def publish(env: str, publishobject: PublishObject):
```
rohanpm, line 18: We'll need to define the response model.

exodus_gw/gateway.py
```diff
@@ -1,7 +1,23 @@
...
+    """Publish artifacts"""
```
rohanpm, line 19: API docs should be fleshed out at some point.

exodus_gw/gateway.py
nathanegillett: I think leaving the env for consistency is a good idea.

exodus_gw/gateway.py
lioramilbaum: Filed a sub-task to handle the model definition

exodus_gw/gateway.py
lioramilbaum: Where is the publish_id coming from? Do we have an end-point which produces publish_ids?

exodus_gw/gateway.py
lioramilbaum: Filed a sub-task to handle response model

exodus_gw/gateway.py
lioramilbaum: Filed a JIra issued to handle it

exodus_gw/gateway.py
lioramilbaum: Filed an issue to handle this. Those can be worked in parallel.

exodus_gw/gateway.py
```diff
@@ -1,7 +1,23 @@
...
+    if env not in ["dev", "qa", "stage", "prod"]:
+        return {"error": "environment {0} not found".format(env)}
```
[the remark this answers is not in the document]
lioramilbaum: This question will be addressed as soon as we define the response model.

exodus_gw/publish.py
lioramilbaum: +1 It should be addressed when dealing with the Response model

lioramilbaum:
The PR should not be merged till #15 is. Some of the code changes in this PR will be removed as soon as #15 is merged and this PR is rebased.

exodus_gw/gateway.py
nathanegillett: Generating the ID is one of the responsibilities of the publish method.
https://projects.engineering.redhat.com/browse/RHELDST-2159:
> The endpoint returns a unique "publish ID" for the created publish object (uuid.uuid4 recommended)

Also, issue 2158 doesn't include adding items/artifacts to the publish object in it's scope. That belongs to https://projects.engineering.redhat.com/browse/RHELDST-2160.

lioramilbaum: @nathanegillett Than publish_id is unknown when calling the end-point. @rohanpm suggested with use the publish_id at calling time.

lioramilbaum: > Also, issue 2158 doesn't include adding items/artifacts to the publish object in it's scope. That belongs to https://projects.engineering.redhat.com/browse/RHELDST-2160.

Does that mean that we should only create a publish_id with this end-point? I am a bit confused.

lioramilbaum:
> The corresponding issue for this PR is about creating the publish objects in the database. The publish endpoint should only be responsible for creating the publish database object and that object's ID. So, I don't think it's in scope to `publish_artifacts`; that's encroaching on https://projects.engineering.redhat.com/browse/RHELDST-2160, I believe.

I have read again the issue and figured out my confusion.  @nathanegillett thanks for that :-)

exodus_gw/gateway.py
nathanegillett: Yeah, @rohanpm is right, when adding items to a publish object, we'll need to provide the ID of that publish object so that the items go to the correct publish. Though, that shouldn't be in scope of 2159. 2159 is just about creating the publish object and returning that object's ID.

nathanegillett: So, when the publish endpoint is called (/env/publish), a publish object should be created within the database and we should have the ID of the object to use later.

exodus_gw/publish.py
```diff
@@ -0,0 +1,5 @@
+def create_publish_id(env):
+    """Create a new, empty pubplish id"""
+    return {
+        "detail": "Creating a new publish id in environment {0}".format(env)
+    }
```
nathanegillett, line 5: I don't think this needs to be handled by a separate Jira issue or anything. It should be as simple as using uuid4 to generate a random ID. That ID will then be assigned to the publish when it's created in the database.

lioramilbaum: I am not sure this is the right path. Don't we want to create a new row in a publish ids table and than use this id as a foreign key in the artifacts table?

nathanegillett: Something along those lines, yes. But we need to create the ID before we can use it, right? Or do you expect the DB to generate the ID itself?

exodus_gw/publish.py
```diff
@@ -0,0 +1,13 @@
+import uuid
+
+
+def create_publish_id(env):
+    """Create a new, empty publish id"""
+
+    pusblish_id = uuid.uuid4()
```
rohanpm, line 7:
```suggestion
    publish_id = uuid.uuid4()
```

exodus_gw/publish.py
```diff
@@ -0,0 +1,13 @@
...
+
+    return {
+        "detail": "Creating a new publish id {0} for environment {1}".format(
+            pusblish_id, env
```
rohanpm, line 11:
```suggestion
            publish_id, env
```

exodus_gw/publish.py
```diff
@@ -0,0 +1,13 @@
+import uuid
+
+
+def create_publish_id(env):
+    """Create a new, empty publish id"""
+
+    pusblish_id = uuid.uuid4()
+
+    return {
```
rohanpm, line 9: This doesn't make sense to me at all, why are we generating a publish ID and then only returning it embedded in a human-oriented message?  The point of generating an ID is to give it back to the caller, so that they can use it for further operations on the publish object.

exodus_gw/gateway.py
```diff
@@ -1,7 +1,17 @@
...
 
 @app.get("/healthcheck")
 def healthcheck():
     """Returns a successful response if the service is running."""
     return {"detail": "exodus-gw is running"}
+
+
+@app.post("/{env}/publish")
+def publish(env: str):
+    """Returns a new, empty publish id"""
```
rohanpm, line 13: In the current implementation it doesn't actually do this... the caller has no way of seeing the generated ID, which is really odd.

rohanpm (changes requested):

exodus_gw/gateway.py
lioramilbaum: As I wrote in a previous comment, I had a misunderstanding to what the scope of the issue. With that, the implementation was incorrect. @rohanpm and @nathanegillett your reviews helped me realize the confusion early. With that, there was a great value for this PR review.

exodus_gw/publish.py
lioramilbaum: The response object will be defined and implemented in a separate sub-task. The purpose of this message is to establish a return value and the corresponding tests.

exodus_gw/gateway.py
nathanegillett: I think his point is that the docstring should reflect the actual behavior of the function, regardless of our intentions for it.

test-requirements.txt
```diff
@@ -33,9 +33,13 @@ pytest-asyncio==0.14.0 \
...
     --hash=sha256:9882c0c6b24429449f5f969a5158b528f39bde47dc32e85b9f0403965017e700 \
     # via -r test-requirements.in
+pytest-mock==3.3.1 \
+    --hash=sha256:024e405ad382646318c4281948aadf6fe1135632bea9cc67366ea0c4098ef5f2 \
+    --hash=sha256:a4d6d37329e4a893e77d9ffa89e838dd2b45d5dc099984cf03c703ac8411bb82 \
+    # via -r test-requirements.in
 pytest==5.4.3 \
     --hash=sha256:5c0db86b698e8f170ba4582a492248919255fcd4c79b1ee64ace34301fb589a1 \
-    # via -r test-requirements.in, pytest-asyncio
+    # via -r test-requirements.in, pytest-asyncio, pytest-mock
```
nathanegillett, line 42: I don't really understand what this comment is support to communicate. Wasn't it just to say that the pytest hash was generated from test-requirements.in? If that's the case, why are pytest-asyncio and pytest-mocked noted here?
@rohampm @lioramilbaum not sure who this question should go to.

lioramilbaum: I can only guess that pytest is a dependency of both pytest-asyncio and pytest-mock. Maybe this is the reason why it is listed here.

exodus_gw/publish.py
nathanegillett: I guess it doesn't matter too much at this stage but, if we're only concerned with returning something for the tests right now, wouldn't it be simpler to just return the bare publish_id as @rohanpm suggests?

nathanegillett: Actually, I think in the end that's all this function should do, right? It only needs to return the publish_id to the caller, not some complex object. Or am I missing something?

exodus_gw/publish.py
```diff
@@ -0,0 +1,13 @@
...
+def create_publish_id(env):
+    """Create a new, empty publish id"""
+
+    publish_id = uuid.uuid4()
+
+    return {
+        "detail": "Creating a new publish id {0} for environment {1}".format(
+            publish_id, env
+        )
+    }
```
nathanegillett, line 13: I think create_publish_id should just return the bare publish_id. I don't see a reason to place it in some other object, as the goal is to just get the ID back to the caller. If there is a reason to return a more complex structure here, could you explain that a bit?

Testing this should be easier also without a dict and strings to dig through.

nathanegillett (changes requested):

exodus_gw/gateway.py
```diff
@@ -1,7 +1,17 @@
...
 
 @app.get("/healthcheck")
 def healthcheck():
     """Returns a successful response if the service is running."""
     return {"detail": "exodus-gw is running"}
+
+
+@app.post("/{env}/publish")
+def publish(env: str):
+    """WIP: Returns a new, empty publish id"""
```
nathanegillett, line 13:
```suggestion
    """WIP: Returns a new publish id"""
```

exodus_gw/publish.py
```diff
@@ -0,0 +1,9 @@
+import uuid
+
+
+def create_publish_id():
+    """Create a new, empty publish id"""
+
+    publish_id = uuid.uuid4()
+
+    return {"detail": publish_id}
```
nathanegillett, line 9: Sorry, I still don't understand the reason to return a dictionary.
This is just a backend function, right? So, it can just return `publish_id` to the caller, which, right now, is the publish endpoint.
```suggestion
    return publish_id
```
The caller can then expose to the user whatever details it needs to in the forthcoming response object.

nathanegillett (changes requested):

exodus_gw/publish.py
crungehottman: +1

exodus_gw/publish.py
lioramilbaum: The requirements from the response model were not collected yet. Not sure we need to block this PR over this.

exodus_gw/publish.py
lioramilbaum: See my previous comment

exodus_gw/publish.py
nathanegillett: This function shouldn't need to return a response object, was my point. Simply `return publish_id` is all that will ever be needed.

lioramilbaum: I am not sure of it and I would prefer not to over-engineer this. We will gain more understanding of what is needed as we drive this effort forward.

crungehottman: > I am not sure of it and I would prefer not to over-engineer this. We will gain more understanding of what is needed as we drive this effort forward.

I think that is an argument in favor of `return publish_id`. I hope we're not bikeshedding here :smile:, but @nathanegillett raises an important point. Wrapping the publish_id in a response object overcomplicates the function. When creating a foundation to build upon, we should opt for the simplest solution; if we don't have a specific use case or clear path forward for the more complicated option, why use it when a simpler option will suffice at this stage? If, in the future, we find that that the publish_id must be wrapped in a response object, we can make that change later.

lioramilbaum: We are defiantly bikeshedding and over-engineering here. I will do the change only because it potentially unblocks the approval process.

nathanegillett (approved):

crungehottman (approved):

exodus_gw/publish.py
rohanpm: I think the PR is structured in a way which encouraged bikeshedding because there is a bunch of code which produces values, returns them to nowhere and is not actually usable in any way.  It was and still is kind of unclear what the point is here as we have a function which has no useful side effect and whose return value is not used in any way.

Why even add the function now?  It obviously doesn't "work", so you must expect it to be evaluated by some criteria other than "does it work", but I don't think it is obvious to reviewers what that criteria should have been (certainly wasn't clear to me anyway).

I also just want this PR to be done with, but for later PRs, would it be possible to have some clearly defined scope (like "this fully implements X, while Y and Z are just placeholders", or just defining endpoint URLs and parameters while leaving implementation completely blank).  Or else could you next time provide some guidance on how reviewers should do their review?

rohanpm (approved):

PR #17, opened 2020-09-06
Fix pylint failure "raise-missing-from"

rohanpm:
pylint 2.6.0 released on 2020-08-20 added a new raise-missing-from
check, which fails on this code. It appears to suggest using an
explicit "from" whenever a chained exception is raised.

In this case, raising "from None" seems most appropriate, as we
are aiming to give the user a terse HTTP error without details of
any internal exceptions (which have already been logged).


lioramilbaum (approved):

lioramilbaum:
pylint version should be pinned to avoid such cases

PR #25, opened 2020-09-09
CI: pin versions of all testing tools & fix build

rohanpm:
Previously we had pinned versions of the test suite's
dependencies, but not additional testing tools such as black and
pylint. We should also pin the version of those, as otherwise new
versions of the tools can be released and break our CI without
warning.

Add all such tools to test-requirements.txt. Note that there is
a downside here that some requirements are being installed for
environments which don't need them. This could be fixed by
maintaining more requirements files (e.g. test-requirements-static.txt,
test-requirements-docs.txt, ...), but this doesn't seem worth the
complexity.

This change also has a couple of side-effects:

- allows tox.ini to be simplified, as there are no longer several
  different methods of installing dependencies

- fixes container image build, which was broken due to dataclasses and
  idna-ssl missing from requirements.txt; it did not show up in any of
  the tox envs previously because other dependencies listed in those envs
  happen to install these packages.

Note also that dataclasses must be explicitly listed as a dependency
only for Python versions less than 3.7, as it is not installable on
later versions of Python. This does not affect CI, but could impact
developers using newer versions of Python locally.


rohanpm:
I updated this after discovering that the first revision in c06d5d0 broke "tox -e docs" locally if using python >= 3.7.

rohanpm:
Updated this again to fix conflicts.

crungehottman (approved):

nathanegillett (approved):

PR #28, opened 2020-09-10
auth: add an endpoint for basic auth diagnosis

rohanpm:
Let's add a trivial endpoint which can be used to determine whether
the current user is authenticated, along with their roles. Without
this, there is no way to verify directly that the authentication
reverse-proxy is working correctly, or for a user to verify that
they were given the roles they requested, etc.

This also gives the healthcheck and whoami endpoints the same
openapi tag, so that they appear in the API docs together (as both
of them are kind of "meta APIs" relating to the service itself
rather than the exodus CDN.)

### DIFF
--- a/exodus_gw/gateway.py
+++ b/exodus_gw/gateway.py
@@ -1,8 +1,11 @@
+from fastapi import Depends
+
 from .app import app
 from .publish import create_publish_id
+from .auth import call_context, CallContext
 
 
-@app.get("/healthcheck")
+@app.get("/healthcheck", tags=["service"])
 def healthcheck():
     """Returns a successful response if the service is running."""
     return {"detail": "exodus-gw is running"}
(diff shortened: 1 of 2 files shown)


rohanpm:
Travis is acting a bit odd today.

exodus_gw/gateway.py
```diff
@@ -1,8 +1,11 @@
+from fastapi import Depends
+
 from .app import app
 from .publish import create_publish_id
+from .auth import call_context, CallContext
 
 
-@app.get("/healthcheck")
+@app.get("/healthcheck", tags=["service"])
```
lioramilbaum, line 8: This change is unrelated to this PR

tests/test_gateway.py
```diff
@@ -28,3 +28,10 @@ def test_healthcheck():
 )
 def test_publish(env, expected):
     assert gateway.publish(env) == expected
+
+
+def test_whoami():
```
lioramilbaum, line 33: I am not sure if this test covers all options: 
a request without any payload
a request with incorrect payload
a request with correct payload

lioramilbaum (changes requested):

lioramilbaum:
Is there an issue this PR can be linked to?

exodus_gw/gateway.py
nathanegillett: Tags could be added in an additional commit, at least.

tests/test_gateway.py
nathanegillett: If I understand `whoami` correctly, there's only one possible response; it just returns whatever context it's given.

exodus_gw/gateway.py
lioramilbaum: The value of the change is unquestionable. Mixing this effort with another effort is the issue here. The value of keeping PRs small and related to the PR subject+linked issue is for future trouble shooting/debugging.

nathanegillett: > Now in this change I'm adding one more endpoint which seems to belong in the same group, so I've put the same tag on the old endpoint and my new endpoint to group them.

Makes sense to me.

nathanegillett (approved):

nathanegillett:
It seems CI is still acting strange.

exodus_gw/gateway.py
rohanpm: > The value of keeping PRs small and related to the PR subject+linked issue is for future trouble shooting/debugging.

I agree with the principle of keeping PRs small and focused, but I don't think we agree on how to apply that principle.  I explained (already in the commit message) how I think this change *is* related to the purpose of the PR.

PR #29, opened 2020-09-11
PR CI should cover container image build [RHELDST-2973]

dichen16:
Add image build by using docker in travis-ci.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,14 @@ language: python
 
 install: pip install tox
 
+services: docker
+
 jobs:
   include:
+  - name: build docker image
+    script:
+    - docker --version
+    - docker build . -f openshift/containers/exodus-gw/Dockerfile
   - python: "3.6"
     env: TOX_ENV=cov-travis
   - python: "3.6"
@@ -13,7 +19,8 @@ jobs:
   - python: "3.6"
     env: TOX_ENV=docs
 
-script: tox -e $TOX_ENV
+script: 
+- tox -e $TOX_ENV
 
 after_success:
 - test $TOX_ENV = docs && scripts/push-docs


dichen16:
Can capture the build failure in travis-ci job log.
```
In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    idna-ssl>=1.0; python_version < "3.7" from https://files.pythonhosted.org/packages/46/03/07c4894aae38b0de52b52586b24bf189bb83e4ddabfe2e2c8f2419eec6f4/idna-ssl-1.1.0.tar.gz#sha256=a933e3bb13da54383f9e8f35dc4f9cb9eb9b3b78c6b36f311254d6d0d92c6c7c (from aiohttp==3.6.2->-r requirements.txt (line 15))
    dataclasses>=0.6; python_version < "3.7" from https://files.pythonhosted.org/packages/e1/d2/6f02df2616fd4016075f60157c7a0452b38d8f7938ae94343911e0fb0b09/dataclasses-0.7-py3-none-any.whl#sha256=3459118f7ede7c8bea0fe795bff7c6c2ce287d01dd226202f7c9ebc0610a7836 (from pydantic==1.6.1->-r requirements.txt (line 119))
The command '/bin/sh -c microdnf -y install shadow-utils     && microdnf -y install python3 python3-devel gcc make     && cd /usr/local/src/exodus-gw     && pip3 install --require-hashes -r requirements.txt     && pip3 install --no-deps .     && microdnf clean all && rm -rf /var/cache/yum && rm -rf /usr/local/src/exodus-gw' returned a non-zero code: 1
The command "docker build . -f openshift/containers/exodus-gw/Dockerfile" failed and exited with 1 during .
```

crungehottman:
@dichen16 - https://github.com/release-engineering/exodus-gw/pull/25 has been merged, which should fix the Travis CI failure

crungehottman:
Am I correct to assume that intent of this PR is to validate that the image can be built successfully, so we can more easily detect whether a change will break an image build?

dichen16:
@crungehottman yes, "validate that the image can be built successfully" exactly.

nathanegillett (approved):

rohanpm (changes requested):
I don't think this is implemented in the right way - if you have a look at the Travis logs, you can see that it builds the container image at the beginning of each of the three jobs in the test matrix, although as far as the build is concerned, there is no difference in environment between them.

It's at least wasteful, and also seems arbitrary to handle this one kind of test differently in this way.  Before the change, we had this:

- one job running python tests & collecting coverage
- one job building docs
- one job applying static checkers (e.g. pylint)

If we want to add another kind of test, should that just be adding one more job to the above list?  It seems more logical than adding an unrelated step at the beginning of all the existing jobs.

dichen16:
@rohanpm separated job for build-docker-image added.

.travis.yml
```diff
@@ -4,16 +4,23 @@ language: python
...
   - python: "3.6"
     env: TOX_ENV=docs
 
-script: tox -e $TOX_ENV
+script: 
+  - tox -e $TOX_ENV
 
 after_success:
-- test $TOX_ENV = docs && scripts/push-docs
+  - test $TOX_ENV = docs && scripts/push-docs
```
rohanpm, line 26: Not sure if this was intended to make the style consistent throughout the file. It doesn't - if you look under jobs.include, the '-' lines up with the parent 'include' (no indent) while here the '-' does not line up with the parent 'after_success' (there's an indent).

If you were going for a consistent style, there should either be no indent or always an indent whenever there's a list, but it doesn't matter to me.

.travis.yml
```diff
@@ -4,16 +4,23 @@ language: python
 
 install: pip install tox
 
+services:
+  - docker
+
 jobs:
   include:
+  - script:
```
rohanpm, line 12: I'm wondering is there any setting in .travis.yml which can give this job a meaningful name?

Right now, it seems to work correctly, but if you have a look at https://travis-ci.com/github/release-engineering/exodus-gw/builds/184137822, you can see the docker build job doesn't have any meaningful label - there is no way to know what that job is doing without clicking through to the log.  In contrast, it's clear what the other jobs are for due to the display of `TOX_ENV`.

rohanpm (approved):

rohanpm (approved):

lioramilbaum (changes requested):
Building the image and pushing it to right registry should be handled in a separate task before the CI tests are executed.

dichen16:
@lioramilbaum This PR introduces another image-building process besides the Quay.io Build Triggers.
1. Travis-CI build the image for validating that the image can be built successfully. This check happens at every PRs.
2. Quay.io build the image and store it in registry after a PR is merged(unit tests are all passed, and reviews are done).

So building-image **is** happened before CI tests. Not sure why pushing to registry should happened before CI tests as well.
PS. The CI tests above are unit tests.

lioramilbaum:
The long term plan is to enable integration tests in CI. Unit tests are not sufficient validation.
Will approve the PR as it is.

lioramilbaum (approved):

dichen16:
Yes, upstream CI is not enough, downstream CI for integration test running in openshift is necessary.

crungehottman (approved):

nathanegillett (approved):

PR #35, opened 2020-09-15
Make dependabot use Python 3.6 while resolving deps

rohanpm:
Dependabot is currently permitted to use the latest available
Python version. This doesn't work for us - we need pip-compile to
be run using Python 3.6, same as used in our container images,
otherwise the dependencies can be resolved incompatibly.

There is no dependabot-specific config file to control this. It
instead reads the configuration of various other python tools, and
there are several options for setting the version.
Adding it to pyproject.toml in this way seems the best option, as
we can have an explanatory comment and it does not block
development & testing of the project on later python versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,3 +10,13 @@ exclude = '''
   | \dist
 )/
 '''
+
+[tool.poetry.dev-dependencies]
+# The point of this setting is to force dependabot to use python 3.6
+# while updating our requirements files. This should match the python
+# version used in Travis CI and in container images.
+#
+# If we omit this, dependabot can use later python versions which can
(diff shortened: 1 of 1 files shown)


rohanpm:
This is intended to fix the broken PR seen in https://github.com/release-engineering/exodus-gw/pull/32

The dependabot behavior isn't documented in detail, so I cloned the dependabot repo and was testing it locally to check this.  I can confirm that this causes dependabot to use a python 3.6 environment when processing the project, but it is a bit difficult to get the environment set up to have dependabot run to completion locally, and I wasn't able to 100% confirm whether it'll fix the pull requests.

lioramilbaum (approved):

nathanegillett (approved):

PR #36, opened 2020-09-16
Get handle of the db engine

lioramilbaum:
https://projects.engineering.redhat.com/browse/RHELDST-3074

db settings (user, password, host, port) were added to the Settings class.


nathanegillett (changes requested):
There should be some basic unit tests included in this PR.
As for testing this change as a whole, you could just deploy it to exodus-gw-dev and try the endpoint as we've done in the past.

lioramilbaum:
> There should be some basic unit tests included in this PR.
> As for testing this change as a whole, you could just deploy it to exodus-gw-dev and try the endpoint as we've done in the past.

Thanks @nathanegillett for the review. I have quickly created this PR (which is still WIP) to demonstrate a point over the delivery pipeline document.

rohanpm:
@lioramilbaum would it make sense to change this PR to a draft until it's ready for a next round of review?

nathanegillett (approved):

lioramilbaum:
The PR is ready for review. Not sure if the test I've provided is enough. Do you have any other suggestions on which tests to implement?

nathanegillett:
> Not sure if the test I've provided is enough.

It's okay, I think. It ensures the function creates a database object, which is the expected outcome. I'm not sure what more you could test without connecting to a live database and that doesn't seem practical for a unit test, imo.

exodus_gw/gateway.py
```diff
@@ -13,7 +13,7 @@ def healthcheck():
 
 @app.post("/{env}/publish")
 def publish(env: str):
-    """WIP: Returns a new, empty publish id"""
+    """Returns a new, empty publish id"""
```
rohanpm, line 16: Removal of WIP here is odd as this commit doesn't seem to have any bearing on whether this function is a work in progress.  This is still not yet involving the DB in any way.

rohanpm (changes requested):

exodus_gw/app.py
```diff
@@ -1,3 +1,7 @@
 from fastapi import FastAPI
 
+from .database import get_db
+
+db = get_db()
```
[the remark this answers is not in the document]
lioramilbaum: > I just want to point out that no actual DB connection is established yet.
> 
> That's not a problem, in fact it's a good thing so that we can import all exodus_gw modules without having the dependency of a running DB server (e.g. docs publishing needs this).

What does that mean "we can import all exodus_gw modules"?  What action do we take to perform this action?
> 
> I point it out because it connects with a discussion elsewhere about how to test this change, but there is no connection to be tested yet, so the environment used for testing has no impact on it at all, e.g. this works even though I have not set up any exodus-gw-db host:
> 
> ```
> $ python
> Python 3.7.7 (default, Mar 11 2020, 21:18:58) 
> [GCC 10.0.1 20200216 (Red Hat 10.0.1-0.8)] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> from exodus_gw.app import db
> >>> db
> Engine(postgresql://None:***@exodus-gw-db:5432/exodus-gw)
> ```
Nice. I haven't thought of testing it like that. Will take a look of what is missing to establish a connection.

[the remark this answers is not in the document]
rohanpm: > What does that mean "we can import all exodus_gw modules"? What action do we take to perform this action?

I'm referring to at least these two cases where exodus_gw modules are imported outside the context of a running/deployed exodus-gw service:

- in the test suite, most tests are doing something like "import exodus_gw.module_under_test" and most of them depend on exodus_gw.app module.  Therefore, this `db = get_db()` is executed when pytest imports the tests and before there's any chance to put mocks in place.
- during build of the online docs (`tox -e docs`), the exodus_gw app is imported so that the openapi spec can be generated.  See scripts/gen-openapi.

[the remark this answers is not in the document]
lioramilbaum: Thanks for sharing the info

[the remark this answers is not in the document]
lioramilbaum: Now that I am thinking, it is a good think that the db session is not established yet. I will defer it to a later PR.

exodus_gw/database.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+from sqlalchemy import create_engine
+
+
+def get_db():
+    db_user = os.getenv("POSTGRES_USER")
```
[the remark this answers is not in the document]
lioramilbaum: I think it would be make more sense in setting those values in the ConfigMap. The deploy script would consume the value from the .env file. WDYT?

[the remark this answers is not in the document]
rohanpm: "add to the Settings class" and "setting those values in the ConfigMap" are compatible.  Anything you declare in the Settings class will be settable via ConfigMap, because ConfigMap is / can be used to set environment variables and Settings reads environment variables.

If you consider these two approaches for settings:

- Have ConfigMap set environment variables, use os.getenv in code to read them
- Have ConfigMap set environment variables, use Settings in code to read them

They are very very similar, Settings is just adding a little more structure on top, e.g. collecting all the env vars in one place rather than having them sprinkled around the code, and also giving a standard way to document them.  Under the hood it works the same and is fully compatible with ConfigMap.

> The deploy script would consume the value from the .env file

I think in the context of this PR, we only care about how the settings are read from the environment.  How they're deployed is a separate matter and we should avoid writing the code in a way which ties it to any particular deployment mechanism.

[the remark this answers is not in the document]
lioramilbaum: IIUC, the Settings class consumes the key/value items only from the ini file.  I don't see any value in storing the db attributes in the ini file. Do you?  
The key/value items can be injected during deployment and consumed by the service at run time.  We can expend the capability of the Settings file to consume additional items from the ConfigMap. Not sure it should be in the scope of this PR.

exodus_gw/settings.py
```diff
@@ -66,6 +72,11 @@ def get_settings() -> Settings:
             )
         )
 
+    settings.db_user = os.getenv("POSTGRES_USER", default="exodus-gw")
```
rohanpm, line 75: You shouldn't need any of these lines.  Settings class is already doing this for you, that's the point.  (Note: all the env vars are prefixed with `EXODUS_GW_`)

So, if you have set `EXODUS_GW_DB_USER=foo` in your environment, then you should already see that `settings.db_user == 'foo'` here, additional code to read from env vars is redundant as that's the point of `Settings` class in the first place.

exodus_gw/settings.py
```diff
@@ -29,6 +29,12 @@ class Settings(BaseSettings):
     environments: List[Environment] = []
     """List of environment objects derived from exodus-gw.ini"""
 
+    db_user: str = "exodus-gw"
+    db_pass: str = "exodus-gw"
+    db_host: str = "exodus-gw-db"
+    db_port: str = "5432"
+    """db settings"""
```
rohanpm, line 36: This isn't going to render properly in docs, each setting will pick up the doc string at the next line and that'll make it into our deployment guide.

So, right now this is going to end up with just db_port being documented as "db settings" and the rest being undocumented.

Could you please document them all individually or otherwise figure out a reasonable way for them to be rendered in docs?

exodus_gw/database.py
```diff
@@ -0,0 +1,12 @@
+from sqlalchemy import create_engine
+
+from .settings import get_db_settings
+
+
+def get_db():
+    SQLALCHEMY_DATABASE_URL = "postgresql://{0}:{1}@{2}:{3}/exodus-gw".format(
+        *get_db_settings()
+    )
+    print(SQLALCHEMY_DATABASE_URL)
```
rohanpm, line 10: Shouldn't have this print here.  If this would be useful sometimes, I suggest using a logger to log the message at DEBUG level.

exodus_gw/gateway.py
```diff
@@ -3,8 +3,8 @@
 from starlette.exceptions import HTTPException as StarletteHTTPException
```
rohanpm, line 3: This file doesn't have any functional changes in the PR, only style changes.  Can you please avoid including style changes on unrelated files?  Just make a separate PR for that if you feel this change is worthwhile.

exodus_gw/settings.py
```diff
@@ -83,3 +94,16 @@ def get_environment(env: str):
     raise HTTPException(
         status_code=404, detail="Invalid environment=%s" % repr(env)
     )
+
+
+def get_db_settings():
```
rohanpm, line 99: I would rather see the relevant code just use the Settings object directly instead of creating an unnamed tuple like this.

This way invites bugs, the order in which the caller uses the returned settings needs to match exactly the order they're returned here.  Also since the convention is for some of these to be equal in dev/test, e.g. db_user and db_pass both being "exodus-gw", a simple bug like swapping around the order of the fields could easily be missed all the way up to final deployments.

rohanpm (changes requested):

exodus_gw/database.py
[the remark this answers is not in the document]
rohanpm: > IIUC, the Settings class consumes the key/value items only from the ini file. 

This isn't right, it reads environment variables as mentioned above.

exodus_gw/settings.py
```diff
@@ -29,6 +29,15 @@ class Settings(BaseSettings):
...
     """List of environment objects derived from exodus-gw.ini"""
 
+    db_user: str = "exodus-gw"
+    """db user name"""
+    db_pass: str = "exodus-gw"
+    """db user password"""
+    db_host: str = "exodus-gw-db"
+    """db host"""
+    db_port: int = 5432
+    """db port"""
```
nathanegillett, line 39: These are fine default values but they should be configurable right? I'm not seeing how that's currently possible.

nathanegillett (changes requested):

exodus_gw/settings.py
```diff
@@ -83,3 +92,40 @@ def get_environment(env: str):
...
+
+
+def get_db_port():
+    """Return the db port"""
+
+    settings = get_settings()
+
+    print(settings.db_port, flush=True)
+    # db_port is set to tcp://<ip>:5432
+    # although the value in the ConfigMap is 5432
```
lioramilbaum, line 128: @rohanpm @nathanegillett Any idea why that happens?
Log file:
https://console-openshift-console.apps.ocp4.prod.psi.redhat.com/k8s/ns/exodus-gw-dev-3/pods/exodus-gw-8478cd747f-9r8k2/logs

nathanegillett: Nothing's immediately obvious to me. The call to get_settings() should set db_port to 5432. It might help to share a link to your playbook/role changes.

rohanpm: I think it's because the env var name EXODUS_GW_DB_PORT happens to clash with one already provided by the platform as default.

The platform already puts loads of environment variables into each pod by default based on which other pods and services exist.  I am surprised that the one from ConfigMap apparently doesn't override this though.

That behavior is briefly documented at https://kubernetes.io/docs/concepts/containers/container-environment/#cluster-information although actual behavior seems a lot more complex than just what's written there.

The conflict needs to be resolved somehow.

For reference, I checked an exodus-gw pod on another namespace and here are all the relevant environment variables which the platform is already setting by default:

```
EXODUS_GW_DB_PORT=tcp://172.30.139.78:5432
EXODUS_GW_DB_PORT_5432_TCP=tcp://172.30.139.78:5432
EXODUS_GW_DB_PORT_5432_TCP_ADDR=172.30.139.78
EXODUS_GW_DB_PORT_5432_TCP_PORT=5432
EXODUS_GW_DB_PORT_5432_TCP_PROTO=tcp
EXODUS_GW_DB_SERVICE_HOST=172.30.139.78
EXODUS_GW_DB_SERVICE_PORT=5432
EXODUS_GW_DB_SERVICE_PORT_EXODUS_GW_DB=5432
EXODUS_GW_PORT=tcp://172.30.143.111:8443
EXODUS_GW_PORT_8443_TCP=tcp://172.30.143.111:8443
EXODUS_GW_PORT_8443_TCP_ADDR=172.30.143.111
EXODUS_GW_PORT_8443_TCP_PORT=8443
EXODUS_GW_PORT_8443_TCP_PROTO=tcp
EXODUS_GW_SERVICE_HOST=172.30.143.111
EXODUS_GW_SERVICE_PORT=8443
EXODUS_GW_SERVICE_PORT_HTTPS=8443
```

And looking at that, I guess a simple option to fix this is just to rename these settings to `db_service_host`, `db_service_port`.  Then they will happen to match env vars with the correct values.

exodus_gw/database.py
```diff
@@ -0,0 +1,14 @@
+from sqlalchemy import create_engine
+
+from .settings import get_db_host, get_db_pass, get_db_port, get_db_user
+
+
+def get_db():
+    SQLALCHEMY_DATABASE_URL = "postgresql://{0}:{1}@{2}:{3}/exodus-gw".format(
```
rohanpm, line 7: Can the database name please be configurable as well?

exodus_gw/settings.py
```diff
@@ -83,3 +92,39 @@ def get_environment(env: str):
...
+
+def get_db_port():
+    """Return the db port"""
+
+    settings = get_settings()
+
+    print(settings.db_port, flush=True)
+    # db_port is set to tcp://172.30.88.248:5432
+    # although the value in the ConfigMap is 5432
+    print(settings.db_port, flush=True)
```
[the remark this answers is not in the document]
lioramilbaum: @nathanegillett Have you seen my comment before the print and here in the PR?

exodus_gw/settings.py
lioramilbaum: The key values can be overridden by environment variables. Take a look at the Ansible role MR.

exodus_gw/settings.py
lioramilbaum: Nice catch. Thanks.

exodus_gw/database.py
lioramilbaum: Yes. Will do. Although, The database URI might be different format for different vendors.

nathanegillett (approved):

exodus_gw/database.py
```diff
@@ -0,0 +1,20 @@
...
+    get_db_service_host,
+    get_db_service_pass,
+    get_db_service_port,
+    get_db_service_user,
+    get_db_name,
+)
+
+
+def get_db():
+    SQLALCHEMY_DATABASE_URL = "postgresql://{0}:{1}@{2}:{3}/{4}".format(
```
rohanpm, line 13: Hmm, why is this written in all caps?  It's not a global constant or anything like that.

exodus_gw/database.py
```diff
@@ -0,0 +1,20 @@
...
+        get_db_service_user(),
```
rohanpm, line 14: Not a blocker or anything, but I'm not really a fan of the way that a separate function was added for each piece of this.  Settings objects already store each of these in separate and appropriately-named attributes.  If there is an advantage to this extra layer of getters, it's not clear to me.

I'd probably have gone for something more like this (untested):

```
("postgresql://{s.db_service_user}:{s.db_service_pass}@{s.db_service_host}:"
 "{s.db_service_port}/{s.db_name}").format(s=settings)
```

rohanpm (approved):

exodus_gw/database.py
lioramilbaum: I've used the same variable name as in all the documentation I reviewed for this task. Feel free to suggest another name if you find it not suitable for our conventions.

exodus_gw/database.py
lioramilbaum: +1

PR #37, opened 2020-09-16
Implement environment support in upload API [RHELDST-2188]

nathanegillett:
This commit removes the 'bucket' parameter from the s3 upload API and
it's endpoints, replacing it with 'env'. The bucket name is now
generated to match the corresponding environment defined in
exodus-gw-config.ini.

### DIFF
--- a/docs/api-guide.rst
+++ b/docs/api-guide.rst
@@ -60,8 +60,9 @@ Usage:
                         verify='/path/to/bundle.pem',
                         config=Config(client_cert=('client.crt', 'client.key')))
 
+    # Bucket name must match one of the environments defined in exodus-gw.ini
+    bucket = s3.Bucket('dev')
     # Basic APIs such as upload_file now work as usual
-    bucket = s3.Bucket('exodus-cdn-dev')
     bucket.upload_file('/tmp/hello.txt',
                        'aec070645fe53ee3b3763059376134f058cc337247c978add178b6ccdfb0019f')
 
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -76,7 +76,7 @@ Settings
 
 To enable per-environment configuration of exodus-gw, exodus-gw.ini is available to point the
 application at specific AWS resources and declare the AWS profile to use when interacting with
-those resources. Each environment must appear in it's own section with the prefix "env.".
+those resources. Each environment must appear in its own section with the prefix "env.".
 
 .. code-block:: ini
 
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -67,9 +67,20 @@ Settings
 --------
 
 .. autoclass:: exodus_gw.settings.Settings()
-    :members:
+    :members: call_context_header
 
     exodus-gw may be configured by the following settings.
 
     Each settings value may be overridden using an environment variable of the
     same name, prefixed with ``EXODUS_GW_`` (example: ``EXODUS_GW_CALL_CONTEXT_HEADER``).
+
+To enable per-environment configuration of exodus-gw, exodus-gw.ini is available to point the
+application at specific AWS resources and declare the AWS profile to use when interacting with
+those resources. Each environment must appear in it's own section with the prefix "env.".
+
+.. code-block:: ini
+
+  [env.prod]
+  aws_profile = production
+  bucket = cdn-prod-s3
+  table = cdn-prod-db
--- a/exodus-gw.ini
+++ b/exodus-gw.ini
@@ -1,0 +1,14 @@
+[env.test]
+aws_profile = test
+bucket = my-bucket
+table = my-table
+
+[env.test2]
+aws_profile = test2
+bucket = my-bucket2
+table = my-table2
+
+[env.test3]
+aws_profile = test3
+bucket = my-bucket3
+table = my-table3
--- a/exodus_gw/s3/api.py
+++ b/exodus_gw/s3/api.py
@@ -37,6 +37,7 @@ import aioboto3
 from fastapi import Request, Response, Path, Query, HTTPException
 
 from ..app import app
+from ..settings import get_environment
 
 from .util import extract_mpu_parts, xml_response, RequestReader
 
@@ -47,31 +48,31 @@ LOG = logging.getLogger("s3")
 # - format of 'key' should be enforced (sha256sum)
 # - a way to check if object is already uploaded, e.g. HEAD
 # - limits on chunk sizes during multipart upload should be decided and enforced
-# - should support configuration of the target S3 environment rather than using
-#   whatever's the boto default
-# - {bucket} shouldn't be a parameter, as the target bucket for each exodus env
-#   is predefined and not controlled by the caller.
 # - requests should be authenticated
 
 
-def s3_client():
+def s3_client(profile: str):
     # Certain aspects of the boto client can be tweaked by environment variables
     # for development.
-    # This is expected to be replaced with a proper configuration system at some point.
-    return aioboto3.client(
-        "s3", endpoint_url=os.environ.get("EXODUS_GW_S3_ENDPOINT_URL") or None
+
+    # Note: Session creation will fail if provided profile cannot be found.
+    session = aioboto3.Session(profile_name=profile)
+
+    return session.client(
+        "s3",
+        endpoint_url=os.environ.get("EXODUS_GW_S3_ENDPOINT_URL") or None,
     )
 
 
 @app.post(
-    "/upload/{bucket}/{key}",
+    "/upload/{env}/{key}",
     tags=["upload"],
     summary="Create/complete multipart upload",
     response_class=Response,
 )
 async def multipart_upload(
     request: Request,
-    bucket: str = Path(..., description="S3 bucket name"),
+    env: str = Path(..., description="Target CDN environment"),
     key: str = Path(..., description="S3 object key"),
     uploadId: Optional[str] = Query(
         None,
@@ -99,7 +100,7 @@ async def multipart_upload(
     """Create or complete a multi-part upload.
 
     To create a multi-part upload:
-    - include ``uploads`` in query string, with no value (e.g. ``POST /upload/{bucket}/{key}?uploads``)
+    - include ``uploads`` in query string, with no value (e.g. ``POST /upload/{env}/{key}?uploads``)
     - see also: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html
 
     To complete a multi-part upload:
@@ -110,11 +111,11 @@ async def multipart_upload(
 
     if uploads == "":
         # Means a new upload is requested
-        return await create_multipart_upload(bucket, key)
+        return await create_multipart_upload(env, key)
 
     elif uploads is None and uploadId:
         # Given an existing upload to complete
-        return await complete_multipart_upload(bucket, key, uploadId, request)
+        return await complete_multipart_upload(env, key, uploadId, request)
 
     # Caller did something wrong
     raise HTTPException(
@@ -125,14 +126,14 @@ async def multipart_upload(
 
 
 @app.put(
-    "/upload/{bucket}/{key}",
+    "/upload/{env}/{key}",
     tags=["upload"],
     summary="Upload bytes",
     response_class=Response,
 )
 async def upload(
     request: Request,
-    bucket: str = Path(..., description="S3 bucket name"),
+    env: str = Path(..., description="Target CDN environment"),
     key: str = Path(..., description="S3 object key"),
     uploadId: Optional[str] = Query(
         None, description="ID of an existing multi-part upload."
@@ -157,19 +158,20 @@ async def upload(
 
     if uploadId is None and partNumber is None:
         # Single-part upload
-        return await object_put(bucket, key, request)
+        return await object_put(env, key, request)
 
     # Multipart upload
-    return await multipart_put(bucket, key, uploadId, partNumber, request)
+    return await multipart_put(env, key, uploadId, partNumber, request)
 
 
-async def object_put(bucket: str, key: str, request: Request):
+async def object_put(env: str, key: str, request: Request):
     # Single-part upload handler: entire object is written via one PUT.
     reader = RequestReader.get_reader(request)
+    env_obj = get_environment(env)
 
-    async with s3_client() as s3:
+    async with s3_client(profile=env_obj.aws_profile) as s3:
         response = await s3.put_object(
-            Bucket=bucket,
+            Bucket=env_obj.bucket,
             Key=key,
             Body=reader,
             ContentMD5=request.headers["Content-MD5"],
@@ -180,17 +182,17 @@ async def object_put(bucket: str, key: str, request: Request):
 
 
 async def complete_multipart_upload(
-    bucket: str, key: str, uploadId: str, request: Request
+    env: str, key: str, uploadId: str, request: Request
 ):
+    env_obj = get_environment(env)
     body = await request.body()
-
     parts = extract_mpu_parts(body)
 
     LOG.debug("completing mpu for parts %s", parts)
 
-    async with s3_client() as s3:
+    async with s3_client(profile=env_obj.aws_profile) as s3:
         response = await s3.complete_multipart_upload(
-            Bucket=bucket,
+            Bucket=env_obj.bucket,
             Key=key,
             UploadId=uploadId,
             MultipartUpload={"Parts": parts},
@@ -206,9 +208,13 @@ async def complete_multipart_upload(
     )
 
 
-async def create_multipart_upload(bucket: str, key: str):
-    async with s3_client() as s3:
-        response = await s3.create_multipart_upload(Bucket=bucket, Key=key)
+async def create_multipart_upload(env: str, key: str):
+    env_obj = get_environment(env)
+
+    async with s3_client(profile=env_obj.aws_profile) as s3:
+        response = await s3.create_multipart_upload(
+            Bucket=env_obj.bucket, Key=key
+        )
 
     return xml_response(
         "CreateMultipartUploadOutput",
@@ -219,14 +225,15 @@ async def create_multipart_upload(bucket: str, key: str):
 
 
 async def multipart_put(
-    bucket: str, key: str, uploadId: str, partNumber: int, request: Request
+    env: str, key: str, uploadId: str, partNumber: int, request: Request
 ):
     reader = RequestReader.get_reader(request)
+    env_obj = get_environment(env)
 
-    async with s3_client() as s3:
+    async with s3_client(profile=env_obj.aws_profile) as s3:
         response = await s3.upload_part(
             Body=reader,
-            Bucket=bucket,
+            Bucket=env_obj.bucket,
             Key=key,
             PartNumber=partNumber,
             UploadId=uploadId,
@@ -238,14 +245,14 @@ async def multipart_put(
 
 
 @app.delete(
-    "/upload/{bucket}/{key}",
+    "/upload/{env}/{key}",
     tags=["upload"],
     summary="Abort multipart upload",
     response_description="Empty response",
     response_class=Response,
 )
 async def abort_multipart_upload(
-    bucket: str = Path(..., description="S3 bucket name"),
+    env: str = Path(..., description="Target CDN environment"),
     key: str = Path(..., description="S3 object key"),
     uploadId: str = Query(..., description="ID of a multipart upload"),
 ):
@@ -258,9 +265,11 @@ async def abort_multipart_upload(
     """
     LOG.debug("Abort %s", uploadId)
 
-    async with s3_client() as s3:
+    env_obj = get_environment(env)
+
+    async with s3_client(profile=env_obj.aws_profile) as s3:
         await s3.abort_multipart_upload(
-            Bucket=bucket, Key=key, UploadId=uploadId
+            Bucket=env_obj.bucket, Key=key, UploadId=uploadId
         )
 
     return Response()
--- a/exodus_gw/settings.py
+++ b/exodus_gw/settings.py
@@ -1,6 +1,18 @@
+import os
+import configparser
+from typing import List
 from functools import lru_cache
 
 from pydantic import BaseSettings
+from fastapi import HTTPException
+
+
+class Environment(object):
+    def __init__(self, name, aws_profile, bucket, table):
+        self.name = name
+        self.aws_profile = aws_profile
+        self.bucket = bucket
+        self.table = table
 
 
 class Settings(BaseSettings):
@@ -13,6 +25,9 @@ class Settings(BaseSettings):
     """Name of the header from which to extract call context (for authentication
     and authorization).
     """
+
+    environments: List[Environment] = []
+    """List of environment objects derived from exodus-gw.ini"""
 
     class Config:
         env_prefix = "exodus_gw_"
@@ -28,4 +43,43 @@ def get_settings() -> Settings:
     afterward.
     """
 
-    return Settings()
+    settings = Settings()
+    config = configparser.ConfigParser()
+
+    config.read(
+        [
+            os.path.join(os.path.dirname(__file__), "../exodus-gw.ini"),
+            "/opt/app/config/exodus-gw.ini",
+        ]
+    )
+
+    for env in [sec for sec in config.sections() if sec.startswith("env.")]:
+        aws_profile = config.get(env, "aws_profile", fallback=None)
+        bucket = config.get(env, "bucket", fallback=None)
+        table = config.get(env, "table", fallback=None)
+        settings.environments.append(
+            Environment(
+                name=env.replace("env.", ""),
+                aws_profile=aws_profile,
+                bucket=bucket,
+                table=table,
+            )
+        )
+
+    return settings
+
+
+def get_environment(env: str):
+    """Return the corresponding environment object for the given environment
+    name.
+    """
+
+    settings = get_settings()
+
+    for env_obj in settings.environments:
+        if env_obj.name == env:
+            return env_obj
+
+    raise HTTPException(
+        status_code=404, detail="Invalid environment=%s" % repr(env)
+    )
--- a/tests/s3/test_manage_mpu.py
+++ b/tests/s3/test_manage_mpu.py
@@ -22,7 +22,7 @@ async def test_create_mpu(mock_s3_client):
 
     response = await multipart_upload(
         None,
-        bucket="my-bucket",
+        env="test",
         key=TEST_KEY,
         uploads="",
     )
@@ -80,7 +80,7 @@ async def test_complete_mpu(mock_s3_client):
 
     response = await multipart_upload(
         request=request,
-        bucket="my-bucket",
+        env="test",
         key=TEST_KEY,
         uploadId="my-better-upload",
         uploads=None,
@@ -123,7 +123,7 @@ async def test_bad_mpu_call():
     with pytest.raises(HTTPException) as exc_info:
         await multipart_upload(
             request=None,
-            bucket="my-bucket",
+            env="test",
             key=TEST_KEY,
             uploadId="oops",
             uploads="not valid to mix these args",
@@ -137,7 +137,7 @@ async def test_abort_mpu(mock_s3_client):
     """Aborting a multipart upload is correctly delegated to S3."""
 
     response = await abort_multipart_upload(
-        bucket="my-bucket",
+        env="test",
         key=TEST_KEY,
         uploadId="my-lame-upload",
     )
--- a/tests/s3/test_upload.py
+++ b/tests/s3/test_upload.py
@@ -10,15 +10,6 @@ from exodus_gw.s3.util import xml_response
 TEST_KEY = "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
 
 
-@pytest.fixture(autouse=True)
-def mock_s3_client():
-    with mock.patch("aioboto3.client") as mock_client:
-        s3_client = mock.AsyncMock()
-        s3_client.__aenter__.return_value = s3_client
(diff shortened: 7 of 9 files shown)


exodus_gw/s3/api.py
```diff
@@ -64,14 +60,14 @@ def s3_client():
 
 
 @app.post(
-    "/upload/{bucket}/{key}",
+    "/{env}/upload/{key}",
```
rohanpm, line 63: Does this break the compatibility with boto library?  If not, I guess at least the docs need updating for the new way of using it.

Here's the example currently of how you can use boto with exodus-gw: https://release-engineering.github.io/exodus-gw/api-guide.html#using-boto3-with-the-upload-api

The thing is that the S3 API always expects a bucket in the URL, so I was kinda expecting we would still have some bucket component in the URL, maybe just the literal value "bucket", and then our docs would say something like "if using exodus-gw via boto, always provide a bucket name of of "bucket" - the true bucket name will be determined via exodus-gw configuration and cannot be influenced by the caller".

As far as I can tell, this change might still be compatible with boto but you might need to do something like:

```
# e.g. if env is "qa"
s3 = boto3.resource('s3',
                    endpoint_url='https://exodus-gw.example.com/qa',
                    # If SSL needs to be configured:
                    verify='/path/to/bundle.pem',
                    config=Config(client_cert=('client.crt', 'client.key')))

# bucket name is always "upload" (to confirm with URL: <endpoint_url>/<bucket_name>/<key>)
bucket = s3.Bucket('upload')
```

It feels a bit weird.  Is one of these two options perhaps better:

- Make the URL /{env}/upload/bucket/{key}.  Then the endpoint_url is /{env}/upload, and the name given to s3.Bucket is always "bucket".
- Or make the URL /upload/{env}/{key}.  Then the endpoint_url is /upload, and the name given to s3.Bucket is the environment you want to use.  Which seems kinda nice, but then we need to consider if we would make other exodus-gw APIs consistent with that, e.g. /publish/{env}/... rather than /{env}/publish/...

exodus_gw/s3/util.py
```diff
@@ -94,3 +94,18 @@ def get_reader(cls, request):
         # a helper to make tests easier to write.
         # tests can patch over this to effectively disable streaming.
         return cls(request)
+
+
+def bucket_name(env: str):
```
rohanpm, line 99: Was this intended as the permanent mechanism?  I think it would ideally be something controllable from a config file rather than being locked into this convention.  Of course it would be possible to implement that later.

rohanpm (changes requested):
Basically fine, can we please tweak either the implementation or the docs to keep https://release-engineering.github.io/exodus-gw/api-guide.html#using-boto3-with-the-upload-api working?

exodus_gw/s3/util.py
rohanpm: Thinking about it a bit more, I don't think it should remain like this permanently as there could be differences between the buckets beyond their name.  e.g. the region or credentials.

lioramilbaum: +1. Bucket names should not be hard coded but pulled from the ConfigMap.

exodus_gw/s3/api.py
lioramilbaum: I am not following. What is the value of keeping consistency with s3 API? exodus-gw should provide a layer on top of the storage we use. Today it is S3, tomorrow it might be something else.

exodus_gw/s3/util.py
nathanegillett: > Thinking about it a bit more, I don't think it should remain like this permanently as there could be differences between the buckets beyond their name. e.g. the region or credentials.

This is specifically for the name but, sure, it doesn't need to be permanent.

> Bucket names should not be hard coded but pulled from the ConfigMap.

Bucket names aren't currently in the ConfigMap.

> I think it would ideally be something controllable from a config file rather than being locked into this convention.

It doesn't seem we currently use any configuration outside of ConfigMap.
Do we want to add more info there? Environment name, s3 bucket name, DynamoDB table name?

exodus_gw/s3/api.py
nathanegillett: So, if I understand correctly, the boto3 approach is an alternative to the upload API. However, we would need AWS account credentials to upload directly to the bucket using boto3, which almost no one will have, so we need the upload API to proxy for authentication.
Is that right, @rohanpm?

Of the three options suggested, I think the last is probably the best. While it doesn't seem necessary to change the publish endpoint path, I think consistency would be nice.
Wdyt, @lioramilbaum?

rohanpm: About the upload API being consistent with S3, there are two main points:

1. If we invent our own upload API, we must implement all the client logic ourselves.  If we instead follow an API which already exists, we gain access to a selection of existing client libraries.  But this is only an argument for following *some* existing API and not S3 specifically.
2. Upload is one of the more performance-sensitive parts, and if the exodus-gw API doesn't align with the API of the underlying storage, there's a risk of introducing major bottlenecks  Having the upload API follow the S3 API ensures this isn't a problem - if a certain task would need N requests when dealing with S3 directly, then it also requires only N requests when dealing with exodus-gw.  Also the request bodies can be streamed to S3 immediately without any buffering (e.g. if uploading a 10MB part, exodus-gw starts streaming the first bytes of that to S3 without waiting for the client to finish the entire part.).  It also works without having to store any state within exodus-gw during the upload (which is why the upload API is already fully functional without having a DB).

The storage backend indeed could change in the future, that's a valid concern which I considered while working on it, but if you look at the details of how the API works you'll see that the API being S3-compatible does not imply that the implementation is locked in to S3.  This is also evidenced by other projects like localstack.  Still, I think if the storage backend were to change, we nevertheless might also have to change the upload API due to performance reasons as described in point (2).  It's very challenging to make an API which wraps another API and is both abstracting the implementation perfectly, and also not introducing some performance overhead.

exodus_gw/s3/util.py
rohanpm: ConfigMap is just how the config is stored within Openshift, it doesn't tell us how the config should be loaded by exodus-gw which I think is more relevant here.

Config which we will need and don't currently have includes:

- the set of environments
- per-environment, the configuration for the boto library when doing uploads to S3; this includes everything which can be configured in boto (e.g. bucket name, credentials)

The ways that exodus-gw could retrieve this config include:

- environment variables, via Settings class
- a settings file
- loaded from DB

In my opinion, introducing a settings file seems to be the right choice here, an ini-style file which supports any of the config items used by boto library, plus a bit more.

If we choose environment variables or a settings file, ConfigMap can be used to store them.

exodus_gw/s3/api.py
nathanegillett: I've updated the endpoints to /upload/{env}/{key} and reflected the change in the api-guide doc.

exodus_gw/s3/util.py
nathanegillett: I've added a config file under root, a function to parse it in settings.py, and updated bucket_name function in s3/util.py to draw the name from there.

exodus_gw/s3/util.py
```diff
@@ -94,3 +95,14 @@ def get_reader(cls, request):
...
+    """Produces s3 bucket name for the given environment name."""
+
+    config = exodus_gw_config()
+
+    if env in config.sections() and "bucket" in config.options(env):
+        return config.get(env, "bucket")
+
+    raise ValueError("Environment '%s' not supported" % env)
```
rohanpm, line 108: I think this is just going to result in a 500 error if the caller requests a nonexistent env.

Could we please give a more appropriate status?  404 seems to be the correct value.  We can do that by `raise HTTPException(...)`.

exodus_gw/settings.py
```diff
@@ -1,6 +1,16 @@
 from functools import lru_cache
 
 from pydantic import BaseSettings
+import configparser
+
+
+def exodus_gw_config(file_name: str = "exodus-gw-config.ini"):
```
rohanpm, line 7: A few separate issues here:

1. Can we please name it just "exodus-gw.ini"?  I think the fact that it's a configuration file is already implied by ".ini".

2. This is loading the file from the current working directory only, that's probably not the best idea as we don't strongly control that.  e.g. can you tell without looking what the CWD is when the container is running in openshift?  In my experience, a convenient way to do it which is friendly both to devs and sysadmins is having multiple paths defined, one being relative to the code and one being hardcoded, and using the first existing.  e.g. if we look for `["<dirname(__file__)>/../exodus-gw.ini", "/opt/app/config/exodus-gw.ini"]` and document in our deployment guide that config will be loaded from /opt/app/config/exodus-gw.ini then we have a good way to handle it in both development and production.  By the way, /opt/app/config is fairly arbitrary but is consistent with platform-sidecar.

3. Now the biggest issue is that this doesn't fit with existing get_settings in any way:

- why does this one have exodus_gw in the name while get_settings doesn't?
- why does this one not have get_ prefix while get_settings does?
- why is get_settings cached via lru_cache while this one loads config again at every call?
- what is the difference between "config" and "settings" anyway - if I'm writing code to use one of these, how would I know which one to call?

I think these should probably be integrated into the same object.  From the point of view of code using the settings - there is no reason why settings should be arbitrarily split between a ConfigParser object and a Settings object.  That should ideally be an implementation detail which is wholly encapsulated within settings.py.

How about defining a model for these Environment settings, adding an `environments: List[Environment] = []` onto Settings, and adding a bit of code into `get_settings` to populate that list from the config file?

docs/api-guide.rst
```diff
@@ -61,7 +61,7 @@ Usage:
                         config=Config(client_cert=('client.crt', 'client.key')))
```
rohanpm, line 62: We should probably mention this:

```suggestion

    # Bucket name must match one of the environments defined in exodus-gw.ini.
```

tests/s3/test_manage_mpu.py
```diff
@@ -15,21 +15,21 @@ async def test_create_mpu(mock_s3_client):
     """Creating a multipart upload is delegated correctly to S3."""
 
     mock_s3_client.create_multipart_upload.return_value = {
-        "Bucket": "my-bucket",
+        "Bucket": "exodus-cdn-dev",
```
rohanpm, line 18: I don't think it's a good idea for all these tests to depend on the top-level exodus-gw.ini which would also be used by developers if running the service locally.

At least it seems a little dangerous & inconvenient if a dev wants to both run the service with real config locally (e.g. against exodus-cdn-dev) while also working with the test suite.  If you need to tweak settings for your dev server to manually test something, you can break the automated tests.  Also if the tests are exposed to the real details and credentials of target environments, there is increased risk that they might actually modify data in the target environment unexpectedly (as it's only the mocking setup which prevents this, which can be broken).

Could the test suite perhaps use a separate exodus-gw.ini?  The values could be consistent with the current behavior of tests to remove the need for (some of) these updates.

rohanpm (changes requested):

tests/s3/test_upload.py
```diff
@@ -114,3 +114,35 @@ async def test_part_upload(mock_s3_client, mock_request_reader):
 
     # It should have an empty body
     assert response.body == b""
+
+
+@pytest.mark.asyncio
+async def test_upload_invalid_env(mock_s3_client, mock_request_reader):
```
lioramilbaum, line 120: You might want to consider using fixtures for modularity. It doesn't have to implemented in this PR.
https://docs.pytest.org/en/stable/fixture.html

tests/s3/test_upload.py
```diff
@@ -114,3 +114,35 @@ async def test_part_upload(mock_s3_client, mock_request_reader):
...
+    """Uploading to an invalid environment does not delegate to s3."""
```
lioramilbaum, line 121: Another suggestion which is not a must but might make the code robust and scale-able: Parameterize test functions.
https://docs.pytest.org/en/stable/parametrize.html

tests/s3/test_manage_mpu.py
lioramilbaum: We have a ConfigMap object as a key/value store.  The `exodus-gw-config.ini` file is yet another key/value store. Consolidation of key/value stores into one might be a simpler solution.

exodus_gw/s3/util.py
lioramilbaum: I am not sure I am following the whole discussion here. But, I do think that all key/value pairs should be tracked in one place.

lioramilbaum (changes requested):

exodus-gw-config.ini
```diff
@@ -0,0 +1,19 @@
+# Configuration for the exodus-gw application
```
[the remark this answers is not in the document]
nathanegillett: So, this ini file is temporary? I thought you suggested this as the permanent solution.
> Having production-like config promotes misunderstanding how the config is managed.

How will the actual config for these values be handled? Without that this issue is incomplete, as the app won't be pointed to any real buckets.

exodus_gw/s3/api.py
```diff
@@ -47,10 +47,6 @@
 # - format of 'key' should be enforced (sha256sum)
 # - a way to check if object is already uploaded, e.g. HEAD
 # - limits on chunk sizes during multipart upload should be decided and enforced
-# - should support configuration of the target S3 environment rather than using
```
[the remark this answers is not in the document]
nathanegillett: > When deploying exodus-gw, it's the responsibility of the service owners to deploy ~/.aws/credentials with all profiles they want to use.

What do you mean by "deploy ~/.aws/credentials"? Where would they be deployed? I thought the plan was to store AWS auth in the playbook's secrets.

docs/api-guide.rst
nathanegillett: exodus-gw.ini won't have valid environments, though?
> Maybe it can just have some obviously phony environments like "test1", "test2" [...]

tests/s3/test_upload.py
```diff
@@ -89,15 +89,15 @@ async def test_part_upload(mock_s3_client, mock_request_reader):
...
+        env="dev",
         key=TEST_KEY,
         uploadId="my-best-upload",
         partNumber=88,
     )
 
     # It should delegate request to real S3
     mock_s3_client.upload_part.assert_called_once_with(
-        Bucket="my-bucket",
+        Bucket="exodus-cdn-dev",
```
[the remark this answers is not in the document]
nathanegillett: This is an assertion by the test so it should be hard-coded as the expected value.

tests/s3/test_upload.py
nathanegillett: I'm just using the existing infrastructure. If we want to improve that, it should be done in another PR.

tests/s3/test_upload.py
nathanegillett: I normally would but this test requires a whole new set of assertions.

tests/s3/test_manage_mpu.py
nathanegillett: We can have a separate ini file for tests but that introduces a bit more complexity; it'll probably require a lot of mocking to get tests pointed at the test config.

exodus-gw-config.ini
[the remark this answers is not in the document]
rohanpm: No, using an ini file wouldn't be temporary.  It's just that the exodus-gw repo isn't the definitive source for the specific config to be used in the production environment (and stage, qa etc environments).  The config file in this repo could be used for local development and/or for documentation purposes.

> How should the actual config for these values be handled?

The production (and stage, qa etc...) configuration should be deployed by ansible-roles/exodus-gw similar as was done for platform-sidecar.yml.

exodus_gw/s3/api.py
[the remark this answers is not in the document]
rohanpm: Yes they would be stored in the playbook's secrets and somehow deployed by ansible-roles/exodus-gw.

With respect to that "somehow", the main option which comes to mind is to put a file using the format of ~/.aws/credentials into a Secret and then mount that as a volume.  Other options could include using a Secret to provide environment variables, but I'm not sure if boto supports using environment variables to define multiple profiles.

docs/api-guide.rst
rohanpm: The one in this repo, yes.  The "exodus-gw.ini" here doesn't mean the one checked into this repo, it means the one you deploy with the service.

tests/s3/test_manage_mpu.py
rohanpm: > We have a ConfigMap object as a key/value store. The exodus-gw-config.ini file is yet another key/value store. Consolidation of key/value stores into one might be a simpler solution.

I don't think it would be.

With this config file, the entire file is likely to be deployed as one (key=exodus-gw.ini, value=<content of file>) within a ConfigMap.

It would certainly be possible to instead flatten all content of the file into e.g. (key=qa.table, value=exodus-cdn-qa), (key=stage.table, value=exodus-cdn-stage), ...

Then, to make the config usable to exodus-gw, we would either attach the ConfigMap pairs as environment variables (e.g. QA_TABLE=exodus-cdn-qa, STAGE_TABLE=exodus-cdn-stage, ...) or mount them as files (e.g. /config/qa.table contains "exodus-cdn-qa", /config/stage.table contains "exodus-cdn-stage", ...) as in example https://kubernetes.io/docs/concepts/configuration/configmap/ .

To me it's not simpler, it's about the same level of complexity except that some complexity is moved from exodus-gw code into openshift resources, where it's harder to test.  I think deploying the file in a configmap is fine, it seems to be a typical use-case as shown in above example.

exodus-gw-config.ini
```diff
@@ -0,0 +1,19 @@
...
+
+
+# qa currently shares storage with dev
+[qa]
```
rohanpm, line 5: I've just realized that these sections ought to have some kind of prefix, e.g. "env.qa" rather than just "qa".

The reason is that we may want to introduce sections not tied to a specific environment later.  `[logging]` would be a realistic example.  In such case, we wouldn't want exodus-gw to behave as though there is an environment named "logging".  Right now, it'd give you an error since "bucket" wouldn't be defined under logging, but ideally it would not even attempt to parse that section as an environment.

exodus-gw-config.ini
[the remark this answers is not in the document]
lioramilbaum: This can be enforced if we have a downstream repo which syncs from upstream repo (this one). The downstream repo can track additional changes which are specific to our production deployment.

exodus_gw/s3/api.py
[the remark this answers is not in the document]
lioramilbaum: What is the need which enforce support of multiple AWS profiles?

tests/s3/test_manage_mpu.py
lioramilbaum: I was thinking of have a one key per ConfigMap, e.g. key=table value=exodus-cdn-qa for a qa environment. The playbook is supposed to populate the right value according to the environment deployed.  The code should be amended to consume the value from the ConfigMap. Using a config file in parallel sounds like an hybrid solution between the Openshift way and what as before Openshift.
I don't know how to answer the statement that the suggested approach is harder to test.  It will become easier as soon as we gain more knowledge/experience with Openshift.

tests/s3/test_upload.py
lioramilbaum: +1

tests/s3/test_upload.py
lioramilbaum: +1

exodus-gw-config.ini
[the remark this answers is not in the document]
nathanegillett: > The config file in this repo could be used for local development and/or for documentation purposes.
[...]
The production (and stage, qa etc...) configuration should be deployed by ansible-roles/exodus-gw similar as was done for platform-sidecar.yml.

That makes a lot more sense. Thank you.

exodus_gw/s3/api.py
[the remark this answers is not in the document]
nathanegillett: > in the exodus-gw ini file, under each env, we define an 'aws_profile = xxx'. Then when constructing the client within exodus-gw, we request usage of that profile.

The best we can do here seems to be setting the `AWS_PROFILE` in the s3_client function. There's no profile parameter for `boto3.client` or even `botocore.config.Config`. I considered using the `Config` model instead but it doesn't seem to support key_id, key, and tokens. https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html

exodus-gw-config.ini
nathanegillett: Yeah, that makes sense. I would prefer to use nested sections but that isn't supported by configparser, unfortunately.

exodus_gw/settings.py
```diff
@@ -14,6 +26,8 @@ class Settings(BaseSettings):
     and authorization).
     """
 
+    environments: List[Environment] = []
```
rohanpm, line 29: Have you checked how this shows up in the generated docs? (Deployment guide documents this Settings class.)

I think we would want it to not appear there if possible, since it doesn't work the same way as the other attribute.

rohanpm (changes requested):

tests/s3/test_manage_mpu.py
rohanpm: > I was thinking of having one key per ConfigMap, e.g. key=table value=exodus-cdn-qa for a qa environment. The playbook is supposed to populate the right value according to the environment deployed.

I don't think that would work, remember one exodus-gw deployment supports multiple target environments, so there's more than one "table".

> I don't know how to answer the statement that the suggested approach is harder to test. It will become easier as soon as we gain more knowledge/experience with Openshift.

I don't see my opinion as related much to the level of experience.  Let me explain my thinking here:

For any settings we deal with, we ultimately want to be able to maintain them and make changes to them in a git repo, as is the convention for configuration management.  Those settings need to be made available to the bits of code in exodus-gw which depend on them.

So, how to get those settings maintained in a git repo connected with the exodus-gw code at runtime?  Well, the options which have been discussed here include:

- store them all directly as keys/values in a ConfigMap.  Our ansible needs changes relating to ConfigMap resource, and changes to ensure the right keys/values are attached to the pod which needs them, at the right place (most likely as env vars).  We need to make similar changes whenever we introduce a new key/value.  Within exodus-gw, the code for loading the settings would be near-trivial, as everything is loaded in the same way, e.g. by looking at os.environ.  So this is more complex and requires more work long-term on the ansible/openshift side, and less complex on the Python side.
- put them in a config file, and deploy that config file in a single key/value in a ConfigMap.  Our ansible needs a change to add one item to our ConfigMap, and connect it to our pod in one place, and there will not be any further changes needed to that structure as new config is added.  Within exodus-gw, the code for loading the config is more complex as the responsibility for somehow translating that flat file into the values needed by the code lives in Python maintained by us.  So this is less complex on the ansible/openshift side and more complex on the Python.

OK, so there is a choice of adding a bit of complexity either to our openshift resources managed by ansible, or in our Python code here.  Which is better?

Well let's compare the development experience between the two.  If you're working on a Python change then you've got:

- ability to write whatever unit tests you need to verify that the config file is parsed in the way you expect
- ability to run the entire test suite for the project in seconds as you iterate on your changes
- a code coverage report
- mature linter/formatter/IDE support, etc.

While if you're working on ansible/openshift changes you've got:

- virtually no ability to write unit tests at all
- no clear answer how to verify correctness of changes.  Note if there is an answer to this which involves running the ansible to deploy somewhere - that's probably going to be 100 times slower or more compared to running pytest.
- generally, less mature tools overall

If you've got to introduce a new setting for exodus-gw, which codebase would you rather work on between those two?  For me it would certainly be Python.

It's clear to me that we have much more effective means to manage complexity in the Python domain than we do in the ansible & openshift domain.  Therefore if there's a choice between increasing the complexity in Python or increasing the complexity in ansible/openshift, it makes more sense to favor the Python approach.

lioramilbaum: I appreciate the detailed feedback. You have raised few points which we might want to review.  The current media (GitHub) is ineffective (for me) to handle such a discussion. Will start an email thread for further discussion.

exodus_gw/s3/api.py
```diff
@@ -47,31 +48,32 @@
...
     # for development.
     # This is expected to be replaced with a proper configuration system at some point.
+
+    if profile:
+        os.environ["AWS_PROFILE"] = profile
+
     return aioboto3.client(
-        "s3", endpoint_url=os.environ.get("EXODUS_GW_S3_ENDPOINT_URL") or None
+        "s3",
+        endpoint_url=os.environ.get("EXODUS_GW_S3_ENDPOINT_URL", None),
```
lioramilbaum, line 64: We can consider storing this key/value in the ConfigMap. This is something to consider in a different PR.

exodus_gw/s3/api.py
```diff
@@ -47,31 +48,32 @@
...
 
 
-def s3_client():
+def s3_client(profile: str = ""):
     # Certain aspects of the boto client can be tweaked by environment variables
     # for development.
     # This is expected to be replaced with a proper configuration system at some point.
+
+    if profile:
+        os.environ["AWS_PROFILE"] = profile
```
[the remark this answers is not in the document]
nathanegillett: I'm surprised I failed to look at Session models given my work on chexus
https://github.com/nathanegillett/chexus/blob/eaa0eae66e02af0a2eb9a6b5e1a8c2903c36fd08/chexus/_impl/client/client.py#L62

I can't find docs for aioboto* but for boto3.Session it's `profile_name`.

exodus_gw/settings.py
nathanegillett: The only way to ensure this attribute isn't documented is to specify the Settings members we _do_ want documented.
```
.. autoclass:: exodus_gw.settings.Settings()
    :members: call_context_header
```
The autodoc docs suggest flagging the attribute as private will hide it but that proved ineffective.
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

tests/s3/test_upload.py
```diff
@@ -10,15 +10,6 @@
 TEST_KEY = "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
 
 
-@pytest.fixture(autouse=True)
-def mock_s3_client():
-    with mock.patch("aioboto3.client") as mock_client:
-        s3_client = mock.AsyncMock()
-        s3_client.__aenter__.return_value = s3_client
-        mock_client.return_value = s3_client
-        yield s3_client
```
nathanegillett, line 19: This was duplicated from conftest.py

docs/deployment.rst
```diff
@@ -67,7 +67,7 @@ Settings
 --------
 
 .. autoclass:: exodus_gw.settings.Settings()
-    :members:
+    :members: call_context_header
```
nathanegillett, line 70: This ensures that call_context_header is the only member of Settings documented

exodus-gw.ini
```diff
@@ -0,0 +1,14 @@
+[env.test]
```
rohanpm, line 1: Seems like there is still nothing about exodus-gw.ini in docs, do you intend to add something?  Having it in a later PR would also be fine.

exodus_gw/s3/api.py
```diff
@@ -47,31 +48,31 @@
...
-    # This is expected to be replaced with a proper configuration system at some point.
-    return aioboto3.client(
-        "s3", endpoint_url=os.environ.get("EXODUS_GW_S3_ENDPOINT_URL") or None
+
+    # Note: Session creation will fail if provided profile cannot be found.
+    session = aioboto3.Session(profile_name=profile)
+
+    return session.client(
+        "s3",
+        endpoint_url=os.environ.get("EXODUS_GW_S3_ENDPOINT_URL") or None,
```
rohanpm, line 63: I was curious just now whether this os.environ.get could also be dropped, as maybe this is something which can already be controlled by the "profile".

Unfortunately it looks like maybe not, at least I don't see it in the list of values at https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html .

rohanpm (approved):

tests/s3/test_manage_mpu.py
nathanegillett: I'm going to go ahead and resolve this, as the discussion (at least in part) has moved to email.

exodus-gw.ini
nathanegillett: I did add the following line to the api-guide;
`# Bucket name must match one of the environments defined in exodus-gw.ini`
but I can add some additional info on the ini file under deployment, settings.

exodus_gw/s3/api.py
nathanegillett: That's okay. It's only intended for testing/development cases anyway, right?

docs/deployment.rst
```diff
@@ -67,9 +67,20 @@ Settings
...
 
     exodus-gw may be configured by the following settings.
 
     Each settings value may be overridden using an environment variable of the
     same name, prefixed with ``EXODUS_GW_`` (example: ``EXODUS_GW_CALL_CONTEXT_HEADER``).
+
+To enable per-environment configuration of exodus-gw, exodus-gw.ini is available to point the
+application at specific AWS resources and declare the AWS profile to use when interacting with
+those resources. Each environment must appear in it's own section with the prefix "env.".
```
rohanpm, line 79:
```suggestion
those resources. Each environment must appear in its own section with the prefix "env.".
```

"it's" always means "it is" or "it has" rather than indicating ownership.

exodus_gw/s3/api.py
rohanpm: Yes, it was originally added so Tom could point this at localstack.  It just would have been nice if the profile could control it instead of having a special case for this thing.

It probably should have been migrated to `Settings` class at some point, but I don't think it should be in scope of this PR.

rohanpm (approved):

nathanegillett:
@crungehottman @lioramilbaum @dichen16, please have a look when you can.

crungehottman (approved):

PR #38, opened 2020-09-22
Make dependabot use Python 3.6 while resolving deps via runtime.txt

rohanpm:
This is a second attempt at making dependabot use Python 3.6 while
running pip-compile, since this is our target environment and using
the same Python version is necessary for updates to requirements
files to work properly.

runtime.txt originally came from Heroku, which we don't use, but
it allows dependabot to pick the correct Python version without
affecting anything else.  Compared to the previous approach of
using pyproject.toml, it has a disadvantage that the file format
doesn't allow for comments.

See commit c430043b6884 for the first attempt.


rohanpm:
This time I've done a full test of it in my fork, so I'm more confident it works.

A dependabot run making use of the change: https://github.com/rohanpm/exodus-gw/network/updates/63149084

In case the log is not viewable - we can see that it picked Python 3.6.9 as requested:

```
  proxy | 2020/09/22 00:24:10 [084] GET https://www.python.org:443/ftp/python/3.6.9/Python-3.6.9.tar.xz
  proxy | 2020/09/22 00:24:10 [084] 200 https://www.python.org:443/ftp/python/3.6.9/Python-3.6.9.tar.xz
```

And a resulting PR there for uvicorn avoids the problem which motivated this change: https://github.com/rohanpm/exodus-gw/pull/2  (note: the fact that the PR fails CI can be safely ignored, it's only because my own fork is not enabled for uploading data into coveralls).

lioramilbaum (approved):

crungehottman (approved):

nathanegillett (approved):

PR #50, opened 2020-10-05
examples: make it possible to get debug logs from s3-upload

rohanpm:
This is useful if the example does not work and you want to see
the requests/responses to figure out why.

### DIFF
--- a/examples/s3-upload
+++ b/examples/s3-upload
@@ -19,10 +19,12 @@
 # and larger than 10MB to cover both multipart and single part
 # uploads.
 
-import boto3
+import logging
 import sys
 import hashlib
 import argparse
+
+import boto3
 
 
 def get_object_key(filename):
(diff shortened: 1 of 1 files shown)


lioramilbaum (approved):
I would suggest to use Click instead of argparse.  Click is a newer library and probably additional features.

nathanegillett (changes requested):
I have to add client_cert and verify to the resource in order to make a connection. Should we add those as args also?

rohanpm:
> I have to add client_cert and verify to the resource in order to make a connection. Should we add those as args also?

This doesn't sound quite right:

- the upload APIs don't yet enforce any roles - so, while passing a client_cert should be possible, it should not be mandatory for any of those APIs yet (it will be later).
- If you have to pass "verify" then it probably means your python environment isn't linked up to your system's CA bundle correctly.  I think it should be handled as an environment issue rather than something every script needs an option for.  On my system, I have this in ~/.bash_profile to make the requests library behave better when installed to virtualenvs: `export REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt`

But overall - yeah, this should have support for passing a client cert at some point.  It'll be mandatory when the upload API starts enforcing any role.  I thought about including that already in this PR, but predicted it would be met with feedback "PR seems to be doing multiple changes, can you split it to smaller?" so decided not to do it for now.

nathanegillett (approved):

PR #51, opened 2020-10-07
deps: require uvicorn[standard] rather than just uvicorn

rohanpm:
The reason for this is, as mentioned in the uvicorn 0.12.0 changelog:

> Uvicorn no longer ships extra dependencies uvloop, websockets
> and httptools as default. To install these dependencies use
> uvicorn[standard].

If those deps are not installed, that would effectively mean that
upgrading to uvicorn 0.12.0 switches us from (uvloop & httptools)
backends to (asyncio & h11) backends.

uvicorn docs[1] imply that the former are superior, so let's tweak
the deps to ensure we keep using them.

[1] https://www.uvicorn.org/settings/

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -3,6 +3,6 @@ pydantic
 starlette
 aioboto3
 defusedxml
-uvicorn
+uvicorn[standard]
 gunicorn
 dataclasses;python_version<'3.7'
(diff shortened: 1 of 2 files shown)


rohanpm:
This PR was created in response to https://github.com/release-engineering/exodus-gw/pull/47.

lioramilbaum (approved):

nathanegillett (approved):

crungehottman (approved):

PR #54, opened 2020-10-12
Return XML error responses for upload API [RHELDST-3354] 

nathanegillett:
To get backtraces from the server when using the upload API, responses
must be in XML format or AWS libraries will fail to parse them.

This commit overrides the HTTPException handler for exodus_gw, allowing
us to generate custom responses. Now HTTPExceptions raised by the upload
API will return XML responses. HTTPExceptions raised by all other
endpoints will fallback to the original HTTPException handler.

### DIFF
--- a/exodus_gw/gateway.py
+++ b/exodus_gw/gateway.py
@@ -1,8 +1,26 @@
 from fastapi import Depends
+from fastapi.exception_handlers import http_exception_handler
+from starlette.exceptions import HTTPException as StarletteHTTPException
 
 from .app import app
 from .publish import create_publish_id
 from .auth import call_context, CallContext
+from .s3.util import xml_response
+
+
+@app.exception_handler(StarletteHTTPException)
+async def custom_http_exception_handler(request, exc):
+    # Override HTTPException to produce XML error responses for the
+    # given endpoints.
+
+    path = request.scope.get("path")
+
+    if path.startswith("/upload"):
+        return xml_response(
+            "Error", Code=exc.status_code, Message=exc.detail, Endpoint=path
+        )
+
+    return await http_exception_handler(request, exc)
 
 
 @app.get("/healthcheck", tags=["service"])
--- a/exodus_gw/s3/util.py
+++ b/exodus_gw/s3/util.py
@@ -53,13 +53,19 @@ def xml_response(operation: str, **kwargs) -> Response:
     """
     root = Element(operation)
 
+    status_code = kwargs.get("Code", 200)
+
     for (key, value) in kwargs.items():
         child = SubElement(root, key)
         child.text = str(value)
 
     xml = io.BytesIO()
     ElementTree(root).write(xml, encoding="UTF-8", xml_declaration=True)
-    return Response(content=xml.getvalue(), media_type="application/xml")
+    return Response(
+        content=xml.getvalue(),
+        status_code=status_code,
+        media_type="application/xml",
+    )
 
 
 class RequestReader:
--- a/tests/test_exception_handler.py
+++ b/tests/test_exception_handler.py
@@ -1,0 +1,30 @@
+from fastapi import HTTPException
+
+import pytest
+import mock
+
+from exodus_gw.gateway import custom_http_exception_handler
+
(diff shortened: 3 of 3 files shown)


exodus_gw/s3/api.py
```diff
@@ -51,6 +52,13 @@
 # - requests should be authenticated
 
 
+@app.exception_handler(StarletteHTTPException)
```
rohanpm, line 55: It seems like this will be applied onto all of the API endpoints - I'm not seeing a reason this would be restricted to the S3 APIs.  It should be, as we don't want to use that XML error response anywhere else.

Could you look into that and make sure it applies onto these APIs only?

lioramilbaum:
Can you please provide the reasoning for this change and also fix the broken CI.

nathanegillett:
I've moved the handler to the gateway module since its scope is includes the whole app.
It's designed with the expectation that forthcoming methods may need XML responses, e.g., committing publishes to DynamoDB.

exodus_gw/gateway.py
```diff
@@ -1,8 +1,28 @@
...
+    # given endpoints.
+
+    path = request.scope.get("path")
+
+    if path.startswith("/upload"):
+        return xml_response(
+            "Error", Code=exc.status_code, Message=exc.detail, Endpoint=path
+        )
+
+    return JSONResponse(
```
rohanpm, line 23: If we get here, instead of returning JSONResponse explicitly, could we please fall back to the default fastapi exception handlers?  I think we should only deviate from them where we have a specific reason for it.

As in example: https://fastapi.tiangolo.com/tutorial/handling-errors/#re-use-fastapis-exception-handlers

I think instead of returning JSONResponse we'd be doing:

```
from fastapi.exception_handlers import http_exception_handler
...
return await http_exception_handler(request, exc)
```

exodus_gw/gateway.py
```diff
@@ -1,8 +1,28 @@
...
+
+@app.exception_handler(StarletteHTTPException)
+async def http_exception_handler(request, exc):
+    # Override HTTPException to produce XML error responses for the
+    # given endpoints.
+
+    path = request.scope.get("path")
+
+    if path.startswith("/upload"):
+        return xml_response(
```
rohanpm, line 19: Just curious have you checked how boto library behaves now when receiving these error responses?

The acceptance criteria for the issue doesn't require identical behavior as the real S3.  However I do see one difference here which might be significant: seems like this will result in the HTTP status code (number) appearing in Code, as in `<Code>500</Code>`.  While in the real API, it returns strings such as `<Code>NoSuchKey</Code>`.  So, I wonder if the boto library is going to behave reasonably in that case (which is the goal of the issue).

exodus_gw/gateway.py
rohanpm: (...and then this custom method perhaps should not be named http_exception_handler.)

rohanpm (changes requested):

exodus_gw/gateway.py
nathanegillett: I did notice this but didn't think it'd be of consequence. I'll run a test to make sure.

exodus_gw/gateway.py
nathanegillett: Done!

exodus_gw/gateway.py
nathanegillett: So, this does not affect botocore behavior. We _could_ replace the specific HTTP code with the name of the exception but that will only ever be "HTTPError", which isn't very informative.
I'd prefer to either leave it as-is or change it to a name-code combination.

exodus_gw/s3/util.py
```diff
@@ -53,13 +53,19 @@ def xml_response(operation: str, **kwargs) -> Response:
...
     for (key, value) in kwargs.items():
         child = SubElement(root, key)
         child.text = str(value)
 
     xml = io.BytesIO()
     ElementTree(root).write(xml, encoding="UTF-8", xml_declaration=True)
-    return Response(content=xml.getvalue(), media_type="application/xml")
+    return Response(
+        content=xml.getvalue(),
+        status_code=status_code,
```
nathanegillett, line 66: This was necessary to raise the error, as it defaults to 200 otherwise.

rohanpm (approved):

PR #62, opened 2020-10-23
Make logging levels configurable [RHELDST-3456]

nathanegillett:
Adds loglevels section to exodus-gw.ini which is parsed and loaded by
the settings.get_settings function and applied upon app startup.

### DIFF
--- a/exodus_gw/settings.py
+++ b/exodus_gw/settings.py
@@ -26,8 +26,15 @@ class Settings(BaseSettings):
     and authorization).
     """
 
+    log_config: dict = {
+        "version": 1,
+        "incremental": True,
+        "disable_existing_loggers": False,
+    }
+    """Logging configuration in dictConfig schema."""
+
     environments: List[Environment] = []
-    """List of environment objects derived from exodus-gw.ini"""
+    """List of environment objects derived from exodus-gw.ini."""
 
     db_service_user: str = "exodus-gw"
     """db service user name"""
@@ -61,6 +68,15 @@ def get_settings() -> Settings:
             "/opt/app/config/exodus-gw.ini",
         ]
     )
+
+    for logger in config["loglevels"] if "loglevels" in config else []:
+        if "loggers" not in settings.log_config:
+            settings.log_config.update({"loggers": {}})
+
+        log_config = settings.log_config
+        dest = log_config if logger == "root" else log_config["loggers"]
+
+        dest.update({logger: {"level": config.get("loglevels", logger)}})
 
     for env in [sec for sec in config.sections() if sec.startswith("env.")]:
         aws_profile = config.get(env, "aws_profile", fallback=None)
--- a/exodus_gw/settings.py
+++ b/exodus_gw/settings.py
@@ -70,8 +70,7 @@ def get_settings() -> Settings:
     )
 
     for logger in config["loglevels"] if "loglevels" in config else []:
-        if "loggers" not in settings.log_config:
-            settings.log_config.update({"loggers": {}})
+        settings.log_config.setdefault("loggers", {})
 
         log_config = settings.log_config
         dest = log_config if logger == "root" else log_config["loggers"]
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -1,0 +1,19 @@
+import logging
+
+from exodus_gw.app import configure_loggers
+
+
+def test_log_levels():
+    """Ensure loggers are configured according to exodus-gw.ini."""
+
+    logging.getLogger("old-logger").setLevel("DEBUG")
+
+    configure_loggers()
+
+    # Should not alter existing loggers.
+    assert logging.getLogger("old-logger").level == 10
+
+    # Should set level of new loggers according to exodus-gw.ini.
+    assert logging.getLogger().level == 20
+    assert logging.getLogger("exodus-gw").level == 30
+    assert logging.getLogger("s3").level == 10
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -84,3 +84,14 @@ those resources. Each environment must appear in its own section with the prefix
   aws_profile = production
   bucket = cdn-prod-s3
   table = cdn-prod-db
+
+Logger levels may also be configured via exodus-gw.ini. Under a section named "loglevels",
+users may specify a logger name and the level at which to set said logger.
+
+.. code-block:: ini
+
+  [loglevels]
+  root = NOTSET
+  exodus-gw = INFO
+  s3 = DEBUG
+  ...
--- a/exodus-gw.ini
+++ b/exodus-gw.ini
@@ -12,3 +12,8 @@ table = my-table2
 aws_profile = test3
 bucket = my-bucket3
 table = my-table3
+
+[loglevels]
+root = INFO
+exodus-gw = WARNING
+s3 = DEBUG
--- a/exodus_gw/app.py
+++ b/exodus_gw/app.py
@@ -1,7 +1,15 @@
+import logging.config
 from fastapi import FastAPI
 
+from .settings import get_settings
 from .database import get_db
 
 db = get_db()
 
 app = FastAPI(title="exodus-gw")
+
+
+@app.on_event("startup")
+def configure_loggers():
+    settings = get_settings()
+    logging.config.dictConfig(settings.log_config)


exodus_gw/settings.py
```diff
@@ -53,6 +54,19 @@ def get_settings() -> Settings:
         ]
     )
 
+    if config.has_section("loglevels"):
+        log_config = {
+            "version": 1,
+            "incremental": "True",
+            "disable_existing_loggers": "False",
+            "loggers": {
```
rohanpm, line 62: Does this work for the root logger?  Based on my reading of dictConfig docs, it seems as though, if that logger needs any config, then it needs to be passed in a separate "root" item rather than in the "loggers" item along with the rest.

So I'd suspect that a small amount of code is needed here to specialize the case where key is "root".

exodus_gw/settings.py
```diff
@@ -53,6 +54,19 @@ def get_settings() -> Settings:
...
+                key: {"level": config["loglevels"].get(key)}
+                for key in config["loglevels"]
+            },
+        }
+
+        logging.config.dictConfig(log_config)
```
rohanpm, line 68: I don't think this should be happening directly within get_settings like this.

Previously, get_settings was a pure getter, it just returned a Settings object and didn't modify anything.

Now the behavior will be "get Settings and also change configuration of loggers" and it's undesirable to start mixing in side-effects like that.

Right now, the main impact would be on tests as many of them end up calling get_settings and each would now be adjusting loggers as they are called.

I think a better approach would be:

1. Make Settings contain the desired log settings (rather than *applying* them).  Could be a {loggername: level} dict or even just the dict in dictConfig format.

2. Use a startup event to apply the log config on server startup, per https://fastapi.tiangolo.com/advanced/events/

Which would look something like:

```
# note: not totally sure event handlers support Depends, actually.
# If not then just call get_settings within the function.
@app.on_event("startup")
def configure_loggers(settings: Settings = Depends(get_settings)):
    logging.config.dictConfig(settings.logging_config)
```

Could we please have it more like that instead?

rohanpm (changes requested):
One of the points in acceptance criteria was that the config should be documented at: https://release-engineering.github.io/exodus-gw/deployment.html

But I'm not seeing anything here which can achieve that.  Could we please have a docs update included?

nathanegillett:
> Is it desirable to have the logging settings hard coded? What if we want to use different settings for different environment?
> In that case, we would like to pull the settings from environment variables which are pulled from the ConfigMap.

We're currently using exodus-gw.ini for environment-specific settings. So, if environment-specific logging config _is_ desired (wasn't mentioned in the Jira issue), I'd imagine we'd want to figure that out there.

exodus_gw/settings.py
```diff
@@ -53,6 +56,21 @@ def get_settings() -> Settings:
         ]
     )
 
+    if config.has_section("loglevels"):
+        settings.log_config = {
+            "version": 1,
+            "root": {
+                key: {"level": config["loglevels"].get(key)}
```
rohanpm, line 63: Is this really the right format?  Seems there will be two levels of "root" here, e.g.

```
{"version": 1, "root": {"root": {"level": "INFO"}}}.
```

tests/test_logging.py
```diff
@@ -0,0 +1,14 @@
...
+
+from exodus_gw.app import configure_loggers
+
+
+def test_log_levels():
+    """Ensure loggers are configured according to exodus-gw.ini"""
+    configure_loggers()
+
+    # Should set logging level for root to WARNING (30)
+    assert logging.getLogger().isEnabledFor(30)
```
rohanpm, line 11: This happens to be the default level for the root logger, so this isn't really testing that it sets the log level - the test would also pass if the root level isn't touched.

I suspect that is indeed what's happening here (see my earlier comment about whether the structure is correct right now for root logger).  Could you please check, and if I was right that the dictConfig structure for root logger right now is incorrect, could you ensure this test covers it somehow?

In general for all tests like this it's a good idea to keep in mind what the default values are, and to never use them as the expected values in tests, otherwise there will always be some doubt - do the tests pass because the code works, or because the code isn't doing anything.

rohanpm (changes requested):

rohanpm:
> Is it desirable to have the logging settings hard coded?

They aren't hard-coded, they're in a settings file.  The settings file *in this repo* is just for development, testing & demonstrative purposes.  The settings file for real deployments is deployed by ansible into a ConfigMap.

> What if we want to use different settings for different environment?

It's fully supported, you would have different content for the settings files in different hosts within your ansible inventory.

> In that case, we would like to pull the settings from environment variables
> which are pulled from the ConfigMap.

Replace "environment variable" with "file" (the other main way of using a ConfigMap) and that's exactly how it works.

lioramilbaum:
> > Is it desirable to have the logging settings hard coded?
> 
> They aren't hard-coded, they're in a settings file. The settings file _in this repo_ is just for development, testing & demonstrative purposes. The settings file for real deployments is deployed by ansible into a ConfigMap.
> 
> > What if we want to use different settings for different environment?
> 
> It's fully supported, you would have different content for the settings files in different hosts within your ansible inventory.

How that would work for the dev-X environments?

> 
> > In that case, we would like to pull the settings from environment variables
> > which are pulled from the ConfigMap.
> 
> Replace "environment variable" with "file" (the other main way of using a ConfigMap) and that's exactly how it works.
file is tracked in git. Overriding the value requires changing the file, even though the change should not land in the head repo.

nathanegillett:
> > It's fully supported, you would have different content for the settings files in different hosts within your ansible inventory.
> 
> How that would work for the dev-X environments?

The config is defined in group_vars under 'exodus-gw-playbooks/inventory'. So, for our developer namespaces, one easy way would be to just directly edit `exodus_gw_config` in 'exodus_gw_dev.yml', since 'dev' is the environment used by the deployment helper script. Another way would be to provide `exodus_gw_config` to the ansible-playbook extra vars.

> > Replace "environment variable" with "file" (the other main way of using a ConfigMap) and that's exactly how it works.
> 
> file is tracked in git. Overriding the value requires changing the file, even though the change should not land in the head repo.

If you were to opt for directly editing the group_vars you wouldn't need to commit the change to deploy it since it's on the playbook side of things.

tests/test_logging.py
nathanegillett: Good catch!

exodus_gw/settings.py
nathanegillett: That is the current schema, as `key` in this case is "root".

> {"version": 1, "root": {key: {"level": config["loglevels"].get(key)}, "loggers": {...}, ...}

The above would translate to something like...

> {"version": 1, "root": {"root": {"level": "INFO"}, "loggers": {...}, ...}

Nonetheless, this isn't working to set the root logger level so there must be something missing in the schema (examples of which have proven hard to find).

nathanegillett: Yep, you're right. Removed duplicate.

exodus_gw/settings.py
```diff
@@ -26,6 +26,9 @@ class Settings(BaseSettings):
     and authorization).
     """
 
+    log_config: dict = {}
```
rohanpm, line 29: This doesn't really matter but I'm curious if this happens to "just work" with environment variables as well.

Per https://pydantic-docs.helpmanual.io/usage/settings/#parsing-environment-variable-values, for a setting defined as a dict, it can be parsed from a JSON-formatted environment variable.

So, I would guess that if you don't put loglevels in the config file, but instead you define EXODUS_GW_LOG_CONFIG="{... the dictconfig structure}", it's going to load that dict and work just fine although it was not the goal of this issue.

I think that's kinda neat, as although we've guessed that only log levels are relevant, seems we have an escape hatch which can allow us to override anything else about loggers without further code changes - if the need arises.

rohanpm (approved):

crungehottman (approved):

exodus_gw/settings.py
nathanegillett: From hacking the tests a bit, this doesn't seem to work right now.

It _should_ work in the scenario you describe -- the environment variable is set prior to call of `get_settings` and no "loglevels" section is present in exodus-gw.ini. However, with `settings.log_config` defaulting to an empty dictionary, `logging.config.dictConfig` will raise an exception about a missing version number.

> ValueError: dictionary doesn't specify a version


I think this is a bug, as calling `configure_loggers` without a "loglevels" section in exodus-gw.ini will result in an error rather than loading default logging configuration.

I'll update this commit with the addition of that missing key-value in `log_config`'s default dict.

tests/test_logging.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_log_levels():
+    """Ensure loggers are configured according to exodus-gw.ini"""
+    configure_loggers()
+
+    # Should set logging levels to INFO (20)
+    assert logging.getLogger().isEnabledFor(20)
+    assert logging.getLogger("exodus-gw").isEnabledFor(20)
+    assert logging.getLogger("s3").isEnabledFor(20)
```
rohanpm, line 13: Per my comments elsewhere about problems with this:

Could we please have the test updated (or a new test added) which verifies the impact on a logger which is NOT listed in the config?  i.e. the expected behavior is not only to set log levels on those loggers in the config, but also to leave all other loggers unmodified.

rohanpm (changes requested):

exodus_gw/settings.py
```diff
@@ -62,6 +69,14 @@ def get_settings() -> Settings:
         ]
     )
 
+    for logger in config["loglevels"] if "loglevels" in config else []:
+        settings.log_config.update({"loggers": {}})
```
rohanpm, line 73: This is wrong, every iteration of the loop will update log_config["loggers"] back to an empty dict and therefore wipe out config for the loggers from earlier iterations (except for root).

If you were expecting update() to do a kind of "deep merge" of dicts - it doesn't work that way.

Added a comment in the test as well regarding why the test manages to pass despite this bug.

tests/test_logging.py
```diff
@@ -0,0 +1,19 @@
...
+    logging.getLogger("old-logger").setLevel("DEBUG")
+
+    configure_loggers()
+
+    # Should not alter existing loggers.
+    assert logging.getLogger("old-logger").getEffectiveLevel() == 10
+
+    # Should set level of new loggers to INFO (20).
+    assert logging.getLogger().getEffectiveLevel() == 20
+    assert logging.getLogger("exodus-gw").getEffectiveLevel() == 20
```
rohanpm, line 18: About the bug I mentioned above, if you dump the log_config dict which is being applied you can see that exodus-gw is wrongly missing:

```
{'disable_existing_loggers': False,
 'incremental': True,
 'loggers': {'s3': {'level': 'INFO'}},
 'root': {'level': 'INFO'},
 'version': 1}
```

But this test still passes because:

- you're testing getEffectiveLevel here, and the "effective level" is inherited from the parent (root) if no level was set
- the tested value happens to match the value set on root

So, the test is weak now as it fails to detect that the logger configuration is completely broken for all except the last logger in settings + root.  To make it stronger and cover the bug, you might want to consider:

- test the level belonging to the logger directly (`.level` property), not the effective level.  That's what's supposed to be set anyway.
- don't set the same level on everything.  If all loggers have the same requested level, then how do you know setting the level per logger even works?  The test is stronger if loggers have different levels so that a bug of the form "one level is wrongly applied to all loggers" would be detected.

rohanpm (changes requested):

exodus_gw/settings.py
nathanegillett: Ah, thanks! No, I only wanted to add the "loggers" key to support subscripting it later. I could add it to the default `log_config` but then any override of that would require a "loggers" key be included.

tests/test_logging.py
nathanegillett: Applied your suggestions. Thanks!

exodus_gw/settings.py
```diff
@@ -62,6 +69,15 @@ def get_settings() -> Settings:
         ]
     )
 
+    for logger in config["loglevels"] if "loglevels" in config else []:
+        if "loggers" not in settings.log_config:
+            settings.log_config.update({"loggers": {}})
```
rohanpm, line 74: There's a built-in method for this to handle it more concisely:

```suggestion
        settings.log_config.setdefault("loggers", {})
```

rohanpm (approved):

tests/test_logging.py
```diff
@@ -0,0 +1,19 @@
...
+
+def test_log_levels():
+    """Ensure loggers are configured according to exodus-gw.ini."""
+
+    logging.getLogger("old-logger").setLevel("DEBUG")
+
+    configure_loggers()
+
+    # Should not alter existing loggers.
+    assert logging.getLogger("old-logger").level == 10
```
lioramilbaum, line 14: Please use the non-numeric logging levels. For example: 10 stands for logging.DEBUG.

lioramilbaum (approved):
> > > It's fully supported, you would have different content for the settings files in different hosts within your ansible inventory.
> > 
> > 
> > How that would work for the dev-X environments?
> 
> The config is defined in group_vars under 'exodus-gw-playbooks/inventory'. So, for our developer namespaces, one easy way would be to just directly edit `exodus_gw_config` in 'exodus_gw_dev.yml', since 'dev' is the environment used by the deployment helper script. Another way would be to provide `exodus_gw_config` to the ansible-playbook extra vars.
> 
> > > Replace "environment variable" with "file" (the other main way of using a ConfigMap) and that's exactly how it works.
> > 
> > 
> > file is tracked in git. Overriding the value requires changing the file, even though the change should not land in the head repo.
> 
> If you were to opt for directly editing the group_vars you wouldn't need to commit the change to deploy it since it's on the playbook side of things.

I am still not convinced this is the simplest approach to handle this. It shouldn't block this PR from been merged. Let's continue with this approach and review this in a later stage, in case we find it necessary.

tests/test_logging.py
nathanegillett: That doesn't work.
```
>       assert logging.getLogger("old-logger").level == "DEBUG"
E       AssertionError: assert 10 == 'DEBUG'
E         +10
E         -'DEBUG'
```
The string representations are a convenience feature for setting logger levels. Levels are stored internally and exposed as an integer.

lioramilbaum: How about `assert logging.getLogger("old-logger").level == logging.DEBUG` ?

nathanegillett: Yeah, I realized I misread your suggestion. Trying that now.

nathanegillett: See PR #76

PR #69, opened 2020-10-30
Refactoring file structure

lioramilbaum:
Every code line in `exodus_gw/__init__.py` is executed for each unit tests which imports a file from the top level folder.
This side affect becomes an issue as soon as we introduce the DB initialization code. It breaks all unit tests.

I have followed the source layout as suggested in the FastAPI documentation:
https://fastapi.tiangolo.com/tutorial/bigger-applications/#import-fastapi


rohanpm (changes requested):
First a couple of notes which are not about the main point of the PR:

- reformatting: I don't think this should be done.  I think it's just that the relevant tool(s) should be configured for line-length=100 - also mentioned in https://github.com/release-engineering/exodus-gw/pull/70.  Current formatting (before this PR) seems fine.
- breaking different parts of the app into separate "routers" - this seems a fine idea.

Now the main thing: I don't think this PR is the right solution to the problem described (actually I don't think it solves the issue at all).

- `db = get_db()` is *still* executed before any tests run.  There's still a test (test_exception_handler) which imports exodus_gw.main, and all tests are imported before any test runs, so I don't see how this PR actually solves anything - `get_db()` was called before all tests previously, and it still is here.
- This will only avoid DB initialization at import time if *not a single test* directly imports whatever module initializes the DB.  That raises the question - how would you get test coverage for that module?  You'd have to do an import from inside of a test, and ensure no other test imports exodus_gw.main.  It's quite fragile.

I think the testing problem is pointing at a design issue: db just shouldn't be a global singleton initialized at import time, i.e. the whole concept of having a `db = get_db()` which is executed during import is wrong.

The better approach I believe would be to make the DB handle a "dependency" using the system described at https://fastapi.tiangolo.com/tutorial/dependencies/ .  DB connection is mentioned as one of the example use-cases there.

I see in your other PR https://github.com/release-engineering/exodus-gw/pull/68/files you've actually started to use this approach - seems like you've removed the `db` global and started to use `Depends` with get_db.

Is that working?  If so, why is *this* PR needed?  Changing the design so that there is no global `db` and we instead use dependency injection seems like it should already solve the problem.

lioramilbaum:
PR #70  was submitted in purpose of rebasing it with the current effort.

lioramilbaum:
> First a couple of notes which are not about the main point of the PR:
> 
> * reformatting: I don't think this should be done.  I think it's just that the relevant tool(s) should be configured for line-length=100 - also mentioned in #70.  Current formatting (before this PR) seems fine.

In general, I agree that reformatting should not be done here. My IDE is configured to perform linting in addition to formatting. I am used to fix those alerts when I work on a new task.
This is something that I hit almost every PR I submit. To avoid that, I've decided to perform a global code base linting and add a linting quality gate to our CI. This is done in PR #70. 
Since PR #70 was merged, I've rebased this commit, which removed the "reformatting" changes you have mentioned.

> * breaking different parts of the app into separate "routers" - this seems a fine idea.

It is not only a good idea. It is also the suggested approach, which was designed and documented by fastAPI maintainers.
Following the guidelines of the maintainers is the preferred approach. A shift from those guidelines should be done only if there is a good reason.
> 
> Now the main thing: I don't think this PR is the right solution to the problem described (actually I don't think it solves the issue at all).

The main idea behind the refactoring was to avoid imports in the __init__.py file. This code is picked up every time a module is imported from the root folder. This is an underside side affect in any case.

> 
> * `db = get_db()` is _still_ executed before any tests run.  There's still a test (test_exception_handler) which imports exodus_gw.main, and all tests are imported before any test runs, so I don't see how this PR actually solves anything - `get_db()` was called before all tests previously, and it still is here.

The get_db function was changed and removed from the main module in PR #68 

> * This will only avoid DB initialization at import time if _not a single test_ directly imports whatever module initializes the DB.  That raises the question - how would you get test coverage for that module?  You'd have to do an import from inside of a test, and ensure no other test imports exodus_gw.main.  It's quite fragile.

I am not following. 

> 
> I think the testing problem is pointing at a design issue: db just shouldn't be a global singleton initialized at import time, i.e. the whole concept of having a `db = get_db()` which is executed during import is wrong.
> 
> The better approach I believe would be to make the DB handle a "dependency" using the system described at https://fastapi.tiangolo.com/tutorial/dependencies/ . DB connection is mentioned as one of the example use-cases there.
> 
I will take a look. Thanks.

> I see in your other PR https://github.com/release-engineering/exodus-gw/pull/68/files you've actually started to use this approach - seems like you've removed the `db` global and started to use `Depends` with get_db.
> 
> Is that working? If so, why is _this_ PR needed? Changing the design so that there is no global `db` and we instead use dependency injection seems like it should already solve the problem.

The code in #68 is working. It was tested manually. I haven't implemented the unit tests yet due to the db initialization which the code in #68 introduced.

nathanegillett (approved):

crungehottman (approved):

exodus_gw/routers/api.py
```diff
@@ -28,20 +28,20 @@
 - The API may enforce stricter limits or policies on uploads than those imposed
   by the AWS API.
 """
```
nathanegillett, line 30: @lioramilbaum @rohanpm @crungehottman I'm not sure this file should've been moved. It's specific to s3 and should probably have remained under the s3 directory.

I think our options are to revert this move or expand the file to include all AWS APIs.
Personally, I'm in favor of the former.

lioramilbaum: It is "cleaner" to have all router files in the same folder. Other than that, I don't have a solid preference here. As long as we have a router object in place, it probably can work in the former structure as well.

lioramilbaum: @nathanegillett Have this change broke the integration tests or anything else for that matter?

crungehottman: If this change doesn't break anything, I have a slight preference for storing the s3 routes in the routers directory because fast-api uses the app/router directory structure in their docs: https://fastapi.tiangolo.com/tutorial/bigger-applications/#import-the-apirouter
Alternatively, we could rename this file to s3.py to make it clear that these routes are s3-specific?

nathanegillett: Nothing is currently "broken" by this but it will become a problem when the DynamoDB API is written. I suppose the real issue is the name of the file; it's specific to S3 operations but is broadly named "api".

I'm willing to accept that router files should live together, so perhaps the solution is to just rename this file. That can be done as part of https://projects.engineering.redhat.com/browse/RHELDST-2161.

If anyone else has other ideas, preferences or suggestions, feel free to share.

nathanegillett: @crungehottman beat me in replying, lol.

It seems the consensus is just to rename the file. Thanks, all.

PR #70, opened 2020-10-31
linting with flake8 via GitHub Actions workflow

lioramilbaum:
flake8 is checking the code base against coding style (PEP8). 
The PR includes a GitHub Actions workflow which runs flake8. It also includes the fixes for the errors flake8 raised in the current code base. 
I have used GitHub Actions in this PR, so that we can see how it works for us as apposed to tox+Travis.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,0 +1,23 @@
+name: Python application
+
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r test-requirements.txt
+      - name: Lint with flake8
+        run: |
+          flake8 --max-line-length=100 .
--- a/exodus_gw/__init__.py
+++ b/exodus_gw/__init__.py
@@ -1,5 +1,6 @@
 # Any modules which add endpoints onto the app must be imported here
-from . import gateway, s3
+from . import gateway, s3  # noqa
 
-# Per ASGI conventions, main entry point is accessible as "exodus_gw:application"
-from .app import app as application
+# Per ASGI conventions, main entry point is accessible as
+# "exodus_gw:application"
+from .app import app as application  # noqa
--- a/exodus_gw/s3/__init__.py
+++ b/exodus_gw/s3/__init__.py
@@ -1,1 +1,1 @@
-from . import api
+from . import api  # noqa
--- a/test-requirements.in
+++ b/test-requirements.in
@@ -19,3 +19,4 @@ six
 sphinx
 wcwidth
 zipp
+flake8
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -92,6 +92,10 @@ docutils==0.15.2 \
     --hash=sha256:9e4d7ecfc600058e07ba661411a2b7de2fd0fafa17d1a7f7361cd47b1175c827 \
     --hash=sha256:a2aeea129088da402665e92e0b25b04b073c04b2dce4ab65caaa38b7ce2e1a99 \
     # via sphinx
+flake8==3.8.4 \
+    --hash=sha256:749dbbd6bfd0cf1318af27bf97a14e28e5ff548ef8e5b1566ccfb25a11e7c839 \
+    --hash=sha256:aadae8761ec651813c24be05c6f7b4680857ef6afaae4651a4eccaef97ce6c3b \
+    # via -r test-requirements.in
 idna==2.10 \
     --hash=sha256:b307872f855b18632ce0c21c5e45be78c0ea7ae4c15c828c20788b26921eb3f6 \
     --hash=sha256:b97d804b1e9b523befed77c48dacec60e6dcb0b5391d57af6a65a312a90648c0 \
@@ -103,7 +107,7 @@ imagesize==1.2.0 \
 importlib-metadata==2.0.0 \
     --hash=sha256:77a540690e24b0305878c37ffd421785a6f7e53c8b5720d211b211de8d0e95da \
     --hash=sha256:cefa1a2f919b866c5beb7c9f7b0ebb4061f30a8a9bf16d609b000e2dfaceb9c3 \
-    # via -r test-requirements.in, pluggy, pytest
+    # via -r test-requirements.in, flake8, pluggy, pytest
 iniconfig==1.1.1 \
     --hash=sha256:011e24c64b7f47f6ebd835bb12a743f2fbe9a26d4cecaa7f53bc4f35ee9da8b3 \
     --hash=sha256:bc3af051d7d14b2ee5ef9969666def0cd1a000e121eaea580d4a313df4b37f32 \
@@ -177,7 +181,7 @@ markupsafe==1.1.1 \
 mccabe==0.6.1 \
     --hash=sha256:ab8a6258860da4b6677da4bd2fe5dc2c659cff31b3ee4f7f5d64e79735b80d42 \
     --hash=sha256:dd8d182285a0fe56bace7f45b5e7d1a6ebcbf524e8f3bd87eb0f125271b8831f \
-    # via pylint
+    # via flake8, pylint
 mock==4.0.2 \
     --hash=sha256:3f9b2c0196c60d21838f307f5825a7b86b678cedc58ab9e50a8988187b4d81e0 \
     --hash=sha256:dd33eb70232b6118298d516bbcecd26704689c386594f0f3c4f13867b2c56f72 \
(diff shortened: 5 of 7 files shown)


.github/workflows/linting.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install flake8
```
rohanpm, line 20: Should this have a locked version of the tool, as we have implemented previously in test-requirements.txt?

Actually I'd suggest including flake8 in test-requirements.{in,txt} and just installing test-requirements.txt here, even though it'll waste a bit of time installing non-flake8 packages.  That will allow it to integrate with the existing dependabot setup.

exodus_gw/__init__.py
```diff
@@ -1,5 +1,6 @@
 # Any modules which add endpoints onto the app must be imported here
-from . import gateway, s3
+from . import gateway, s3  # noqa
```
rohanpm, line 2: What error is suppressed by this noqa?

docs/conf.py
```diff
@@ -92,12 +92,13 @@
     "github_repo": "exodus-gw",
     "github_button": False,
     "github_banner": True,
-    "description": "Publishing microservice for Red Hat's Content Delivery Network",
```
rohanpm, line 95: I see a lot of these changes were to get within the tool's configured line length.

I think it's misconfigured though.  We have typically used a line length of 100 rather than the default of ~80.  Would you mind updating the config instead so the tool is using 100?  That probably will remove the need for many of these changes.

exodus_gw/__init__.py
lioramilbaum: `(exodus) [lmilbaum@lmilbaum exodus-gw]$ flake8 .
./exodus_gw/__init__.py:2:1: F401 '.gateway' imported but unused
./exodus_gw/__init__.py:2:1: F401 '.s3' imported but unused`
This peace of code is about to be removed in a future PR.

docs/conf.py
lioramilbaum: Is there a specific reason to use a non default line length?
Take a look at the following [https://stackoverflow.com/questions/88942/why-does-pep-8-specify-a-maximum-line-length-of-79-characters/88948#88948](discussion)

tests/s3/test_extract_mpu_parts.py
```diff
@@ -6,7 +6,7 @@ def test_typical_body():
 
     body = """
         <?xml version="1.0" encoding="UTF-8"?>
-        <CompleteMultipartUpload xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+        <CompleteMultipartUpload xmlns="http://s3.amazonaws.com/doc/2006-03-01/"> # noqa
```
nathanegillett, line 9: What's this `#noqa` suppressing? It'd be nice if there was more accompanying info with these.

setup.py
```diff
@@ -31,7 +31,7 @@ def get_requirements():
     long_description_content_type="text/markdown",
     classifiers=[
         "Development Status :: 4 - Beta",
-        "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
+        "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",  # noqa
```
nathanegillett, line 34: There's got to be a better way to suppress select lint errors rather than have these unexplained tags everywhere, right? Something like a .pylintrc file would be nice.

docs/conf.py
nathanegillett: I don't have a problem with limiting line length to 80 (79) but I don't think it's really necessary either. It's 2020, we all use an IDE and have 1080p+ displays. If we want to get technical, I believe PEP8 recommends length of 72 characters for docstrings. Should that be enforced as well?

BTW, where is line length even configured? I'm not seeing the metric anywhere.
Also, line 100 suggests 88 characters is expected. Should that and perhaps `page_width` be updated accordingly?

rohanpm: > Is there a specific reason to use a non default line length?

Originally, it comes from my opinion that 80 columns tends to result in code being wrapped too often, especially if you try to use meaningful variable and method names.  Making it slightly more generous at 100 seems to drastically cut down on cases where the code has to be formatted in an ugly manner to keep within the limit, especially if you are going to enforce this on items which aren't sanely able to be autoformatted like doc strings and string literals.

By now it's also about consistency between projects, as the teams working on this family of projects used 100 columns for a few years now, e.g. our standard CI setup in an (internal) Jenkins uses max-line-length=100 when running pylint over all projects; and this value was also used for projects with public CI for some years, e.g. see config in https://github.com/release-engineering/python-fastpurge/blob/master/tox.ini .  We want to use the same setting between projects so we're not reformatting code if we have to move or copy it from one project to another.

So, from my point of view, it was just a project setup bug / oversight that the line length wasn't set to 100 in this project already.  It's an unwanted difference from earlier projects maintained by the same people.

> Take a look at the following discussion

There's a ton of people arguing for every point of view there and no point of view seems particularly compelling.  This is an issue which can sustain bikeshedding indefinitely...

> BTW, where is line length even configured? I'm not seeing the metric anywhere.

There's numerous ways to configure it unfortunately... it can be read from tox.ini, pylintrc, pyproject.toml etc.  We've just got to configure it according to however our current tools use it.

> Also, line 100 suggests 88 characters is expected. Should that and perhaps page_width be updated accordingly?

Not sure (but not in this PR in any case).  I originally set that page_width to get the sources rendering inside of sphinx docs nicely on my browser and I guess it's been copied around uncritically since then.

setup.py
rohanpm: I guess this noqa is for line length as the line happens to be 85 chars.

In my view this is a pretty good point in evidence that the default line length of 80 is just too short.  In my opinion, the above line (without noqa) is entirely reasonable.  It's not intended unnecessarily; it's not mixing too much on a single line.  I believe any method of reformatting this to fit within 80 chars would *decrease* readability rather than increase it.

So I think this line is a good example of why 100 columns is more reasonable.

tests/s3/test_extract_mpu_parts.py
rohanpm: This would be line length again and I think is another point in favor of "80 chars is not enough", as this is completely readable to me and any reformatting is likely to make it only less readable.

exodus_gw/__init__.py
rohanpm: I see, I don't mind these hanging around for now but going forward we should perhaps use the `__all__` solution, like this:

```
__all__ = ["gateway", "s3"]
```

docs/conf.py
```diff
@@ -92,12 +92,13 @@
...
+    "description": "Publishing microservice for Red Hat's CDN",
```
rohanpm, line 95: Setting line length aside for a moment, I also don't like this change on its own merit.  Using an acronym without defining it first is bad form.

rohanpm (changes requested):
Could we please have any of the reformatting around line length taken out of this?

docs/conf.py
lioramilbaum: You might want to take a look at the PEP8 Style guide - https://www.python.org/dev/peps/pep-0008/#maximum-line-length.
The consistency issue is a good reason to continue working with the suggested line width. 
I will continue enforce the default PEP8 style guide in my code contributions. Let me know, when you find code to be unreadable due to this.

tests/s3/test_extract_mpu_parts.py
lioramilbaum: I don't disagree with you and I don't want to get into bikeshedding.  Removing the `#noqa` stanza.

setup.py
lioramilbaum: Again, I see it differently than you do.

docs/conf.py
```diff
@@ -97,7 +97,8 @@
         "Source": "https://github.com/release-engineering/exodus-gw",
         "Index": "genindex.html",
     },
-    # default is 940px which seems to be a little too small to display 88 chars code
+    # default is 940px which seems to be a little too small to display 88 chars
+    # code
```
nathanegillett, line 101: I agree that, if we're not limiting line length under the current 100 chars in this PR, the PR shouldn't include these length-type changes.

nathanegillett (changes requested):

docs/conf.py
lioramilbaum: I will remove those changes. Not a problem. 
Something for consider when providing a code review - 'Why does it bother if comment lines are wrapped?'

nathanegillett: My reason for this request is that a PR or commit shouldn't include changes outside of its objective. If this change _was_ limiting line length, then it would make sense to implement that restriction throughout the project, but that's no longer the case so these changes are therefore unnecessary in implementing flake. Does that make sense?

Comment wrapping itself doesn't bother me at all. In fact, I tend to keep the length of my comments/docstring <=72 characters, per PEP8. That's a personal preference, though, and I don't currently have a strong desire to push that practice on anyone else.

nathanegillett (approved):

docs/conf.py
lioramilbaum: I tend to agree that a PR or commit shouldn't include changes outside of its objective as a general rule. But, general rules are general.  Which means that we are using those rules when it is appropriate and band the rules when it is not. 
Our target is to produce a service which does what it needs to do in an effective way. 

As you said, comment wrapping is not a bad thing to do. It can be done in this PR or in another one.  Does it matter which PR?
Those changes are not pushing anyone to practice something which was not agreed.
Those changes are not making the code unreadable.

nathanegillett: The point is that changes to wrap preexisting lines is unnecessary, as the enforced line length limit isn't changing, so said changes should not be made.

crungehottman (approved):

docs/conf.py
rohanpm: I want to point out something in case it's overlooked:

Are you aware that this project already has an autoformatting tool (black) integrated with CI and a pre-commit hook (mentioned in README)?

One of the points of having an autoformatter is to cut down on exactly this kind of code style bikeshedding, because the rule is simply "your code is formatted as the autoformatter says - end of story".

I'm not sure if you just missed that we already use this tool, or made a conscious decision that it's not sufficient for whatever reason (which isn't explained anywhere).  Anyway I didn't reject this PR due to that, as I know that flake8 and black can be compatible if you configure the line length accordingly and I'd also like to get more experience with github actions.

So the PR as it stands is fine with me, but it's really galling that you would do:

- ignore the formatting tool already used by this project
- make a PR introducing a second slightly-incompatible tool & reformatting all code to fit that tool
- when reviewers complain about this mass reformatting, tell them that they shouldn't care about it

lioramilbaum: Linting and formatting is not the [same](https://medium.com/@awesomecode/format-code-vs-and-lint-code-95613798dcb3#:~:text=Code%20linter%20works%20differently%20than,the%20whole%20codebase%2C%20and%20code). 
With that, I am not sure the assumptions you have listed are relevant/appropriate.

nathanegillett: To change the layout of a file is the definition reformatting. In this case, preexisting lines in several files were wrapped. With the linter set to 100 character line length, those changes were no longer a result of the linting process.

If you're writing _new_ code or comments and want to format it a certain way (within reason), that's fine, but including changes to preexisting content for no particular reason will generally be met with resistance, as we've discussed.

PR #77, opened 2020-11-04
Use 'Python Code Quality and Lint' Action for flake8+isort

lioramilbaum:
Refactoring the linting workflow to use a marketplace Action + validation imports order with isort.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -10,14 +10,18 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.6
-        uses: actions/setup-python@v2
+      - uses: ricardochaves/python-lint@v1.3.0
         with:
-          python-version: 3.6
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -r test-requirements.txt
-      - name: Lint with flake8
-        run: |
-          flake8 --max-line-length=100 .
+          python-root-list: "."
+          use-pylint: false
+          use-pycodestyle: false
+          use-flake8: true
+          use-black: false
+          use-mypy: false
+          use-isort: true
+          extra-pylint-options: ""
+          extra-pycodestyle-options: ""
+          extra-flake8-options: "--max-line-length=100"
+          extra-black-options: ""
+          extra-mypy-options: ""
+          extra-isort-options: ""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,3 +10,11 @@ exclude = '''
   | \dist
 )/
 '''
+
+[tool.isort]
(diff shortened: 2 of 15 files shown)


pyproject.toml
```diff
@@ -9,4 +9,12 @@ exclude = '''
...
\ No newline at end of file
+'''
+
+[tool.isort]
+multi_line_output = 3
+include_trailing_comma = true
+force_grid_wrap = 0
+use_parentheses = true
+ensure_newline_before_comments = true
+line_length = 88
```
nathanegillett, line 20: Should this be 100?
```suggestion
line_length = 100
```

exodus_gw/routers/api.py
```diff
@@ -37,6 +37,8 @@
 from ..s3.client import S3ClientWrapper as s3_client
 from ..s3.util import RequestReader, extract_mpu_parts, xml_response
 from ..settings import get_environment
+from .client import S3ClientWrapper as s3_client
```
crungehottman, line 40: Is this line meant to replace line 37?
` from ..s3.client import S3ClientWrapper as s3_client`

exodus_gw/routers/api.py
```diff
@@ -37,6 +37,8 @@
...
+from .util import RequestReader, extract_mpu_parts, xml_response
```
crungehottman, line 41: Is this line meant to replace line 38?
` from ..s3.util import RequestReader, extract_mpu_parts, xml_response`

.github/workflows/linting.yml
```diff
@@ -10,14 +10,18 @@ jobs:
...
-          python -m pip install --upgrade pip
-          pip install -r test-requirements.txt
-      - name: Lint with flake8
-        run: |
-          flake8 --max-line-length=100 .
+          python-root-list: "."
+          use-pylint: false
+          use-pycodestyle: false
+          use-flake8: true
+          use-black: false
```
crungehottman, line 19: Why is use-black set to false?

nathanegillett: Interesting. Does this mean we set up the action to use the Black tool we already use instead of implementing flake?

lioramilbaum: black quality gate is enforced in Travis. No need to run it twice, unless we want to GitHub Actions to handle that.
black is a formatter where as flake8 is a linter. They do not perform the same checks although there is some overlap.

pyproject.toml
lioramilbaum: Changing this value to 100 breaks black. I have taken this value from here - https://black.readthedocs.io/en/stable/compatible_configs.html

crungehottman (approved):

nathanegillett (approved):

PR #81, opened 2020-11-10
Use functional tests as a quality gate in CI

lioramilbaum:
Adding a GitHub Action which spins the service with the docker-compose.y
Invoke tests from the tests/functional folder


rohanpm (changes requested):
I disagree with the concept of this change, I think it's pushing the test suite in the wrong direction.

What we had before:

- there's a test suite you can run locally by running tox or py.test
- it runs all test code within the py.test process
- it doesn't depend on additional tools beyond what's pip-installable
- it doesn't depend on any running containers
- it achieves 100% test coverage

I think that was ideal.

This change to me is making it worse, as there is now a secondary test suite which has a different process for running, has additional tricky dependencies, isn't going to integrate with debuggers or coverage tools as well, etc.

Can you please explain why it is necessary to fragment the test suite in this way?

lioramilbaum:
> I disagree with the concept of this change, I think it's pushing the test suite in the wrong direction.
> 
> What we had before:
> 
> * there's a test suite you can run locally by running tox or py.test

How does this change contradict this?

> * it runs all test code within the py.test process

How does this change contradict this?

> * it doesn't depend on additional tools beyond what's pip-installable

Let's assume this assumption is correct. Is this a target of the test suite? What value does it serve?

> * it doesn't depend on any running containers

What this assumption means is that the current test suite is limited to only specific set of quality gates. Does that serve well for us?

> * it achieves 100% test coverage

It achieves 100% **code** test coverage.
Does it achieve ?% of functionality test coverage?

> 
> I think that was ideal.
Ideal for which end? 
Tests are implemented to mitigate the product quality as fast and as early as possible. Can you say that was ideal with respect to this statement?

> 
> This change to me is making it worse, as there is now a secondary test suite which has a different process for running, has additional tricky dependencies, isn't going to integrate with debuggers or coverage tools as well, etc.

Please provide evidence to support the listed claims.

> 
> Can you please explain why it is necessary to fragment the test suite in this way?

Let me answer your question with few questions of my own:
Do you think functional tests are reacquired for producing a high quality product?
If the answer to the previous question is yes, in what stage of the delivery pipeline the functional tests should be invoked?

rohanpm:
> > * there's a test suite you can run locally by running tox or py.test
> 
> How does this change contradict this?
> 

It doesn't, but it introduces a dependency on docker, which I deem to be an inappropriate dependency (noting it's not supported by either RHEL or Fedora).

> > * it runs all test code within the py.test process
> 
> How does this change contradict this?
> 

Code under test now runs in a separate container for some of the tests.  (Well, that is, based on https://github.com/release-engineering/exodus-gw/pull/68 - in *this* commit there isn't any real test added, so it's hard to point at anything and demonstrate how it's used.)

> > * it doesn't depend on additional tools beyond what's pip-installable
> 
> Let's assume this assumption is correct. Is this a target of the test suite? What value does it serve?
> 

Yes it is, it gives a standard dev & test setup across projects and ensures developers don't need to learn a new set of dependencies to install for each project to get started on development there.

> > * it doesn't depend on any running containers
> 
> What this assumption means is that the current test suite is limited to only specific set of quality gates. Does that serve well for us?

Well, I'm not against adding some additional test suites which make use of additional tools, beyond the limits I've indicated above.  But I am against allowing any of the code to *only* be tested that way, i.e. allowing less than 100% statement coverage from the so-called "unit" tests with the argument that it'll be covered by one of the other test suites.  So, given that the low-level "unit" test suite needs to achieve 100% coverage, and also given that there is another higher level of testing which occurs on the deployed app as well, I doubt the value of adding an additional test suite in the middle.

> 
> > * it achieves 100% test coverage
> 
> It achieves 100% **code** test coverage.
> Does it achieve ?% of functionality test coverage?
> 

Don't know, but I think it's irrelevant as introducing a new test suite based on docker-compose isn't needed to increase functionality test coverage by reasonable definitions of the term.

> > I think that was ideal.
> > Ideal for which end?
> > Tests are implemented to mitigate the product quality as fast and as early as possible. Can you say that was ideal with respect to this statement?

Ideal for giving developers quick feedback on the quality of their changes which I see as one of the primary goals of the current test suite.

There are multiple levels of testing and I don't expect the tests invoked when you run py.test to necessarily cover everything.  For example, I know the current test suite is not going to cover "does it *really* connect OK to a postgres server" and I think that's OK.

> 
> > This change to me is making it worse, as there is now a secondary test suite which has a different process for running, has additional tricky dependencies, isn't going to integrate with debuggers or coverage tools as well, etc.
> 
> Please provide evidence to support the listed claims.
> 

Don't waste my time, I'm sure you're well aware that matters such as whether a dependency is "tricky" or whether tools are working "well" are going to verge into subjective territory, for which it'll be impractical or at best very time consuming to provide solid evidence.  Every single code review could be tanked by demanding evidence for opinions you disagree with.

> > Can you please explain why it is necessary to fragment the test suite in this way?
> 
> Let me answer your question with few questions of my own:
> Do you think functional tests are reacquired for producing a high quality product?
> If the answer to the previous question is yes, in what stage of the delivery pipeline the functional tests should be invoked?

I'm not sure we're thinking of the same thing for functional testing, but the answers are probably "Yes" and "on incoming pull requests".

But I think this is a strawman.  It's not the case that you're proposing to introduce a functional test suite and I'm against it.  What's really happening is that you're splitting up the test suite according to which technology is used during the test, labelling the suites as "functional" or "unit" and having the "functional" suite work by issuing requests to the app running in a container rather than testing the code in-process.

I'm not arguing we shouldn't have functional tests, I'm arguing it's unnecessary and undesirable to split the test suite in this way and have the functional tests depend on docker-compose (or any external process).

I'm not going to engage in yet another extended debate about this which is little more than two people sharing their subjective opinion, rarely with either managing to persuade the other.  That has already cost enough time and I'm sure we both have better things to do.

Just take it as hard requirements for this project that:

- there must be a test suite which can be run with py.test & tox
- that test suite must not depend on external processes or containers
- that test suite must achieve 100% statement coverage

It's fine if you want to add additional test suites *beyond* that, but I consider it unnecessary and I'll be insisting that full statement coverage is met without involving those.

lioramilbaum:
> > > * there's a test suite you can run locally by running tox or py.test
> > 
> > 
> > How does this change contradict this?
> 
> It doesn't, but it introduces a dependency on docker, which I deem to be an inappropriate dependency (noting it's not supported by either RHEL or Fedora).
> 
> > > * it runs all test code within the py.test process
> > 
> > 
> > How does this change contradict this?
> 
> Code under test now runs in a separate container for some of the tests. (Well, that is, based on #68 - in _this_ commit there isn't any real test added, so it's hard to point at anything and demonstrate how it's used.)
> 
> > > * it doesn't depend on additional tools beyond what's pip-installable
> > 
> > 
> > Let's assume this assumption is correct. Is this a target of the test suite? What value does it serve?
> 
> Yes it is, it gives a standard dev & test setup across projects and ensures developers don't need to learn a new set of dependencies to install for each project to get started on development there.
> 
> > > * it doesn't depend on any running containers
> > 
> > 
> > What this assumption means is that the current test suite is limited to only specific set of quality gates. Does that serve well for us?
> 
> Well, I'm not against adding some additional test suites which make use of additional tools, beyond the limits I've indicated above. But I am against allowing any of the code to _only_ be tested that way, i.e. allowing less than 100% statement coverage from the so-called "unit" tests with the argument that it'll be covered by one of the other test suites. So, given that the low-level "unit" test suite needs to achieve 100% coverage, and also given that there is another higher level of testing which occurs on the deployed app as well, I doubt the value of adding an additional test suite in the middle.
> 
> > > * it achieves 100% test coverage
> > 
> > 
> > It achieves 100% **code** test coverage.
> > Does it achieve ?% of functionality test coverage?
> 
> Don't know, but I think it's irrelevant as introducing a new test suite based on docker-compose isn't needed to increase functionality test coverage by reasonable definitions of the term.
> 
> > > I think that was ideal.
> > > Ideal for which end?
> > > Tests are implemented to mitigate the product quality as fast and as early as possible. Can you say that was ideal with respect to this statement?
> 
> Ideal for giving developers quick feedback on the quality of their changes which I see as one of the primary goals of the current test suite.
> 
> There are multiple levels of testing and I don't expect the tests invoked when you run py.test to necessarily cover everything. For example, I know the current test suite is not going to cover "does it _really_ connect OK to a postgres server" and I think that's OK.
> 
> > > This change to me is making it worse, as there is now a secondary test suite which has a different process for running, has additional tricky dependencies, isn't going to integrate with debuggers or coverage tools as well, etc.
> > 
> > 
> > Please provide evidence to support the listed claims.
> 
> Don't waste my time, I'm sure you're well aware that matters such as whether a dependency is "tricky" or whether tools are working "well" are going to verge into subjective territory, for which it'll be impractical or at best very time consuming to provide solid evidence. Every single code review could be tanked by demanding evidence for opinions you disagree with.
> 
> > > Can you please explain why it is necessary to fragment the test suite in this way?
> > 
> > 
> > Let me answer your question with few questions of my own:
> > Do you think functional tests are reacquired for producing a high quality product?
> > If the answer to the previous question is yes, in what stage of the delivery pipeline the functional tests should be invoked?
> 
> I'm not sure we're thinking of the same thing for functional testing, but the answers are probably "Yes" and "on incoming pull requests".
> 
> But I think this is a strawman. It's not the case that you're proposing to introduce a functional test suite and I'm against it. What's really happening is that you're splitting up the test suite according to which technology is used during the test, labelling the suites as "functional" or "unit" and having the "functional" suite work by issuing requests to the app running in a container rather than testing the code in-process.
> 
> I'm not arguing we shouldn't have functional tests, I'm arguing it's unnecessary and undesirable to split the test suite in this way and have the functional tests depend on docker-compose (or any external process).
> 
> I'm not going to engage in yet another extended debate about this which is little more than two people sharing their subjective opinion, rarely with either managing to persuade the other. That has already cost enough time and I'm sure we both have better things to do.
> 
> Just take it as hard requirements for this project that:
> 
> * there must be a test suite which can be run with py.test & tox
> * that test suite must not depend on external processes or containers
> * that test suite must achieve 100% statement coverage
> 
> It's fine if you want to add additional test suites _beyond_ that, but I consider it unnecessary and I'll be insisting that full statement coverage is met without involving those.

Thank you for the suggestion. I will start a write-up on this topic (as a follow up to the previous write-up on the dev environments).

rohanpm (changes requested):
Recapping my opinion on this PR (and by extension, the related PRs):

It's not necessary or desirable to introduce use of containers into the CI.  The target metric for the project's test suite is 100% statement coverage, which is currently met.  That metric can and should continue to be met without these changes to the test suite; dividing tests into "unit" and "functional", and making the functional tests depend on docker-compose, is an unnecessary complication which should be avoided.

lioramilbaum:
> Recapping my opinion on this PR (and by extension, the related PRs):
> 
> It's not necessary or desirable to introduce use of containers into the CI. The target metric for the project's test suite is 100% statement coverage, which is currently met. That metric can and should continue to be met without these changes to the test suite; dividing tests into "unit" and "functional", and making the functional tests depend on docker-compose, is an unnecessary complication which should be avoided.

Thank you for your feedback. Will you please add it to the write-up?

PR #83, opened 2020-11-13
Autoformat existing code with Black

nathanegillett:


nathanegillett:
Not sure how these made it past our local tox and pre-commit checks.

lioramilbaum (changes requested):
Please take a look why the CI is broken

rohanpm:
> Not sure how these made it past our local tox and pre-commit checks.

Usually that means the output of the tool has changed in a new release, so you may need to ensure you're using exactly the version applied in CI which appears to be black==20.8b1.

nathanegillett:
Very strange. I'm not seeing the issue locally anymore. Black _must have_ upgraded without my knowledge at some point on my system.

PR #89, opened 2020-11-18
Refactor for additional AWS services

nathanegillett:
This commit aims to prepare exodus-gw for the implementation of
additional AWS service APIs, namely DynamoDB.

- Renames exodus_gw/s3 => exodus_gw/aws
- Renames exodus_gw/routers/api.py => exodus_gw/routers/s3.py
- Moves several test files to reflect application structure
- Updates several tests to reflect name & signature changes


lioramilbaum (approved):

exodus_gw/aws/client.py
```diff
@@ -4,28 +4,29 @@
...
-class S3ClientWrapper:
-    """Helper class to obtain preconfigured S3 clients.
+class ClientWrapper:
+    """Helper class to obtain preconfigured clients.
 
     Clients may be wrapped with additional config and event handlers.
     """
 
-    def __init__(self, profile: str):
+    def __init__(self, service: str, profile: str):
```
crungehottman, line 13: Instead of identifying the service via a string, it think it would better to:
* use the ClientWrapper as a parent class
* create a s3 client class inherit the ClientWrapper class/use it as a parent/base class
* override the s3 class's methods where appropriate
* the future dynamodb client will do the same

Thoughts? Otherwise, LGTM.

BTW, this suggestion is not very well thought out and might just be nit-picky (let me know).

nathanegillett: Done!

crungehottman (approved):

lioramilbaum (dismissed):
Please fix broken CI

crungehottman (approved):

lioramilbaum (approved):

PR #90, opened 2020-11-19
Implement initial DynamoDB API [RHELDST-2161]

nathanegillett:
This commit introduces a function to write and delete items on a
DynamoDB table with retry, exponential backoff, and jitter.

This function serves as a base for forthcoming exodus-gw endpoints.

### DIFF
--- a/exodus_gw/routers/dynamodb.py
+++ b/exodus_gw/routers/dynamodb.py
@@ -1,0 +1,42 @@
+import logging
+from typing import List
+
+import backoff
+
+from ..aws.client import DynamoDBClientWrapper as ddb_client
+from ..settings import get_environment, get_settings
+
+LOG = logging.getLogger("exodus-gw")
+
+
+@backoff.on_predicate(
+    backoff.expo,
+    lambda response: response["UnprocessedItems"],
+    max_time=get_settings().write_max_time,
+)
+async def batch_write(env: str, items: List[dict], delete: bool = False):
+    """Write (or delete) 25 items on the given environment's table."""
+
+    if len(items) > 25:
+        LOG.exception("Cannot process more than 25 items")
+        raise ValueError("Received too many items (%s)" % len(items))
+
+    env_obj = get_environment(env)
+    profile = env_obj.aws_profile
+    table = env_obj.table
+
+    if delete:
+        request = {table: [{"DeleteRequest": {"Key": item} for item in items}]}
+        exc_msg = "Exception while deleting %s items from table '%s'"
+    else:
+        request = {table: [{"PutRequest": {"Item": item} for item in items}]}
+        exc_msg = "Exception while writing %s items to table '%s'"
+
+    try:
+        async with ddb_client(profile=profile) as ddb:
+            response = await ddb.batch_write_item(RequestItems=request)
+    except Exception:
+        LOG.exception(exc_msg, len(items), table)
+        raise
+
+    return response
--- a/tests/routers/test_dynamodb.py
+++ b/tests/routers/test_dynamodb.py
@@ -1,0 +1,109 @@
+import mock
+import pytest
+
+from exodus_gw.routers import dynamodb
+
+TEST_ITEMS = [
+    {"web_uri": {"S": "/example/one"}, "from_date": {"S": "2021-01-01"}},
+    {"web_uri": {"S": "/example/two"}, "from_date": {"S": "2021-01-01"}},
+]
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "delete,expected_request",
+    [
+        (False, {"PutRequest": {"Item": item} for item in TEST_ITEMS}),
+        (True, {"DeleteRequest": {"Key": item} for item in TEST_ITEMS}),
+    ],
+    ids=["Put", "Delete"],
+)
+async def test_batch_write(mock_aws_client, delete, expected_request):
+    """Ensure batch_write/delete delegates correctly to DynamoDB."""
+
+    # Represent successful write/delete of all items to the table.
+    mock_aws_client.batch_write_item.return_value = {"UnprocessedItems": {}}
+
+    await dynamodb.batch_write("test", TEST_ITEMS, delete)
+
+    # Should've requested write of all items.
+    mock_aws_client.batch_write_item.assert_called_once_with(
+        RequestItems={"my-table": [expected_request]}
+    )
+
+
+@pytest.mark.asyncio
+async def test_batch_write_retry(mock_aws_client):
+    """Ensure batch_write is retried with unprocessed items."""
+
+    items = [
+        {"web_uri": {"S": "/example/one"}, "from_date": {"S": "2021-01-01"}},
+        {"web_uri": {"S": "/example/two"}, "from_date": {"S": "2021-01-01"}},
+    ]
+
+    # Represent successful write of all items after one retry.
+    mock_aws_client.batch_write_item.side_effect = [
+        {
+            "UnprocessedItems": {
+                "my-table": [{"PutRequest": {"Item": items[1]}}]
+            }
+        },
+        {"UnprocessedItems": {}},
+    ]
+
+    await dynamodb.batch_write("test", items)
+
+    # Should've requested write of all items, then retried one.
+    mock_aws_client.batch_write_item.assert_has_calls(
+        [
+            mock.call(
+                RequestItems={
+                    "my-table": [
+                        {"PutRequest": {"Item": item} for item in items}
+                    ]
+                }
+            ),
+            mock.call(
+                RequestItems={"my-table": [{"PutRequest": {"Item": items[1]}}]}
+            ),
+        ],
+        any_order=False,
+    )
+
+
+@pytest.mark.asyncio
+async def test_batch_write_item_limit(mock_aws_client, caplog):
+    """Ensure batch_write does not accept more than 25 items."""
+
+    items = TEST_ITEMS * 13
+
+    with pytest.raises(ValueError) as exc:
+        await dynamodb.batch_write("test", items)
+
+    assert "Cannot process more than 25 items" in caplog.text
+    assert str(exc.value) == "Received too many items (26)"
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "delete,expected_msg",
+    [
+        (False, "Exception while writing 2 items to table 'my-table'"),
+        (True, "Exception while deleting 2 items from table 'my-table'"),
+    ],
+    ids=["Put", "Delete"],
+)
+async def test_batch_write_excs(mock_aws_client, delete, expected_msg, caplog):
+    """Ensure messages are emitted for exceptions."""
+
+    items = [
+        {"web_uri": {"S": "/example/one"}, "from_date": {"S": "2021-01-01"}},
+        {"web_uri": {"S": "/example/two"}, "from_date": {"S": "2021-01-01"}},
+    ]
+
+    mock_aws_client.batch_write_item.side_effect = ValueError()
+
+    with pytest.raises(ValueError):
+        await dynamodb.batch_write("test", items, delete)
+
+    assert expected_msg in caplog.text
--- a/exodus_gw/aws/client.py
+++ b/exodus_gw/aws/client.py
@@ -78,3 +78,31 @@ class S3ClientWrapper:
         request_dict = kwargs.get("request_dict") or {}
         context = request_dict.get("context") or {}
         context["s3_redirected"] = True
+
+
+class DynamoDBClientWrapper:
+    """Helper class to obtain preconfigured DynamoDB clients.
+
+    Clients may be wrapped with additional config and event handlers.
+    """
+
+    def __init__(self, profile: str):
+        """Prepare a client for the given profile. This object must be used
+        via 'async with' in order to obtain access to the client.
+
+        Note: Session creation will fail if provided profile cannot be found.
+        """
+
+        session = aioboto3.Session(profile_name=profile)
+
+        self._client_context = session.client(
+            "dynamodb",
+            endpoint_url=os.environ.get("EXODUS_GW_DYNAMODB_ENDPOINT_URL")
+            or None,
+        )
+
+    async def __aenter__(self):
+        return await self._client_context.__aenter__()
+
+    async def __aexit__(self, exc_type, exc, tb):
+        await self._client_context.__aexit__(exc_type, exc, tb)
--- a/exodus_gw/settings.py
+++ b/exodus_gw/settings.py
@@ -45,6 +45,9 @@ class Settings(BaseSettings):
     db_service_port: str = "5432"
     """db service port"""
 
+    write_max_time: int = 60
+    """Maximum time for DynamoDB write operations."""
+
     class Config:
         env_prefix = "exodus_gw_"
 
--- a/requirements.in
+++ b/requirements.in
@@ -10,3 +10,4 @@ psycopg2
 sqlalchemy
 async-exit-stack
 async-generator
+backoff
--- a/requirements.txt
+++ b/requirements.txt
@@ -67,6 +67,10 @@ attrs==20.3.0 \
     --hash=sha256:31b2eced602aa8423c2aea9c76a724617ed67cf9513173fd3a4f03e3a929c7e6 \
     --hash=sha256:832aa3cde19744e49938b91fea06d69ecb9e649c93ba974535d08ad92164f700 \
     # via aiohttp
+backoff==1.10.0 \
+    --hash=sha256:5e73e2cbe780e1915a204799dba0a01896f45f4385e636bcca7a0614d879d0cd \
+    --hash=sha256:b8fba021fac74055ac05eb7c7bfce4723aedde6cd0a504e5326bcb0bdd6d19a4 \
+    # via -r requirements.in
 boto3==1.12.32 \
     --hash=sha256:57398de1b5e074e715c866441e69f90c9468959d5743a021d8aeed04fbaa1078 \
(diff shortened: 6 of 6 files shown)


nathanegillett:
For the record, I'm not appreciating container build in CI and flake enforcing import order.

lioramilbaum:
> For the record, I'm not appreciating container build in CI and flake enforcing import order.

You might want to challenge both decisions with submitting a PR or preparing a write-up.  This is the most effective way to get the buy-in of the other team members. Does that make sense?

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Trigger retry if there are unprocessed items.
+    assert not response["UnprocessedItems"]
+
+    return response
+
+
+@backoff.on_exception(
+    backoff.expo,
+    AssertionError,
```
crungehottman, line 66: This scenario is probably extremely unlikely so feel free to ignore: if batch_write_item were to fail on an assertion error, the retry would commence. Did you consider using backoff.on_predicate here instead?

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+    max_time=0,
+    on_giveup=_batch_write_rollback,
+)
```
crungehottman, line 69: I'm curious if you looked into adding jitter: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,110 @@
...
+    # Trigger retry if there are unprocessed items.
+    assert not response["UnprocessedItems"]
+
+    return response
+
+
+@backoff.on_exception(
+    backoff.expo,
+    AssertionError,
+    max_time=0,
```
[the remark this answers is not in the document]
nathanegillett: Good catch! I forgot to revert that after some testing.

exodus_gw/routers/dynamodb.py
nathanegillett: A "full jitter" is applied by default.

exodus_gw/routers/dynamodb.py
nathanegillett: It seems unlikely that boto3/botocore would raise an assertion error but, yes, it would trigger a retry. I'll look into it a bit more. Using `on_predicate` might be better.

exodus_gw/routers/dynamodb.py
nathanegillett: I think I misread the documentation, assumed full jitter was automatically applied, and removed the kwarg from the decorators. Looking at the code, that doesn't seem to be right. I'll put those jitter kwargs back. Thanks.

nathanegillett:
Comments on this PR point to a hole in test coverage. Will mark as ready when that's patched.

exodus_gw/routers/dynamodb.py
nathanegillett: Oh, I found it. `full_jitter` is applied by default.
https://github.com/litl/backoff/blob/128bba5377568b8e2c0230973e887e0b4f5da800/backoff/_decorator.py

crungehottman: Whoops, I missed that. Sorry for the noise :)

nathanegillett:
Switched from exception-based retry to predicate and added `test_batch_write_retry` to ensure it works.

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+
+    env_obj = get_environment(env)
+    profile = env_obj.aws_profile
+    table = env_obj.table
+
+    try:
+        async with ddb_client(profile=profile) as ddb:
+            # Filter unwritten items.
+            items = item_filter(items)
+            response = await ddb.batch_get_item(
```
rohanpm, line 43: I could be missing something here but I don't think we need to do both "get items" and then "delete items which exist" because I expect a DeleteRequest to work just fine even if the item doesn't already exist.

https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DeleteItem.html says:

> Unless you specify conditions, the DeleteItem is an idempotent operation; running it multiple times on the same item or attribute does not result in an error response.

I think we should just directly request deletion of the relevant keys without first trying to detect which of them actually exist.

In fact I think we pretty much have to do it that way, or make other changes, due to eventual consistency.  https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html says:

> When you read data from a DynamoDB table, the response might not reflect the results of a recently completed write operation.

...which means that this code may not even be able to reliably find whatever items it has recently written.

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+                response = await ddb.batch_write_item(
+                    RequestItems={
+                        table: [
+                            {"DeleteRequest": {"Key": item} for item in items}
+                        ]
+                    }
+                )
+    except Exception as exc:
+        LOG.exception(exc)
+        raise exc
```
rohanpm, line 59: The logging method here is not correct I believe, as the point of the `LOG.exception` method is that it already implicitly logs the current exception, i.e. you're meant to use it like:

```
LOG.exception("Error while doing foo")
```

...and then it'll result in a log message like:

```
Error while doing foo
<full exception details>
```

If you pass the exception to `LOG.exception` then I think you're going to just get the exception logged twice, e.g.

```
<str(exc)>
<full exception details>
```

You also don't have to explicitly catch the exception object because LOG.exception knows how to access it already.  So, in summary I think this code should be more like:

```suggestion
    except Exception:
        LOG.exception("Exception while rolling back %s items on %s", len(items), table)
        raise
```

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+    profile = env_obj.aws_profile
+    table = env_obj.table
+
+    request_items = {table: [{"PutRequest": {"Item": item} for item in items}]}
+
+    try:
+        async with ddb_client(profile=profile) as ddb:
+            response = await ddb.batch_write_item(RequestItems=request_items)
+    except Exception as exc:
+        LOG.exception(exc)
```
rohanpm, line 90: Same comment here about usage of LOG.exception.

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+
+from ..aws.client import DynamoDBClientWrapper as ddb_client
+from ..settings import get_environment
+
+LOG = logging.getLogger("exodus-gw")
+
+
+@backoff.on_predicate(
+    backoff.expo,
+    lambda response: response["UnprocessedItems"],
```
crungehottman, line 14: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html#API_BatchGetItem_ResponseSyntax 

```suggestion
    lambda response: response["UnprocessedKeys"],
```

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+    return response
+
+
+@backoff.on_predicate(
+    backoff.expo,
+    lambda response: response["UnprocessedItems"],
+    max_time=600,
+    on_giveup=_batch_write_rollback,
+)
+async def batch_write(env: str, items: List[dict]):
```
rohanpm, line 70: I think there is a bit missing in the overall logic, although maybe it's just planned for later:

Can you clarify how many "items" are expected to be passed here?  One publish might have something like 500,000 items.  Is the caller expected to divide that into chunks of 25 and call this function in a loop, and that's just not written yet?  Or was it expected that this can accept any number of items, and the backoff decorator will end up iterating through them in chunks according to the UnprocessedItems.  If the latter, I don't think that will work well.

I'm also not sure if rollback can be implemented correctly until that looping code exists.  The rollback behavior mustn't just roll back a single batch, e.g. if we're doing a publish with 500,000 items and we start getting errors around item 250,000, it should attempt to roll back all the items leading up to that and not just the latest batch of 25.

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+from ..aws.client import DynamoDBClientWrapper as ddb_client
+from ..settings import get_environment
+
+LOG = logging.getLogger("exodus-gw")
+
+
+@backoff.on_predicate(
+    backoff.expo,
+    lambda response: response["UnprocessedItems"],
+    max_time=600,
```
rohanpm, line 15: I suggest these numbers should be customizable by env vars, so if we've picked the wrong values we can easily change them (or test changes to them).  Ideally the env vars are read via Settings class.

exodus_gw/routers/dynamodb.py
nathanegillett: The response this function returns (and therefore checks) is from `batch_write_item`. While the function does call `batch_get_item` to tell us which of the items are actually on the table, that response, seen filtered on line 48, is consumed by the `batch_write_item` call.

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+        LOG.exception(exc)
+        raise exc
+
+    return response
+
+
+@backoff.on_predicate(
+    backoff.expo,
+    lambda response: response["UnprocessedItems"],
+    max_time=600,
```
rohanpm, line 67: This is probably too large for this operation.  If we get part way through and start to hit errors, and ultimately we can't recover from them, that can be customer-impacting (e.g. can cause yum to fail).  If we set this up so it spends max X minutes retrying this, that's like saying we'll accept customer-impacting broken data on the CDN for up to X minutes before giving up and trying to fix things, so X should not be too high.

To me it makes sense for the rollback operation to have a higher max time, but this one probably should be quite shorter, maybe around 60?

exodus_gw/routers/dynamodb.py
crungehottman: Hmm, nevermind the above suggestion. I now see that this was referring to the response of batch_write_item. But, it's still possible for this function to return a batch_get_item response, which would result in a KeyError.

tests/routers/test_dynamodb.py
```diff
@@ -0,0 +1,143 @@
...
+        {"web_uri": {"S": "/example/two"}, "from_date": {"S": "2021-01-01"}},
+    ]
+
+    mock_aws_client.batch_write_item.side_effect = ValueError("Bad write")
+
+    with pytest.raises(ValueError):
+        await dynamodb.batch_write("test", items)
+
+        captured = capsys.readouterr()
+        assert "[ERROR] Bad write" in captured.out
```
rohanpm, line 132: Per my earlier comments on usage of LOG.exception:

I think we'd want this message to rather give some context on what was being done when the exception happened, e.g. "[ERROR] Exception while writing 2 item(s) to 'test'\n<exception details>".

rohanpm (changes requested):

rohanpm:
A bit of a higher-level comment on this - I would recommend against spending too much time developing the rollback stuff before we have figured out the execution environment for background tasks, because I think these things need to be tied together.

If you've just set up rollback by decorators and such so it all happens in the current process as done here, that should work fine in the case of dynamodb errors.  But this isn't the only scenario, this is also an important scenario:

- exodus-gw worker is in the middle of writing 500,000 items
- the process/container where the worker is running is unexpectedly killed around item 250,000

The system MUST somehow detect this and recover, i.e. resume the interrupted work or roll it back; we can't leave the work partially completed if a process unexpectedly dies.  So this implies to me that the rollback stuff needs to be somehow designed in accordance with how the background workers are picking up tasks in the first place.

exodus_gw/routers/dynamodb.py
nathanegillett: Ah, yeah, I suppose if the `items` list resulting from the query was empty it would return the query's `reponse`. We definitely don't want that.  Thanks.

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,113 @@
...
+
+LOG = logging.getLogger("exodus-gw")
+
+
+@backoff.on_predicate(
+    backoff.expo,
+    lambda response: response["UnprocessedItems"],
+    max_time=600,
+)
+async def _batch_write_rollback(details: dict):
```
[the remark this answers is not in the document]
nathanegillett: Yeah, that looks better to me as well. Thanks.

exodus_gw/routers/dynamodb.py
nathanegillett: Hmm. I was under the impression that an error would be raised if you tried to delete a non-existent key. Re-reading the documentation again, I'm not seeing anything stating or inferring that so maybe it's safe.

That's a good point about eventual consistency, though, that I hadn't considered either.

exodus_gw/routers/dynamodb.py
nathanegillett: Solved this by removing the query altogether, as Rohan suggested.

exodus_gw/routers/dynamodb.py
nathanegillett: > [...] this can accept any number of items, and the backoff decorator will end up iterating through them in chunks according to the UnprocessedItems. If the latter, I don't think that will work well.

That's right. Could you expand a bit on what that might not work to our liking?

> [...] if we're doing a publish with 500,000 items and we start getting errors around item 250,000, it should attempt to roll back all the items leading up to that and not just the latest batch of 25.

Great point. I'll have to think on a solution for this.

exodus_gw/routers/dynamodb.py
nathanegillett: Yeah, I was on the fence with this before, but will do.

exodus_gw/routers/dynamodb.py
rohanpm: >> [...] this can accept any number of items, and the backoff decorator will end up iterating through them in chunks according to the UnprocessedItems. If the latter, I don't think that will work well.
>
> That's right. Could you expand a bit on why that might not work to our liking?

Mainly because of the timing/performance.  One of the purposes of these backoff decorators is to "back off", i.e. wait a bit of time before trying again.  But it doesn't make sense to do that if everything is working normally, you'd rather want to go as fast as possible.

e.g. if you have 500,000 items you do not want this behavior:

- call batch write with 500,000
- write returns 499,975 items unprocessed
- backoff decorator sleeps a bit
- call batch write with 499,975 items
- write returns 499,950 items unprocessed
- backoff decorator sleeps a bit longer...

...etc.  Those sleeps exist for no reason and it would give completely unusable performance with default tuning of backoff decorator.

It seems like that's not possible anyway, https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html says that a batch write request with more than 25 items is simply rejected without even attempting the write.  So, as this is written right now, calling this method with more than 25 items is just not going to work at all, the logic for splitting up a large item list into chunks of 25 has to exist somewhere.

Probably a good idea to ensure tests are updated with some cases of >25 items.

Or alternatively, the scope of this PR could be limited to handling only <= 25 items and we implement the splitting into batches in later PRs.

nathanegillett: Oh, right, of course. Back offs must've slipped my mind at the time.

And, yeah, I guess I misread the doc. Thanks.
The scope for this will be limited to <= 25, then, and in a subsequent PR I'll implement something to split up all the items for writing.

I wonder if executors would be helpful in running multiple `batch_write`s, the results of which could inform the rollback decision. @rohanpm

nathanegillett:
I've finished tweaking finally. Sorry for any update spam.
Please review when able, @rohanpm @crungehottman @lioramilbaum @dichen16

crungehottman (approved):

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,42 @@
...
+@backoff.on_predicate(
+    backoff.expo,
+    lambda response: response["UnprocessedItems"],
+    max_time=get_settings().write_max_time,
+)
+async def batch_write(env: str, items: List[dict], delete: bool = False):
+    """Write (or delete) 25 items on the given environment's table."""
+
+    if len(items) > 25:
+        LOG.exception("Cannot process more than 25 items")
```
rohanpm, line 21: I don't think this is the right logger method, `exception` is meant for cases where there is a genuine Python exception.  https://docs.python.org/3/library/logging.html#logging.exception mentions "This function should only be called from an exception handler".  This should probably just be LOG.error.

rohanpm (commented):
Only saw that one minor logging issue.

nathanegillett:
It seems as though all the publish endpoints are landing in 'routers/gateway.py', so that's probably where I'll end up placing the commit endpoint (for which this PR exists). Therefore, the 'dynamodb.py' doesn't need to be in routers (since it won't contain any endpoints) and can probably live in 'gateway.py' or as a method of the dynamodb client class.
WDYT, @crungehottman, @rohanpm, @lioramilbaum?

exodus_gw/routers/dynamodb.py
```diff
@@ -0,0 +1,42 @@
...
+from ..settings import get_environment, get_settings
+
+LOG = logging.getLogger("exodus-gw")
+
+
+@backoff.on_predicate(
+    backoff.expo,
+    lambda response: response["UnprocessedItems"],
+    max_time=get_settings().write_max_time,
+)
```
nathanegillett, line 16: I may just delete this from the PR, as dramatiq, which we'll use for multi-process, has retry and exponential back-off built into its "actors".

nathanegillett:
closing, -> #95

PR #91, opened 2020-11-19
Add isort to pre-commit hooks

nathanegillett:
Since isort standards are now enforced by CI, including it in pre-commit
will ensure related issues are dealt with locally rather than waiting
for failed Github checks.

- Added isort as a pre-commit hook
- Changed isort 'line_length' to match that of Black
- Updated hook revisions with 'pre-commit autoupdate'
- Ran 'pre-commit run --all-files' (all passed)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,4 +17,4 @@ include_trailing_comma = true
 force_grid_wrap = 0
 use_parentheses = true
 ensure_newline_before_comments = true
-line_length = 88
+line_length = 79
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,11 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: 026c81b83454f176a9f9253cbfb70be2c159d822
+    rev: 20.8b1
     hooks:
     - id: black
+      language_version: python3.6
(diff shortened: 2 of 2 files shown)


crungehottman (approved):

pyproject.toml
```diff
@@ -17,4 +17,4 @@ include_trailing_comma = true
 force_grid_wrap = 0
 use_parentheses = true
 ensure_newline_before_comments = true
-line_length = 88
+line_length = 79
```
rohanpm, line 20: Is this right?  Commit message says:

> Changed isort 'line_length' to match that of Black

But the line length of Black is 88, that's where this number came from in the first place - https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length

So it doesn't seem right that changing a number from 88 to 79 could make something consistent with black.

nathanegillett: The configuration for Black in this same file sets `line-length` to 79. So, either Black or isort should be changed and, as these limits are arbitrary, I didn't have a reason to change the existing Black config.

nathanegillett:
Feel free to merge this on my behalf next week should it get 2+ approvals.

pyproject.toml
rohanpm: Thanks, I missed that.

PR #102, opened 2020-12-10
Implement update_publish_items endpoint [RHELDST-2160]

nathanegillett:
Includes introduction of the Item schema and model and the Publish
model's items attribute.

Co-authored-by: lioramilbaum <lmilbaum@redhat.com>

### DIFF
--- a/exodus_gw/crud.py
+++ b/exodus_gw/crud.py
@@ -1,11 +1,30 @@
+from typing import List, Union
+
+from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.orm import Session
 
-from .models import Publish
+from . import models, schemas
 
 
-def create_publish(db: Session):
-    db_publish = Publish()
+def create_publish(db: Session) -> models.Publish:
+    db_publish = models.Publish()
     db.add(db_publish)
     db.commit()
     db.refresh(db_publish)
     return db_publish
+
+
+def update_publish(
+    db: Session,
+    items: Union[schemas.ItemBase, List[schemas.ItemBase]],
+    publish_id: UUID,
+) -> None:
+
+    # Coerce single items to list.
+    if not isinstance(items, list):
+        items = [items]
+
+    for item in items:
+        db.add(models.Item(**item.dict(), publish_id=publish_id))
+
+    db.commit()
--- a/exodus_gw/routers/gateway.py
+++ b/exodus_gw/routers/gateway.py
@@ -1,10 +1,14 @@
+from typing import List, Union
+from uuid import UUID
+
 from fastapi import APIRouter, Depends
 from sqlalchemy.orm import Session
 
-from .. import schemas
+from .. import models, schemas
 from ..auth import CallContext, call_context
-from ..crud import create_publish
+from ..crud import create_publish, update_publish
 from ..database import SessionLocal
+from ..settings import get_environment
 
 router = APIRouter()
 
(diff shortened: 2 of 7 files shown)


nathanegillett:
Replaces #94

exodus_gw/routers/gateway.py
```diff
@@ -24,16 +28,41 @@ def healthcheck():
...
+    db: Session = Depends(get_db),
+) -> dict:
+    """Update the publish objects with items"""
+
+    # Validate environment from caller.
+    get_environment(env)
+
+    update_publish(db, items, publish_id)
+
+    return {}
```
crungehottman, line 63: The example in the AC suggests that the response would return `{"200": "OK"}` instead of an empty dict -- I see the status code is included on line 47, but I wonder if there is some benefit to returning a non-empty dict?

nathanegillett: It looked strange to me too but @rohanpm explained its purpose on #94: https://github.com/release-engineering/exodus-gw/pull/94#discussion_r538893708
Basically, it still serves as an empty response but allows us to add fields down the road.

crungehottman (approved):

exodus_gw/crud.py
```diff
@@ -1,11 +1,30 @@
...
+from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.orm import Session
 
-from .models import Publish
+from . import models, schemas
 
 
-def create_publish(db: Session):
-    db_publish = Publish()
+def create_publish(db: Session) -> List[models.Publish]:
```
rohanpm, line 9: Is the return type here correct?  It doesn't seem to return a list.

exodus_gw/crud.py
```diff
@@ -1,11 +1,30 @@
...
+    db_publish = models.Publish()
     db.add(db_publish)
     db.commit()
     db.refresh(db_publish)
     return db_publish
+
+
+def update_publish(
+    db: Session,
+    items: schemas.ItemBase or List[schemas.ItemBase],
```
rohanpm, line 19: Hmm, I'm not sure this typing declaration is correct.  If there is not some magic going on deep in the python interpreter, then this 'or' should be just the standard python boolean operator which means that it's just going to return the first truthy value out of the two.  So this is probably equal to just 'schemas.ItemBase' exactly.

A quick test at the console seems to bear this out:

```
>>> import typing
>>> import inspect

# Try to declare a function which accepts a str, or list of str
>>> def myfn(list_or_str: typing.List[str] or str): pass
... 

# What signature do we actually end up with?
>>> inspect.signature(myfn)
<Signature (list_or_str: List[str])>

# => since "truthy1 or truthy2" is always going to return truthy1, the second part is simply lost
```

I believe the correct way is to use Union:

```
typing.Union[schemas.ItemBase, List[schemas.ItemBase]]
```

exodus_gw/routers/gateway.py
```diff
@@ -24,16 +28,41 @@ def healthcheck():
...
 
+@router.put(
+    "/{env}/publish/{publish_id}",
+    status_code=200,
+    tags=["publish"],
+)
+async def update_publish_items(
+    env: str,
+    publish_id: UUID,
+    items: schemas.ItemBase or List[schemas.ItemBase],
```
rohanpm, line 53: Same here about `or` not working the way you want for types.

rohanpm (changes requested):
I think there are some issues with type declarations.

About the typing info, I think it's only needed for the FastAPI endpoints because FastAPI will use it to automate some things.  I'm in favor of adding it in other places as well as a kind of documentation, but it's a little surprising that none of the existing test tools are finding mistakes here (does pylint really not detect a wrong return type?)

If you also want to keep adding the typing info in other places, you might want to look into whether your editor can be configured to parse it and find issues.

exodus_gw/crud.py
nathanegillett: You're right. This was carried over from #94. Not sure what was originally intended but it's now corrected.

nathanegillett:
> About the typing info [...] I'm in favor of adding it in other places as well as a kind of documentation [...].
As am I.
> If you also want to keep adding the typing info in other places, you might want to look into whether your editor can be configured to parse it and find issues.
I'm using VS Code, which I believe you use as well. If so, any suggestions for this?

rohanpm:
> > If you also want to keep adding the typing info in other places, you might want to look into whether your editor can be configured to parse it and find issues.
> 
> I'm using VS Code, which I believe you use as well. If so, any suggestions for this?

No.  I'm using VS Code but haven't looked into this topic.  I've noticed that the typing info influences autocomplete, so sometimes mistakes can be detected indirectly that way.

rohanpm (approved):

nathanegillett:
> > > If you also want to keep adding the typing info in other places, you might want to look into whether your editor can be configured to parse it and find issues.
> > 
> > 
> > I'm using VS Code, which I believe you use as well. If so, any suggestions for this?
> 
> No. I'm using VS Code but haven't looked into this topic. I've noticed that the typing info influences autocomplete, so sometimes mistakes can be detected indirectly that way.

Ah, okay. I did a few searches for extensions but didn't come up with anything.

PR #113, opened 2021-01-13
Implement commit_publish endpoint [RHELDST-2161]

nathanegillett:
This commit introduces a new endpoint used to commit the items stored on
existing publish objects to their place in the given environment's DynamoDB
table. Retry with exponential backoff and jitter was added to the
batch_write function.

The endpoint finds the publish based on the given publish_id,
divides the publish's items into appropriate batch sizes and executes
batch_write operations. Upon failure, the entire operation is aborted
and an attempt is made to delete what was written.

### DIFF
--- a/exodus_gw/crud.py
+++ b/exodus_gw/crud.py
@@ -1,7 +1,7 @@
 from typing import List, Union
 
 from sqlalchemy.dialects.postgresql import UUID
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Query, Session, lazyload
 
 from . import models, schemas
 
@@ -28,3 +28,14 @@ def update_publish(
         db.add(models.Item(**item.dict(), publish_id=publish_id))
 
     db.commit()
+
+
+def get_publish_by_id(
+    db: Session, publish_id: UUID
+) -> Query:  # pragma: no cover
+    return (
+        db.query(models.Publish)
+        .filter(models.Publish.id == publish_id)
+        .options(lazyload(models.Publish.items))
+        .first()
+    )
--- a/exodus_gw/routers/gateway.py
+++ b/exodus_gw/routers/gateway.py
@@ -1,3 +1,5 @@
+import logging
+from itertools import islice
 from typing import List, Union
 from uuid import UUID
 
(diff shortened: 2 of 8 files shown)


exodus_gw/crud.py
```diff
@@ -28,3 +28,14 @@ def update_publish(
         db.add(models.Item(**item.dict(), publish_id=publish_id))
 
     db.commit()
+
+
+def get_publish_by_id(
+    db: Session, publish_id: UUID
+) -> Query:  # pragma: no cover
```
rohanpm, line 35: What's the story with this "no cover"?  There's no obvious reason this wouldn't be covered by tests.

nathanegillett: IMO, it doesn't make sense to write a test for this.
`db` is an aioboto3 `Session` object so I don't believe there's anything here that's worth testing, as the whole string of calls would likely just be mocked. Even if we did mange to test some of the calls, we'd just be testing the sqlalchemy library.

exodus_gw/settings.py
```diff
@@ -45,6 +45,11 @@ class Settings(BaseSettings):
     db_service_port: str = "5432"
     """db service port"""
 
+    batch_size: str = "25"
```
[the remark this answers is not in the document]
nathanegillett: Done

rohanpm (approved):

exodus_gw/routers/gateway.py
```diff
@@ -63,6 +68,54 @@ async def update_publish_items(
...
+        LOG.info("Cleaning up written items. . .")
+
+        for index, batch in enumerate(batches):
+            response = await batch_write(env, batch, delete=True)
+
+            if not response["UnprocessedItems"]:
+                batches.pop(index)
+
+        if batches:
+            LOG.error("Unprocessed items: %s", response["UnprocessedItems"])
```
nathanegillett, line 108: If there are more than once unprocessed item among the batches (2D), only the last one tried will be shown in the error log. It should show all unprocessed entities. I'll try to fix this today.

crungehottman (approved):

rohanpm:
OK to submit this now?  I want to do some refactoring which would conflict with this, so it'd be nice to land this first.

PR #114, opened 2021-01-13
Introduce dev env helpers using systemd & podman

rohanpm:
This commit adds some optional helpers which allow running
exodus-gw and supporting services locally within systemd user
units, along with docs on how to use it.

The included components are currently:

- exodus-gw web app (uvicorn)
- sidecar proxy, for auth/https
- postgres server
- localstack (fake s3)

It's enough to develop and test upload features in exodus-gw
locally with no dependency on remote services. The localstack
instance hopefully will provide a working fake dynamodb as well,
but this hasn't been tested or enabled yet.

To get started trying it:

- you should use a modern version of Fedora
- then run scripts/systemd/install and follow instructions


nathanegillett (changes requested):
I think this should also support DynamoDB tables and include the command matching the Exodus schema
```
$ aws --endpoint-url=http://localhost:3377 dynamodb create-table --table-name test --key-schema AttributeName=web_uri,KeyType=HASH,AttributeName=from_date,KeyType=RANGE --attribute-definitions AttributeName=web_uri,AttributeType=string,AttributeName=from_date,AttributeType=string
```
When run as-is, however...
```
An error occurred (400) when calling the CreateTable operation:
```

rohanpm:
> ```
> An error occurred (400) when calling the CreateTable operation:
> ```

Could it be because the localstack unit right now is set up only to start the s3 service?  That's what this env var controls:

```
scripts/systemd/exodus-gw-localstack.service:  -e SERVICES=s3\
```

Can it be left for a follow-up PR?  It'll take a bit of effort to work out the accurate commands for the cheat sheet and so on.

nathanegillett:
> Can it be left for a follow-up PR? It'll take a bit of effort to work out the accurate commands for the cheat sheet and so on.

Yeah, sounds good

nathanegillett (approved):

PR #118, opened 2021-01-18
Introduce the concept of background tasks, via dramatiq

rohanpm:
Within exodus-gw we need to execute certain operations detached
from the context of an HTTP request, via a separate background
worker process.

This commit includes the initial implementation of this background
worker mechanism, using the dramatiq library (which advertises itself
as a more lightweight alternative to celery).

The vital points are:

- in production, we use a broker based on postgres, so we do not
  need to run an additional broker service (such as rabbitmq)

- we initialize dramatiq-related tables in postgres automatically
  on first use

- in tests, we use the dramatiq stub broker, which allows all
  messages to be handled within the test process

- if using the systemd-based dev env, a dramatiq worker will run
  locally with hot reload enabled

- there is currently a single endpoint, healthcheck-worker, which
  exercises dramatiq. It can be used to verify that workers are
  running.

- as dramatiq has been added to requirements.{in,txt}, the same
  container image used for the exodus-gw asgi app can be used to
  run a dramatiq worker.


crungehottman (approved):

rohanpm:
A couple of changes were made:

- change the env var used for enabling stub broker in tests - the previous approach didn't always work.  In the end I could not find an alternative to setting our own env var.
- rebase this on top of systemd dev-env stuff which was recently merged, and integrate with that dev-env

nathanegillett (approved):

PR #136, opened 2021-01-19
Commit publish entry point files last [RHELDST-2161]

nathanegillett:
Adds functionality to only commit publish items with certain file names
(configurable in settings) when all other items are successfully published.

Additionally, batch_write was simplified to a wrapper for
batch_write_items and a new function, write_batches, was introduced to
handle successive calls to batch_write.

### DIFF
--- a/exodus_gw/routers/gateway.py
+++ b/exodus_gw/routers/gateway.py
@@ -1,5 +1,5 @@
 import logging
-from itertools import islice
+from os.path import basename
 from typing import List, Union
 from uuid import UUID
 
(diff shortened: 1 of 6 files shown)


exodus_gw/routers/gateway.py
```diff
@@ -52,7 +52,7 @@ def healthcheck_worker():
     tags=["publish"],
 )
 async def publish(env: str, db: Session = Depends(get_db)) -> models.Publish:
-    """Returns a new, empty publish object"""
+    """Returns a new, empty publish object."""
```
nathanegillett, line 55: Yeah, I snuck in some docstring punctuation because creating a PR for two periods seems silly.

rohanpm (approved):

crungehottman (approved):

PR #144, opened 2021-01-21
Introduce 'exodus-sync' example script

rohanpm:
This example covers a full recursive sync workflow. It works
similarly to a recursive scp or rsync, but by making use of the
exodus-gw APIs.

This can be used as a quick way to exercise the upload and publish
APIs in a development environment. With latest patches, the entire
process is basically working, but there are various TODOs to be
solved later due to some bugs and unimplemented features in
exodus-gw.

### DIFF
--- a/examples/exodus-sync
+++ b/examples/exodus-sync
@@ -1,0 +1,166 @@
+#!/usr/bin/env python3
+#
+# Upload and publish an entire directory tree via exodus-gw.
+#
+# This command behaves similar to a recursive scp or rsync.
+# It'll walk all files in a source directory and upload them.
+#
+# Once all files have been uploaded, they will be atomically
+# published to destination paths matching the source paths,
+# relative to the top-level directory.
+#
+# Example:
+#
+#  examples/exodus-sync . /sync-test
+#
+# ...would publish the exodus-gw source tree under a 'sync-test' prefix.
+#
+
+import argparse
+import hashlib
+import logging
+import os
+from collections import namedtuple
+from urllib.parse import urljoin
+
+import boto3
+import requests
+
+# Represents a single item to be uploaded & published.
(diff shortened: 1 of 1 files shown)


examples/exodus-sync
```diff
@@ -0,0 +1,166 @@
...
+    # This will ensure all blobs exist in the CDN's s3 bucket (if they weren't
+    # already), but won't yet publish them, so they won't be exposed to clients
+    # of the CDN.
+
+    s3_endpoint = urljoin(args.exodus_gw_url, "upload")
+    s3 = boto3.resource("s3", endpoint_url=s3_endpoint)
+    bucket = s3.Bucket(args.env)
+
+    print("Uploading {} item(s) via {}".format(len(items), s3_endpoint))
+
```
nathanegillett, line 75: Is this consistent with the formatting used elsewhere in the project? I was under the impression we were still using % type formatting because it's still needed by logging.

nathanegillett (approved):

examples/exodus-sync
rohanpm: My intent is to follow generally accepted best practices in both cases.  As far as I can tell, that means you just accept an inconsistency between string formatting for log messages and elsewhere.

If you let loggers do the string formatting, as recommended by e.g. pylint, you've got to use %-style strings because the API is designed exclusively for that.

At the same time, the note at https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting is kinda warning against using that style of format strings in general.

So, as best I can tell, the fact that string formatting for 'print' and 'logging' will end up using different styles is something you're just supposed to accept.

That being said, I wouldn't normally use 'print' at all outside of a command-line example... I think for almost all production code, you'd want to use loggers all the way through.

PR #153, opened 2021-01-26
Allow upload of empty files [RHELDST-4728]

nathanegillett:
No Content-MD5 header is added to upload requests for files which have
no content. This caused a key error in our code, so a static value was
added for empty files, matching that which would be generated.

### DIFF
--- a/exodus_gw/aws/util.py
+++ b/exodus_gw/aws/util.py
@@ -5,6 +5,20 @@ from defusedxml.ElementTree import fromstring
 from fastapi import Response
 
 
+def content_md5(request):
+    """Produce ContentMD5 value expected by S3 APIs.
+
+    When uploading empty files, the Content-MD5 header may not be
+    included in the request when content length is 0. In such cases,
+    return the appropriate base64 encoded md5.
+    """
+
+    if int(request.headers["Content-Length"]) == 0:
+        return "1B2M2Y8AsgTpgAmY7PhCfg=="
+
+    return request.headers["Content-MD5"]
+
+
 def extract_mpu_parts(
     body: str, xmlns: str = "http://s3.amazonaws.com/doc/2006-03-01/"
 ):
--- a/exodus_gw/routers/s3.py
+++ b/exodus_gw/routers/s3.py
@@ -35,7 +35,12 @@ from typing import Optional
 from fastapi import APIRouter, HTTPException, Path, Query, Request, Response
 
 from ..aws.client import S3ClientWrapper as s3_client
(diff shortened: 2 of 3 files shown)


crungehottman (approved):

exodus_gw/routers/s3.py
```diff
@@ -155,12 +155,15 @@ async def object_put(env: str, key: str, request: Request):
...
+    # Allow upload of empty files, which don't have Content-MD5 header.
+    empty_content_md5 = "1B2M2Y8AsgTpgAmY7PhCfg=="
+
     async with s3_client(profile=env_obj.aws_profile) as s3:
         response = await s3.put_object(
             Bucket=env_obj.bucket,
             Key=key,
             Body=reader,
-            ContentMD5=request.headers["Content-MD5"],
+            ContentMD5=request.headers.get("Content-MD5", empty_content_md5),
```
rohanpm, line 166: Could we please use the default value only when there's no data?  (e.g. Content-Length == 0 ?)

Setting a static default checksum whenever the client doesn't provide any seems odd, I'd expect it to be a source of confusing error messages at least, e.g. the client getting an error back suggesting that there's a checksum mismatch when it never sent any checksum.

exodus_gw/aws/util.py
```diff
@@ -5,6 +5,20 @@
...
 
+def content_md5(request):
+    """Produces desired ContentMD5 value for uploading files via boto.
+
+    When uploading empty files, the Content-MD5 header is not included in
+    the request. So we simply need the base64 encoded md5sum for 0 bytes.
+    """
+
+    if int(request.headers["Content-Length"]) == 0:
+        assert "Content-MD5" not in request.headers
```
rohanpm, line 16: I don't think we should assert this.  Can we just delete this line?

It may be the case that boto always omits Content-MD5 if Content-Length is zero, but that seems to be an implementation detail which isn't required by the S3 or exodus-gw APIs and we shouldn't make the API crash unnecessarily if a client behaves differently.  It'd be completely legal for a client to send both Content-Length: 0 and the valid Content-MD5 header themselves and if they do that I don't see any reason we should crash.

Note this API isn't expected to be used exclusively via boto, e.g. exodus-rsync would probably use the "AWS SDK for Go" which is a different codebase and might have different behavior for when it does or doesn't send the header.

rohanpm (changes requested):

dichen16 (approved):

rohanpm (approved):

crungehottman (approved):

PR #154, opened 2021-01-27
Migrate container image & CI from python 3.6 => 3.8

rohanpm:
Python 3.8 is available for UBI8. Let's update to that,
from python 3.6.

The motivations for this include:

- generally keeping up-to-date with modern python

- enabling usage of the contextvars feature [1] introduced in
  python 3.7.  As we are using a web framework which supports
  asyncio, some challenges can arise which are difficult to
  solve without access to something like contextvars.

CI and tox environments were also updated to use python 3.8
and requirements files were recompiled for python 3.8, changing
the set of requirements slightly.  One dependency needed only
for python<=3.6 was dropped.

[1] https://docs.python.org/3/library/contextvars.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ jobs:
     script:
     - docker --version
     - docker build . -f openshift/containers/exodus-gw/Dockerfile
-  - python: "3.6"
+  - python: "3.8"
     env: TOX_ENV=cov-travis
-  - python: "3.6"
+  - python: "3.8"
     env: TOX_ENV=static
-  - python: "3.6"
+  - python: "3.8"
     env: TOX_ENV=docs
 
 script: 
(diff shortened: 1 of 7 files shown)


rohanpm:
Can I get some feedback on whether this would impact development for anyone?  This commit sets python 3.8 as the minimum required version, but if you aren't in the habit of upgrading your OS on your workstation regularly you might be missing it.

The system python was upgraded to 3.8 in Fedora 32, which is the oldest supported Fedora at time of writing.  Therefore this change is consistent with advice that you should use a supported version of Fedora, and if you're still using something older you should probably make time to upgrade.  Still, it would be an option to set python 3.7 as the minimum required version if anyone would find that helpful.

nathanegillett (approved):

crungehottman (approved):

nathanegillett:
I usually upgrade my OS to the latest a few weeks after its launch.

lipoja (approved):

PR #156, opened 2021-01-27
Support S3 head_object operation [RHELDST-4705]

nathanegillett:
Introduces a new upload endpoint using the HEAD operation which
retrieves metadata from an S3 object.

### DIFF
--- a/examples/exodus-sync
+++ b/examples/exodus-sync
@@ -25,6 +25,7 @@ from urllib.parse import urljoin
 
 import boto3
 import requests
+from botocore.exceptions import ClientError
 
 # Represents a single item to be uploaded & published.
 Item = namedtuple("Item", ["src_path", "dest_path", "object_key"])
@@ -76,12 +77,19 @@ def upload_items(args, items):
     for item in items:
         object = bucket.Object(item.object_key)
 
-        # Check if we have it already - if so, we don't need to upload again.
-        # (TODO: we must implement RHELDST-4705 first)
-        # object.load()
+        # Check if object is present in the bucket using object.load(),
+        # catching ClientErrors raised for non-2xx codes.
+        # Upload the file object only when it isn't already present.
+        try:
+            object.load()
+        except ClientError as exc_info:
+            if exc_info.response["Error"]["Code"] == "404":
+                object.upload_file(item.src_path)
+                print("Uploaded {} <= {}".format(item.object_key, item.src_path))
+                continue
+            raise
 
-        object.upload_file(item.src_path)
-        print("Uploaded {} <= {}".format(item.object_key, item.src_path))
+        print("Won't upload {}: already present".format(item.object_key))
 
 
 def publish_items(args, items):
--- a/exodus_gw/routers/upload.py
+++ b/exodus_gw/routers/upload.py
@@ -32,6 +32,7 @@ import logging
 import textwrap
 from typing import Optional
 
+from botocore.exceptions import ClientError
 from fastapi import APIRouter, HTTPException, Path, Query, Request, Response
 
 from ..aws.client import S3ClientWrapper as s3_client
@@ -261,3 +262,28 @@ async def abort_multipart_upload(
         )
 
     return Response()
+
+
+@router.head(
+    "/upload/{env}/{key}",
+    tags=["upload"],
+    summary="Request head object",
+    response_class=Response,
+)
+async def head(
+    env: str = Path(..., description="Target CDN environment"),
+    key: str = Path(..., description="S3 object key"),
+):
+    """Retrieve metadata from an S3 object."""
+
+    env_obj = get_environment(env)
+
+    try:
+        async with s3_client(profile=env_obj.aws_profile) as s3:
+            response = await s3.head_object(Bucket=env_obj.bucket, Key=key)
+    except ClientError as exc_info:
+        # According to botocore documentation, it is safe to rely on
+        # the API to throw exceptions for any non-2xx response.
+        return Response(status_code=int(exc_info.response["Error"]["Code"]))
+
+    return Response(headers={"ETag": response["ETag"]})
--- a/tests/routers/upload/test_head.py
+++ b/tests/routers/upload/test_head.py
@@ -1,0 +1,72 @@
+import pytest
+from botocore.exceptions import ClientError
+from fastapi import HTTPException
+
+from exodus_gw.routers.upload import head
+
+TEST_KEY = "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
+
+
+@pytest.mark.asyncio
+async def test_head(mock_aws_client):
+    """Head request is delegated correctly to S3."""
+
+    mock_aws_client.head_object.return_value = {"ETag": "a1b2c3"}
+
+    response = await head(
+        env="test",
+        key=TEST_KEY,
+    )
+
+    # It should delegate request to real S3
+    mock_aws_client.head_object.assert_called_once_with(
+        Bucket="my-bucket",
+        Key=TEST_KEY,
+    )
+
+    # It should succeed
+    assert response.status_code == 200
+
+    # Response should contain the ETag
+    assert response.headers["ETag"] == "a1b2c3"
+
+
+@pytest.mark.asyncio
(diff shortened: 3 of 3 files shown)


exodus_gw/routers/s3.py
```diff
@@ -264,3 +265,31 @@ async def abort_multipart_upload(
         )
 
     return Response()
+
+
+@router.head(
+    "/upload/{env}/{key}",
+    tags=["head"],
```
nathanegillett, line 272: Wasn't quite sure whether this should be "upload" or "head".

exodus_gw/routers/s3.py
```diff
@@ -264,3 +265,31 @@ async def abort_multipart_upload(
...
+            response = await s3.head_object(Bucket=env_obj.bucket, Key=key)
+    except ClientError as exc_info:
+        # According to botocore documentation, it is safe to rely on
+        # the API to throw exceptions for any non-2xx response.
+        response = exc_info.response
+
+    return Response(
+        headers=response["ResponseMetadata"]["HTTPHeaders"],
+        status_code=response["ResponseMetadata"]["HTTPStatusCode"],
+    )
```
nathanegillett, line 295: Wondering if we should only preserve certain headers or all of them. If the latter, we should adopt this method in other endpoints.

exodus_gw/routers/s3.py
```diff
@@ -264,3 +265,31 @@ async def abort_multipart_upload(
...
+    "/upload/{env}/{key}",
+    tags=["head"],
+    summary="Request head object",
+    response_class=Response,
+)
+async def head(
+    env: str = Path(..., description="Target CDN environment"),
+    key: str = Path(..., description="S3 object key"),
+):
+    """Retreive metadata from an S3 object."""
```
rohanpm, line 280:
```suggestion
    """Retrieve metadata from an S3 object."""
```

exodus_gw/routers/s3.py
rohanpm: URL looks right to me.  We've set things up so that "<exodus_gw_root>/upload" can be used as the S3 endpoint, this should work correctly with the URL here.

I think the tag should be "upload" though - I would still consider this endpoint a part of the "upload API" even if it's not actually doing any uploading, as you would generally be using it in combination with other upload APIs.  e.g. for comparison, an endpoint in the publish API which just gets you the state of a publish object would still be a "publish API" even if it doesn't actually do any publishing during the request.

exodus_gw/routers/s3.py
rohanpm: Yeah, I think we should only return a subset.  The other endpoints currently aren't returning all the headers.

It's good to be defensive here.  I guess we don't have a good understanding of what all the headers are, what info they expose and why somebody might want to use them.  If we return them all, there are risks such as: a client of exodus-gw starts building logic on top of some of the headers without our knowledge, thus getting locked in to some behavior specific to the current storage backend of S3.  In theory it's meant to be an implementation detail which we could replace if we have to.

Personally I think a good balance would be to filter out all the `x-amz-*` headers since these are the ones which are clearly invented for AWS and don't map to any standard HTTP concept.  What do you think?

rohanpm (changes requested):

exodus_gw/routers/s3.py
rohanpm: Forgot to mention that our docs already say we're omitting `x-amz-*`.  I believe I wrote that anticipating that's what we'd end up doing.

rohanpm:
It's not in the acceptance criteria of this issue, but you might want to consider updating examples/exodus-sync as well to check that this really works.

The idea is that the call to `object.load()` in that example should now be able to work so that we can skip uploading anything which is already present.

exodus_gw/routers/s3.py
nathanegillett: Thanks. Yeah, it was the tag that I was questioning.

exodus_gw/routers/s3.py
nathanegillett: Thanks

exodus_gw/routers/s3.py
nathanegillett: Sounds good to me.

nathanegillett:
@rohanpm 
> The idea is that the call to `object.load()` in that example should now be able to work so that we can skip uploading anything which is already present.

The call to `object.load()` works but because we're reaching through boto, it raises ClientError based on our endpoint response's status code. So, I've had to wrap the call in another try/except. If the script were calling our code directly or using curl (like `publish_items` does), this wouldn't be an issue.

examples/exodus-sync
```diff
@@ -76,12 +77,20 @@ def upload_items(args, items):
     for item in items:
         object = bucket.Object(item.object_key)
 
-        # Check if we have it already - if so, we don't need to upload again.
-        # (TODO: we must implement RHELDST-4705 first)
-        # object.load()
-
-        object.upload_file(item.src_path)
-        print("Uploaded {} <= {}".format(item.object_key, item.src_path))
+        # Since the outermost layer used here is boto and our head
```
rohanpm, line 80: The commentary here is a bit confusing to me:

All endpoints return status codes.  Is it suggesting there's some difference in handling of status codes between the implementation of our endpoint and the equivalent in S3 which changes the behavior of `object.load()` on a nonexistent object?  If so we should fix it.

nathanegillett: The non-2xx status code returned by our head endpoint seems to be raised by object.load().
So, when a `ClientError(404)` is caught in `head()` it returns `Response(status_code=404)`, then, it seems, `object.load()` receives that `Response` object and raises `ClientError(404)` again due to the `status_code` it contains.

Does that make sense?

nathanegillett: I've updated the comment. Hopefully it's at least a little clearer.

examples/exodus-sync
```diff
@@ -76,12 +77,20 @@ def upload_items(args, items):
     for item in items:
         object = bucket.Object(item.object_key)
 
-        # Check if we have it already - if so, we don't need to upload again.
-        # (TODO: we must implement RHELDST-4705 first)
-        # object.load()
-
-        object.upload_file(item.src_path)
-        print("Uploaded {} <= {}".format(item.object_key, item.src_path))
+        # The object.load() call reraises status code ClientErrors
```
rohanpm, line 80: I don't want to hold up a PR on this, but I still think this comment is potentially misleading.

It's accurate to say that "If the object doesn't exist, object.load() will raise a ClientError".

Going further and referring to this as "re-raising" a ClientError which was caught by the server seems confusing to me.  One, it's not re-raising in the Python sense - it's creating an entirely new exception.  Two, the fact that the server itself might have caught a ClientError within its own code is entirely incidental.  Yes,in this example both the client and server happen to use boto library and so they both will see an exception of the same type when dealing with a nonexistent object.  But the fact that the server also happens to use boto should be irrelevant to the client.

Another way to think of it - imagine that the server wasn't using boto library to access S3, but a separate FooLibrary which raised a different kind of exception, maybe an ObjectNotFound exception.  The behavior from the server's endpoint would still be the same (returns a 404 response if object doesn't exist).

Then in the boto-using example, would you say "we need to catch a ClientError here because our head endpoint catches FooLibrary ObjectNotFound exceptions and returns an error" ?  It'd be pretty strange for a client to worry about this implementation detail of the server, wouldn't it?  It should only care about the HTTP response and not about whatever exceptions the server was handling internally while preparing that response.

So now if that seems strange, if FooLibrary is replaced with boto and ObjectNotFound is replaced with ClientError, does that make it not strange for some reason?  I don't think so - it's still just an implementation detail that the client shouldn't care about.  To me the comment here seems to falsely suggest that the boto used by the client and by the server are "connected" somehow, which they aren't.

rohanpm (approved):

examples/exodus-sync
nathanegillett: > the fact that the server itself might have caught a ClientError within its own code is entirely incidental

Thanks for helping clarify. This behavior defied my initial expectations and the troubleshooting process skewed my perception a bit, I think.

> this comment is potentially misleading

I agree

rohanpm (approved):

PR #157, opened 2021-01-27
Share transactions between sqlalchemy and dramatiq [RHELDST-4715]

rohanpm:
Each HTTP request is wrapped in a transaction.

Previously, this transaction covered only the sqlalchemy ORM.
It is advantageous to also let the dramatiq broker share the
same transaction, so that when an HTTP request handler runs code
such as:

  x = SomeNewObject()
  db.add(x)              # persist object to DB
  some_actor.send(x.id)  # enqueue dramatiq message in DB

...we can be confident that either *both* the ORM and dramatiq
changes were persisted, or *neither*.  Without this, our code would
be vulnerable to edge cases where one of these operations could
succeed and the other fail.

Although we were already using postgres for both sqlalchemy and
dramatiq, the two libraries managed their own DB connections, and
so wouldn't participate in the same transaction.

This commit fixes that by bridging the two libraries: our custom
broker can have an sqlalchemy session injected, and if so, dramatiq
executes all SQL via that session. The solution is asyncio-aware
so it can be safely used from multiple request handlers
concurrently.

One implication of this change is that a call to '.send' on an actor
won't actually send a message until commit occurs - which, by
default, means it's not sent until the HTTP request succeeds.
If we want to block on a reply, as in the healthcheck-worker example,
we need to explicitly commit so that messages are sent.

Note also that the transaction sharing is designed for use in the
web app only - any code running within a dramatiq worker is still
expected to use the broker in its default configuration without
setting a session.

### DIFF
--- a/exodus_gw/main.py
+++ b/exodus_gw/main.py
@@ -1,5 +1,6 @@
 import logging.config
 
+import dramatiq
 from fastapi import FastAPI, Request
 from fastapi.exception_handlers import http_exception_handler
 from starlette.exceptions import HTTPException as StarletteHTTPException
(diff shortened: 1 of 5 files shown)


nathanegillett (approved):

crungehottman (approved):

rohanpm:
There were a couple of additional pushes to resolve conflicts with the module renaming/reorganization PR.

PR #158, opened 2021-01-28
refactor: reorganize endpoints into appropriate set of modules

rohanpm:
This commit moves various endpoints into more appropriate python
modules. It is a refactor only, with no functional impact.

The initial motivation was to ensure the publish APIs and service
APIs don't live in the same module. These endpoints are quite
unrelated and it doesn't make sense to me that their implementation,
and associated tests, were all dumped into the same 'gateway'
module.

Going a bit further, it seems reasonable to have the python modules
matching the OpenAPI tags used for the endpoints. Both the tagging
and the split between modules are ways of grouping related
functionality and there's no reason to have one grouping in the
source code layout and a different grouping in the API spec, so let's
rename the modules to make them match. This also allows applying
default tags in the APIRouter of each module, so you can't forget
to apply them when adding new endpoints.

### DIFF
--- a/docs/api-guide.rst
+++ b/docs/api-guide.rst
@@ -32,7 +32,7 @@ Uploading blobs
 
 .. seealso:: `Upload API reference <api.html#tag/upload>`_
 
-.. automodule:: exodus_gw.routers.s3
+.. automodule:: exodus_gw.routers.upload
 
 Using boto3 with the upload API
 ...............................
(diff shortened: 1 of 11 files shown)


lipoja (approved):

nathanegillett (approved):

nathanegillett:
This looks great, @rohanpm. Going to merge this so I can start working on top of it.

PR #161, opened 2021-01-29
Add links to Publish objects [RHELDST-4706]

nathanegillett:
This commit introduces the "links" attribute on Publish schema objects.
These links are URLs related to the object, served automatically by the
schema so callers needn't know how to construct the URLs themselves.

### DIFF
--- a/examples/exodus-sync
+++ b/examples/exodus-sync
@@ -110,15 +110,6 @@ def publish_items(args, items):
 
     print("Created publish {}".format(publish))
 
-    # TODO: we shouldn't have to assemble these URLs themselves, the object
-    # should have come with 'links' already - RHELDST-4706.
-    publish["links"] = {
-        "self": "/{env}/publish/{id}".format(env=args.env, id=publish["id"]),
-        "commit": "/{env}/publish/{id}/commit".format(
-            env=args.env, id=publish["id"]
-        ),
-    }
-
     put_url = urljoin(args.exodus_gw_url, publish["links"]["self"])
     for item in items:
         r = session.put(
--- a/exodus_gw/crud.py
+++ b/exodus_gw/crud.py
@@ -1,4 +1,5 @@
 from typing import List, Union
+from uuid import uuid4
 
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.orm import Query, Session, lazyload
@@ -6,8 +7,8 @@ from sqlalchemy.orm import Query, Session, lazyload
 from . import models, schemas
 
 
-def create_publish(db: Session) -> models.Publish:
-    db_publish = models.Publish()
+def create_publish(env: str, db: Session) -> models.Publish:
+    db_publish = models.Publish(id=uuid4(), env=env)
     db.add(db_publish)
     db.commit()
     db.refresh(db_publish)
--- a/exodus_gw/models.py
+++ b/exodus_gw/models.py
@@ -19,6 +19,7 @@ class Publish(Base):
         unique=True,
         nullable=False,
     )
+    env = Column(String, nullable=False)
     items = relationship("Item", back_populates="publish")
 
 
--- a/exodus_gw/schemas.py
+++ b/exodus_gw/schemas.py
@@ -1,8 +1,9 @@
+from os.path import join
 from typing import List
 from uuid import UUID
 
 from fastapi import Path
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, root_validator
 
 PathEnv = Path(
     ...,
@@ -47,10 +48,23 @@ class PublishBase(BaseModel):
 
 
 class Publish(PublishBase):
+    env: str = Field(
+        ..., description="""Environment to which this publish belongs."""
+    )
+    links: dict = Field(
+        {}, description="""URL links related to this publish."""
+    )
     items: List[Item] = Field(
         [],
         description="""All items (pieces of content) included in this publish.""",
     )
+
+    @root_validator
+    @classmethod
+    def make_links(cls, values):
+        _self = join("/", values["env"], "publish", str(values["id"]))
+        values["links"] = {"self": _self, "commit": join(_self, "commit")}
+        return values
 
     class Config:
         orm_mode = True
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -93,7 +93,7 @@ async def publish(
     # Validate environment from caller.
     get_environment(env)
 
-    return create_publish(db)
+    return create_publish(env, db)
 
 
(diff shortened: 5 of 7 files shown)


exodus_gw/crud.py
```diff
@@ -1,13 +1,19 @@
...
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.orm import Query, Session, lazyload
 
 from . import models, schemas
 
 
-def create_publish(db: Session) -> models.Publish:
-    db_publish = models.Publish()
+def create_publish(env: str, db: Session) -> models.Publish:
+    db_publish = models.Publish(id=uuid4())
```
rohanpm, line 12: If there are other APIs which can return a publish object, they'll have to include similar logic.  I wonder if there's a way to make it happen automatically.

rohanpm (changes requested):

exodus_gw/crud.py
nathanegillett: After a few days looking into it, I can't come up with any way to do that, unfortunately. The problem is that the Publish's attributes can't be used within the class because they aren't defined until construction.
LMK if you have any ideas. I feel I've exhausted every possibility within sqlalchemy.

I've also realized that the code here still won't make the links visible in the publish response -- it's just a property of the Python object currently. Submitted a change to the PR for that.

examples/exodus-sync
```diff
@@ -110,16 +111,8 @@ def publish_items(args, items):
...
-    # should have come with 'links' already - RHELDST-4706.
-    publish["links"] = {
-        "self": "/{env}/publish/{id}".format(env=args.env, id=publish["id"]),
-        "commit": "/{env}/publish/{id}/commit".format(
-            env=args.env, id=publish["id"]
-        ),
-    }
-
-    put_url = urljoin(args.exodus_gw_url, publish["links"]["self"])
+    links = json.loads(publish["links"])
```
rohanpm, line 114: This looks broken, as 'publish' is already a dict parsed from JSON - so this only makes sense if we define links as "JSON within JSON".  We shouldn't do that.

exodus_gw/models.py
```diff
@@ -19,6 +19,7 @@ class Publish(Base):
         unique=True,
         nullable=False,
     )
+    links = Column(JSON, nullable=False)
```
rohanpm, line 22: Two points here:

1. I don't think we should store links in the DB since it should be easy to calculate them.
2. However you do need to know the 'env' in order to calculate links.  So this probably points that we should be storing env on publish objects.

In fact it seems like we really ought to be storing env on publish objects regardless, right?  Because otherwise you can create a publish object for one env, and then ask to commit it for another, and the system can't tell that this has happened?

rohanpm (changes requested):

exodus_gw/models.py
nathanegillett: 1) If links aren't stored in DB they won't be included in the response, which was the acceptance criteria here.

2) Latest change enforces env.

examples/exodus-sync
nathanegillett: The links column requires a JSON string, it can't take a dict, afaIk. So the links JSON string needs to be loaded here.

exodus_gw/crud.py
nathanegillett: I managed to get this handled automatically within Publish, only requiring env be stored (which I see is recommended anyway).

examples/exodus-sync
nathanegillett: E.g.;
`Created publish {'id': '2b733145-d28f-4dcf-85f5-af02ecdf2e89', 'env': 'test', 'links': '{"self": "/test/publish/2b733145-d28f-4dcf-85f5-af02ecdf2e89", "commit": "/test/publish/2b733145-d28f-4dcf-85f5-af02ecdf2e89/commit"}', 'items': []}`
^ That's the parsed `Response`, which contains the links JSON string.

exodus_gw/models.py
rohanpm: About not including links in the response if not added here:

These classes define our DB models.  It can't be the case that something *must* be included in the DB in order to be returned by the API.  If it is then we have a problem that we have to solve.

examples/exodus-sync
rohanpm: Yeah, I'm saying this response format doesn't make sense from the client's point of view and we shouldn't implement it this way.

The reasoning "it's stored in DB as JSON so it must be returned here as JSON" isn't valid - APIs aren't supposed to be just a raw dump of whatever's in the DB.  We need to find a way to return this as a proper dict.

I don't know exactly what that would look like, but I'm pretty sure 'links' should not exist on the models object at all, only on the schema object.  FastAPI is converting models objects to schema objects 'somehow', we'll have to learn a bit more about how that works.

rohanpm (changes requested):

examples/exodus-sync
nathanegillett: > I'm pretty sure 'links' should not exist on the models object at all, only on the schema object.

Ah, yeah, that makes a lot of sense. I hadn't considered working this out on the schema side. I'll give that a shot.

nathanegillett:
Logic responsible for generating links now exists on the schema object. Commit message updated.

exodus_gw/schemas.py
```diff
@@ -47,10 +48,23 @@ class PublishBase(BaseModel):
...
+        {}, description="""URL links related to this publish."""
+    )
     items: List[Item] = Field(
         [],
         description="""All items (pieces of content) included in this publish.""",
     )
 
+    @root_validator
+    @classmethod
+    def make_links(cls, values):
```
rohanpm, line 64: Cool.  Just wondering did you find any pydantic docs or similar pointing at this solution? (use a validator to generate additional values on model creation).

Seems like a pretty good approach to me, but I couldn't find anything about this when looking at the docs the other day.

rohanpm (approved):

exodus_gw/schemas.py
nathanegillett: I came across this SO thread: https://stackoverflow.com/a/65128634
The pydantic doc regarding this is here: https://pydantic-docs.helpmanual.io/usage/validators/
However, the SO thread illustrates the idea much better, imo.

rohanpm:
Hope you don't mind if I go ahead and merge this - I think it would be better to get this landed prior to some refactors & migration stuff I'm working on.

nathanegillett:
> Hope you don't mind if I go ahead and merge this - I think it would be better to get this landed prior to some refactors & migration stuff I'm working on.

Fine by me. Was just waiting for a 2nd ack.

PR #168, opened 2021-02-02
Commit publishes in background process [RHELDST-4798]

nathanegillett:
This commit moves the writing of items to DynamoDB to the previously
established dramatiq background worker.

As asyncio operations are incompatible with the new task delegation
method, async/await functionality was removed from the commit_publish
endpoint and the dynamodb functions and client.

### DIFF
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -59,16 +59,14 @@ Publish objects should be treated as ephemeral; they are not persisted indefinit
 """
 
 import logging
-from os.path import basename
 from typing import List, Union
 from uuid import UUID
 
 from fastapi import APIRouter, Body
 from sqlalchemy.orm import Session
 
-from .. import deps, models, schemas
-from ..aws.dynamodb import write_batches
-from ..crud import create_publish, get_publish_by_id, update_publish
+from .. import deps, models, schemas, worker
+from ..crud import create_publish, update_publish
 from ..settings import Environment, Settings
 
 LOG = logging.getLogger("exodus-gw")
@@ -154,7 +152,7 @@ async def update_publish_items(
     status_code=200,
     response_model=schemas.EmptyResponse,
 )
-async def commit_publish(
+def commit_publish(
     publish_id: UUID = schemas.PathPublishId,
     env: Environment = deps.env,
     db: Session = deps.db,
@@ -169,7 +167,7 @@ async def commit_publish(
       - This occurs with all-or-nothing semantics; see [Atomicity](#section/Atomicity).
     - The publish object becomes frozen - no further items can be added.
 
-    Commit occurs asynchronously.  This API returns a Task object which may be used
+    Commit occurs asynchronously.  This API returns a message ID which may be used
     to monitor the progress of the commit.
 
     Note that exodus-gw does not resolve conflicts or ensure that any given path is
@@ -178,29 +176,7 @@ async def commit_publish(
     objects from any of those publishes.
     """
 
-    items = []
-    items_written = False
-    last_items = []
-    last_items_written = False
+    msg = worker.commit.send(publish_id=str(publish_id), env=env.name)
+    LOG.info("Enqueued commit for '%s'", msg.kwargs["publish_id"])
 
-    for item in get_publish_by_id(db, publish_id).items:
-        if basename(item.web_uri) in settings.entry_point_files:
-            last_items.append(item)
-        else:
-            items.append(item)
-
-    if items:
-        items_written = await write_batches(env, items)
-
-    if not items_written:
-        # Delete all items if failed to write any items.
-        await write_batches(env, items, delete=True)
-    elif last_items:
-        # Write any last_items if successfully wrote all items.
-        last_items_written = await write_batches(env, last_items)
-
-        if not last_items_written:
-            # Delete everything if failed to write last_items.
-            await write_batches(env, items + last_items, delete=True)
-
-    return {}
+    return {"commit_message_id": msg.message_id}
--- a/exodus_gw/worker/__init__.py
+++ b/exodus_gw/worker/__init__.py
@@ -4,6 +4,8 @@ from .broker import new_broker
 
 dramatiq.set_broker(new_broker())
 
+from .publish import commit  # noqa
+
 
 @dramatiq.actor(store_results=True, priority=100)
 def ping():
--- a/.pylintrc
+++ b/.pylintrc
@@ -115,3 +115,6 @@ disable=print-statement,
         # both pytest fixtures and fastapi Depends will result in unused
         # arguments as part of normal usage
         unused-argument,
+        # we need to import dramatiq actors to the worker after broker
+        # is set
+        wrong-import-position,
--- a/exodus_gw/aws/client.py
+++ b/exodus_gw/aws/client.py
@@ -1,6 +1,7 @@
 import os
 
 import aioboto3
+import boto3.session
 from botocore.config import Config
 
 
@@ -88,12 +89,12 @@ class DynamoDBClientWrapper:
 
     def __init__(self, profile: str):
         """Prepare a client for the given profile. This object must be used
-        via 'async with' in order to obtain access to the client.
+        via 'with' in order to obtain access to the client.
 
         Note: Session creation will fail if provided profile cannot be found.
         """
 
-        session = aioboto3.Session(profile_name=profile)
+        session = boto3.session.Session(profile_name=profile)
 
         self._client_context = session.client(
             "dynamodb",
@@ -101,8 +102,8 @@ class DynamoDBClientWrapper:
             or None,
         )
 
-    async def __aenter__(self):
-        return await self._client_context.__aenter__()
+    def __enter__(self):
+        return self._client_context
 
-    async def __aexit__(self, exc_type, exc, tb):
-        await self._client_context.__aexit__(exc_type, exc, tb)
+    def __exit__(self, exc_type, exc, tb):
+        return
--- a/exodus_gw/aws/dynamodb.py
+++ b/exodus_gw/aws/dynamodb.py
@@ -21,7 +21,7 @@ LOG = logging.getLogger("exodus-gw")
     predicate=lambda response: response["UnprocessedItems"],
     max_tries=Settings().max_tries,
 )
-async def batch_write(env: str, request: dict):
+def batch_write(env: str, request: dict):
     """Wrapper for batch_write_item with retries and item count validation.
 
     Item limit of 25 is, at this time, imposed by AWS's boto3 library.
@@ -34,8 +34,8 @@ async def batch_write(env: str, request: dict):
         LOG.error("Cannot process more than 25 items per request")
         raise ValueError("Request contains too many items (%s)" % item_count)
 
-    async with ddb_client(profile=environment.aws_profile) as ddb:
-        response = await ddb.batch_write_item(RequestItems=request)
+    with ddb_client(profile=environment.aws_profile) as ddb:
+        response = ddb.batch_write_item(RequestItems=request)
 
     return response
 
@@ -55,9 +55,7 @@ def create_request(env: str, items: List[models.Item], delete: bool = False):
     return {table: [{req_type: {item_type: item.aws_fmt}} for item in items]}
 
 
-async def write_batches(
-    env: str, items: List[models.Item], delete: bool = False
-):
+def write_batches(env: str, items: List[models.Item], delete: bool = False):
     """Submit batches of given items for writing via batch_write."""
 
     environment = get_environment(env)
@@ -71,7 +69,7 @@ async def write_batches(
         request = create_request(env, list(batch), delete)
 
         try:
-            response = await batch_write(env, request)
+            response = batch_write(env, request)
         except Exception:
             LOG.exception(
                 "Exception while %s %s items on table '%s'",
--- a/exodus_gw/worker/publish.py
+++ b/exodus_gw/worker/publish.py
@@ -1,0 +1,41 @@
+from os.path import basename
+
+import dramatiq
+from sqlalchemy.orm import Session
+
+from exodus_gw.aws.dynamodb import write_batches
+from exodus_gw.crud import get_publish_by_id
+from exodus_gw.database import db_engine
+from exodus_gw.settings import Settings
+
+
+@dramatiq.actor
+def commit(publish_id: str, env: str):
+    settings = Settings()
+    db = Session(bind=db_engine(settings))
+
+    items = []
+    last_items = []
+
+    for item in get_publish_by_id(db, publish_id).items:
+        if basename(item.web_uri) in settings.entry_point_files:
+            last_items.append(item)
+        else:
+            items.append(item)
+
+    items_written = False
+    last_items_written = False
+
+    if items:
+        items_written = write_batches(env, items)
+
+    if not items_written:
+        # Delete all items if failed to write any items.
+        write_batches(env, items, delete=True)
+    elif last_items:
+        # Write any last_items if successfully wrote all items.
+        last_items_written = write_batches(env, last_items)
+
+        if not last_items_written:
+            # Delete everything if failed to write last_items.
+            write_batches(env, items + last_items, delete=True)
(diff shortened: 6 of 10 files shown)


nathanegillett:
@rohanpm Not sure I'm understanding how this should work. The code I've tried is commented out in these changes. While playing around with it, the logs led me to believe there's a message being enqueued but not consumed. Do I need to initiate that?

exodus_gw/routers/publish.py
```diff
@@ -158,16 +159,24 @@ async def commit_publish(
...
 
     for item in get_publish_by_id(db, publish_id).items:
         if basename(item.web_uri) in settings.entry_point_files:
             last_items.append(item)
         else:
             items.append(item)
 
+    return await do_commit(env, items, last_items)
+    # msg = await do_commit(env, items, last_items)
+    # return msg.get_result()
```
rohanpm, line 172: A few points here:

- the way to enqueue a message to the actor is to use 'send', and it's not async either.  So I think it should be:

```python
msg = do_commit.send(env, items, last_items)
```

- however I'm not sure sending 'items' and 'last_items' will work because they could be very large.  Ultimately, the dramatiq message is inserted to a postgres table and there will be limits on how large a message can be.  I would strongly suspect our larger publishes would exceed the limit and we can't embed the items directly.  So I think you need to just send the publish ID instead and have the code within the actor look up the publish object and items.

- I don't think we want to use get_result here, the HTTP request handler can't wait for the result as it might take too long.  Actually, in scope of the current issue the idea is that you don't worry about the result yet.  Handling of the result is meant to be covered by a later issue (and it would work by having the dramatiq actor update a Publish and/or Task object as one of its last operations).

So overall, right now I would expect all you can really do is something like this:

```python
msg = do_commit.send(publish.id)
LOG.info("Enqueued commit: %s", msg)
# successful response means we enqueued commit (we will add a task ID or similar here later)
return {}
```

exodus_gw/routers/publish.py
```diff
@@ -158,16 +159,24 @@ async def commit_publish(
...
+
+
+# @dramatiq.actor(store_result=True, priority=100)
```
rohanpm, line 175: I think we would want to not use store_result or priority here.

About store_result, actually result handling is removed from scope of the current issue, but in the end it'll have to work by something other than the built-in dramatiq result backend.  For example, updating state of a Publish or Task object at the end of the function.  I'm not sure we will ever use store_result=True for anything except the healthcheck-worker, and even that might need a rethink at some point.

About priority, I set that as priority=100 on the actor used by healthcheck-worker because that endpoint needs a quick answer on "is any worker alive", and if the priority is low then it would be possible the endpoint would claim that workers are offline while in reality they're just busy.  So it's expected for healthcheck-worker to have a higher priority than other actors.  We probably should let this one have the default priority.

(I think that even with the high priority, healthcheck-worker might still have this issue of "all workers are busy", so that's why I'm saying store_result & priority even there might need to be revised at some point.)

exodus_gw/routers/publish.py
```diff
@@ -158,16 +159,24 @@ async def commit_publish(
...
+async def do_commit(env, items, last_items):
```
rohanpm, line 176: It can't be async as dramatiq isn't designed for it.  This should be a regular old function.

Note: if it looked like messages were being enqueued but never processed, this could be the reason why... as an async function is still callable as a regular function, but doesn't actually do anything.  e.g. see:

```python
>>> async def print_foo():
...   print("foo")
... 
# I can call it but it won't print anything...
>>> print_foo()
<coroutine object print_foo at 0x7f2a6dac45c0>
```

So that might explain it, if you were seeing behavior as if messages were being sent successfully but then nothing happens.

nathanegillett:
I still think something is misconfigured, as I'm getting "connection refused" from dramatiq. I also had to add 'rabbitmq' to reqs (or at least the 'pika' package), which doesn't seem right because we should be using the exodus broker.

Attaching exodus-sync logs (excluding upload portion):
file:///home/ngillett/Documents/commit-in-background.txt

rohanpm:
@nathanegillett, right, dramatiq is going to default to rabbitmq, so if it's trying to use that it means our own broker was not set for some reason.

Two ideas here:

1. If you haven't already, can you use healthcheck-worker endpoint to verify that dramatiq workers are able to work at all for you?  If it does or doesn't work, will give us a hint if this is an issue specific to one endpoint.

2. There may be a problem at the moment with the order of imports.  From what I gather, "@dramatiq.actor" grabs a reference to the current broker at the time the decorator is run.  This would imply that you must import exodus_gw.worker, which sets the broker, before you use @dramatiq.actor.  That doesn't seem to be the case at the moment, so if you rearrange the code to import exodus_gw.worker first then that might make it work.

(2) is probably the cause.  Actually the idea when I introduced the exodus_gw.worker module was that all dramatiq actors would live under there, so to be consistent with that you'd want to move this part of the code out of routers/publish and perhaps into e.g. exodus_gw/workers/publish.py .

This global hidden nature of the broker is quite annoying as it means the behavior can change depending on the order in which modules are imported.  The kind of refactors I was doing in https://github.com/release-engineering/exodus-gw/pull/169 is supposed to reduce this kind of issue but I haven't got to the broker yet.

nathanegillett:
@rohanpm 
> Two ideas here [...]

Yeah, No.2 was the issue. I moved `do_commit` to worker/publish.py and it's working better; just have to work out making the actor args JSON serializable.

exodus_gw/routers/publish.py
```diff
@@ -155,31 +156,45 @@ async def commit_publish(
...
+    msg = do_commit.send(publish_id, env, db)
+    LOG.info("Enqueued commit: %s", msg)
+
+    # TODO: implement message tracking
+    return {}
+
+
+@dramatiq.actor
+def do_commit(
+    publish_id: UUID = schemas.PathPublishId,
```
rohanpm, line 168: Haven't seen your latest patch, but clearing up a few other things here too:

- setting the schemas.* things as default values won't do anything helpful here I think.  They are processed by fastapi's router only, which isn't looking at this function.
- passing "db" into the actor won't work.  Everything is getting serialized into the DB and back out again, you can't pass complex objects like that.
- Depends(get_db) also won't work as Depends is again something provided by fastapi which is not active here.

The last point is a bit of an issue, this code will need to obtain a reference to the DB by itself without the help of fastapi.  At the same time we wouldn't want to duplicate the code which is used in the fastapi case.

A simplistic approach like this may be good enough for now:

```python
engine = db_engine(Settings())
db = Session(bind=engine)
```

nathanegillett:
@rohanpm
When testing with the actor in `worker/publish.py`, the broker wasn't able find it. Wasn't able to figure out why.
```
Feb 04 10:33:06 fedora sh[2438717]: [2021-02-04 10:33:06 -0500] [2438717] [INFO] Enqueued commit: do_commit('4f1b23cc-27d2-4549-a5c5-5849a5130fbc', 'test', ['repomd.xml', 'repomd.xml.asc', 'PULP_MANIFEST'])
Feb 04 10:33:06 fedora sh[2438717]: INFO:     127.0.0.1:34628 - "POST /test/publish/4f1b23cc-27d2-4549-a5c5-5849a5130fbc/commit HTTP/1.1" 200 OK
Feb 04 10:33:06 fedora sh[2426923]: [2021-02-04 10:33:06,269] [PID 2438737] [Thread-2] [dramatiq_pg.broker] [INFO] Consumed 46405e8a-d25b-48d2-a6aa-82b014da7a23@default.
Feb 04 10:33:06 fedora sh[2426923]: [2021-02-04 10:33:06,272] [PID 2438737] [Thread-2] [dramatiq.worker.ConsumerThread(default)] [ERROR] Received message for undefined actor 'do_commit'. Moving it to the DLQ.
Feb 04 10:33:06 fedora sh[2426923]: Traceback (most recent call last):
Feb 04 10:33:06 fedora sh[2426923]:   File "/home/ngillett/src/exodus-gw/.tox/dev-worker/lib/python3.9/site-packages/dramatiq/worker.py", line 320, in handle_message
Feb 04 10:33:06 fedora sh[2426923]:     actor = self.broker.get_actor(message.actor_name)
Feb 04 10:33:06 fedora sh[2426923]:   File "/home/ngillett/src/exodus-gw/.tox/dev-worker/lib/python3.9/site-packages/dramatiq/broker.py", line 214, in get_actor
Feb 04 10:33:06 fedora sh[2426923]:     raise ActorNotFound(actor_name) from None
Feb 04 10:33:06 fedora sh[2426923]: dramatiq.errors.ActorNotFound: do_commit
```
However, after moving the `do_commit` actor to `worker/__init__.py`...
```
Feb 04 12:27:52 fedora sh[2470810]: [2021-02-04 12:27:52 -0500] [2470810] [INFO] Enqueued commit: do_commit('3f7ce4ad-40b7-41bc-88aa-54bd6980fa36', 'test', ['repomd.xml', 'repomd.xml.asc', 'PULP_MANIFEST'])
Feb 04 12:27:52 fedora sh[2470810]: INFO:     127.0.0.1:45328 - "POST /test/publish/3f7ce4ad-40b7-41bc-88aa-54bd6980fa36/commit HTTP/1.1" 200 OK
Feb 04 12:27:52 fedora sh[2469470]: [2021-02-04 12:27:52,848] [PID 2470836] [Thread-2] [dramatiq_pg.broker] [INFO] Consumed b670573c-fe94-4ba6-8d98-aa165a3e62ef@default.
Feb 04 12:27:53 fedora sh[2469470]: [2021-02-04 12:27:53,235] [PID 2470836] [Thread-3] [botocore.credentials] [INFO] Found credentials in shared credentials file: ~/.aws/credentials
Feb 04 12:27:54 fedora sh[2469470]: [2021-02-04 12:27:54,252] [PID 2470836] [Thread-3] [botocore.credentials] [INFO] Found credentials in shared credentials file: ~/.aws/credentials
Feb 04 12:27:55 fedora sh[2469470]: [2021-02-04 12:27:55,255] [PID 2470836] [Thread-3] [botocore.credentials] [INFO] Found credentials in shared credentials file: ~/.aws/credentials
Feb 04 12:27:56 fedora sh[2469470]: [2021-02-04 12:27:56,270] [PID 2470836] [Thread-3] [botocore.credentials] [INFO] Found credentials in shared credentials file: ~/.aws/credentials
Feb 04 12:27:57 fedora sh[2469470]: [2021-02-04 12:27:57,473] [PID 2470836] [Thread-3] [botocore.credentials] [INFO] Found credentials in shared credentials file: ~/.aws/credentials
Feb 04 12:27:57 fedora sh[2469470]: [2021-02-04 12:27:57,610] [PID 2470836] [Thread-3] [exodus-gw] [INFO] Items successfully written
```
This did require removing ALL asyncio statements from dynamodb functions (and therefore all publish endpoints), as removing `await` from do_commit triggered a cascade of coroutine problems.

I'll update the commit when tests around publish stuff are fixed.

exodus_gw/routers/publish.py
```diff
@@ -146,29 +144,11 @@ async def commit_publish(
...
     """
 
-    items = []
-    items_written = False
-    last_items = []
-    last_items_written = False
+    msg = worker.commit.send(
+        publish_id=str(publish_id),
+        env=env.name,
+        entry_point_files=settings.entry_point_files,
```
rohanpm, line 150: What's the reason behind sending this setting rather than letting the actor read it from environment?

exodus_gw/routers/publish.py
```diff
@@ -146,29 +144,11 @@ async def commit_publish(
...
-    elif last_items:
-        # Write any last_items if successfully wrote all items.
-        last_items_written = await write_batches(env, last_items)
-
-        if not last_items_written:
-            # Delete everything if failed to write last_items.
-            await write_batches(env, items + last_items, delete=True)
-
-    return {}
+    return {"Commit Message ID": msg.message_id}
```
rohanpm, line 154: Is this meant as temporary so you can tell what it's doing until we have the proper Task stuff?  In the long run I'm not sure we should be returning the message ID.

exodus_gw/worker/__init__.py
```diff
@@ -1,10 +1,50 @@
...
+from exodus_gw.database import db_engine
+from exodus_gw.settings import Settings
 
 from .broker import new_broker
 
 dramatiq.set_broker(new_broker())
 
 
+@dramatiq.actor
+def commit(publish_id: str, env: str, entry_point_files: List[str]):
```
rohanpm, line 18: Could we please put it in a separate file?  We should avoid having to put all of our actors into the same file.  I should have done this with the 'ping' actor as well to get the convention set up.

I know you had a problem with it before.  To me it sounds like the problem was that the file containing this actor was never imported by the worker (while it was from the web app).

The requirement for this all to work is that:
- the actor is imported when `exodus_gw/worker/__init__.py` is imported
- the call to `@dramatiq.actor` happens after the call to `set_broker`

So two ideas how to do this are:

1. Import the other modules containing actors only after the set_broker call, e.g.

```python
from .broker import new_broker

dramatiq.set_broker(new_broker())

# Now that we've set our broker, import any modules containing actors.
from . import publish
```

This is probably going to require some 'noqa' or 'pylint disable' comments due to the fact that all imports wouldn't be grouped.

2. Declare 'actors' as plain functions within their own modules.  Then make them all into actors here.

```python

from .broker import new_broker
from . import publish

# Set our broker.
dramatiq.set_broker(new_broker())

# Now that our broker is set, we can initialize all our actors.
commit = dramatiq.actor(publish.commit)
# (and others could look like this later...)
ping = dramatiq.actor(service.ping)
scheduled_gc = dramatiq.actor(scheduled.garbage_collect)
```

exodus_gw/routers/publish.py
nathanegillett: I guess the intention was the forthcoming api could grab this ID and use it to poll/check completion.

exodus_gw/routers/publish.py
nathanegillett: Can we trust that the Settings() on the worker will be the same as the one here?

exodus_gw/worker/__init__.py
nathanegillett: Ah, okay. Well, of the two, I think I prefer the second approach.

nathanegillett: Done.

exodus_gw/routers/publish.py
rohanpm: Not really.  In the steady state it will be, the YAML we use for openshift is set up to strongly encourage that they get identical env vars and config files.  But around upgrades it might not be since they restart at different times.

I don't think that's enough reason to send the setting though.  This is a choice between these cases:

1. the setting which applies is the current value *when you requested the commit*, or
2. the setting which applies is the current value *when the commit starts running*

To me there's not a clear reason that either one is definitely better than the other.  Both of them have some logic behind it.  I'm leaning towards (2) being better for us when managing the service (if you imagine that we know we have to change some setting, I think we'd want to be confident that the changed value applied everywhere as soon as the deployment was complete, and we didn't have to worry about some old values being stored in some enqueued messages).

Anyway, one of the reasons I'm asking is just that only *one* setting is being captured here.  Other settings are relevant as well, like the DB settings, batch size, max tries, environments.  So *if* it were correct to take a snapshot of settings at the time the commit was requested and include that in the message, it's not clear why we should do that for just this one setting while ignoring the others.

exodus_gw/routers/publish.py
rohanpm: If this was meant to be permanent then there would be a few fixes needed:

- we shouldn't use "Commit Message ID" as attribute name, it's inconsistent with the style in other responses which is more like "commit_message_id"
- a response model should be declared so we get proper docs

But I would recommend just treating this as temporary anyway.  I think similar to the publish creation API, which returns a publish object and not just an ID, we would also want this API to return some kind of task object (and perhaps it would make sense for the task ID to be equal to message ID?) That work I think would fall into RHELDST-2162.

exodus_gw/routers/publish.py
nathanegillett: Okay, yeah, I had forgotten that settings are pulled in dynamodb.py anyway.

exodus_gw/routers/publish.py
nathanegillett: Yeah, agreed. This can just be temporary and a task model will be established in RHELDST-2162 and returned here. And, yes, I think it makes sense to use the message ID as our task ID.

nathanegillett:
Rebased

rohanpm (approved):

PR #169, opened 2021-02-03
Refactor lifecycle of dependencies, reduce usage of globals

rohanpm:
This PR refactors the lifecycle of three kinds of objects and makes
greater usage of the dependency injection system:

- DB engine
- Settings
- Environment

The motivation behind these changes is to increase testability.
With the previous approach for handling these objects, we had
several globals constructed once at import time, and sometimes
the dependency on these globals would be hidden.

That made it more difficult to test code since you could not
simply run code with different values for Settings and expect it
to work properly - while *some* code might respect the new
settings, other code might be using globals initialized from the
old settings. That led to some hacks in tests, like one of the tests
bypassing the settings LRU cache, and the sqlite fixture having
to re-initialize the global DB engine.

With these refactors we're much closer to having it set up so that
one can `monkeypatch.setenv("EXODUS_GW_WHATEVER", ...)`
from within a test and have it "just work" the way one would
expect.

This is relevant to upcoming changes relating to DB
migrations/alembic which I've found difficult to test while having
the global DB engine.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -107,4 +107,6 @@ disable=print-statement,
         bad-continuation,
         line-too-long,
         invalid-name,
-
+        # both pytest fixtures and fastapi Depends will result in unused
+        # arguments as part of normal usage
+        unused-argument,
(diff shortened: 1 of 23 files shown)


rohanpm:
Note there are more details in individual commits if desired.

nathanegillett (approved):

rohanpm:
Rebased to fix conflicts with 'links' change.

PR #172, opened 2021-02-04
Introduce alembic for managing DB schema [RHELDST-4844]

rohanpm:
We will need to make changes to the DB schema over time and we
must have some process for dealing with this. There are many
schema changes which the earlier approach (MetaData.create_all)
will not be able to deal with. Let's integrate with alembic,
which is designed for sqlalchemy.

- the service performs migrations automatically during startup

- this commit includes an initial migration which is destructive;
  existing tables will be dropped to ensure the schema is
  consistent across all environments

- various settings have been added relating to migrations

- migrations can be disabled during local development for rapid
  prototyping of schema changes

- a wrapper script is provided to assist with autogeneration of
  new migrations

- some notes are added to deployment & development guides explaining
  migrations


rohanpm:
Note there's two commits in the PR, one of them is just the output of "alembic init", so you can look at them to understand which code was generated by alembic and the changes I've made on top of that if you like.

crungehottman (approved):

nathanegillett (approved):

PR #187, opened 2021-02-10
tests: add an 'update publish items' test using TestClient

rohanpm:
Ensure we have at least one test of the endpoint which doesn't rely
on mocking, so we know the DB operations are being verified.

### DIFF
--- a/tests/routers/test_publish.py
+++ b/tests/routers/test_publish.py
@@ -1,3 +1,5 @@
+import uuid
+
 import mock
 import pytest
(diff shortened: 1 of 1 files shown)


nathanegillett (approved):

rohanpm:
This will need to be revised a little to continue working with "enforce object key format".

rohanpm:
It was rebased and tweaked slightly to make tests pass now that we are validating the object key format.

PR #188, opened 2021-02-10
Enforce object key format [RHELDST-4845]

nathanegillett:
This commit adds validation for outgoing object keys, esuring all object
keys to be uploaded or published are of sha256 format.

### DIFF
--- a/exodus_gw/aws/util.py
+++ b/exodus_gw/aws/util.py
@@ -1,9 +1,17 @@
 import io
+import re
 from typing import AnyStr
 from xml.etree.ElementTree import Element, ElementTree, SubElement
 
 from defusedxml.ElementTree import fromstring
-from fastapi import Response
+from fastapi import HTTPException, Response
+
+
+def validate_object_key(key: str):
+    pattern = re.compile(r"[0-9a-f]{64}")
+
+    if not re.match(pattern, key):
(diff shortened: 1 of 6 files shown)


exodus_gw/aws/util.py
```diff
@@ -1,9 +1,17 @@
...
+import re
 from typing import AnyStr
 from xml.etree.ElementTree import Element, ElementTree, SubElement
 
 from defusedxml.ElementTree import fromstring
-from fastapi import Response
+from fastapi import HTTPException, Response
+
+
+def validate_object_key(key: str):
```
rohanpm, line 10: This approach should work fine, but there is also regex validation support built-in to both "Path" as used by the fastapi router, and "Field" as used on the model if you weren't aware.

e.g. something like this might already do it:

```
    # for endpoints
    key: str = Path(..., description="S3 object key", regex="<...the appropriate regex>"),

    # and for ItemBase class
    object_key: str = Field(
        ...,
        regex="...",
        description=...,
    )
```

You might want to give it a try and see if it feels cleaner, I guess at least one advantage would be it's less likely to forget adding the validation somewhere.

rohanpm (approved):

nathanegillett:
@crungehottman @lipoja @dichen16 please review when you get a chance

crungehottman (approved):

PR #196, opened 2021-02-16
Add actor time limit setting [RHELDST-4973]

nathanegillett:
Initial time limit set to 30 minutes to accomidate large publishes.

### DIFF
--- a/exodus_gw/worker/publish.py
+++ b/exodus_gw/worker/publish.py
@@ -14,7 +14,7 @@ from exodus_gw.settings import Settings
 LOG = logging.getLogger("exodus-gw")
 
 
-@dramatiq.actor
+@dramatiq.actor(time_limit=Settings().actor_time_limit)
 def commit(publish_id: str, env: str):
     settings = Settings()
     db = Session(bind=db_engine(settings))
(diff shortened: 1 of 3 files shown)


crungehottman (approved):

exodus_gw/worker/publish.py
```diff
@@ -14,7 +14,7 @@
 LOG = logging.getLogger("exodus-gw")
 
 
-@dramatiq.actor
+@dramatiq.actor(time_limit=Settings().actor_time_limit)
```
rohanpm, line 17: The doc string for the setting suggests that this is applied onto *all* actors but instead devs have to explicitly opt-in to setting this on each actor individually.  This most likely means it will be missed on some of them.

Is there any way to set this as the default for *all* actors defined on our broker?

rohanpm (approved):

exodus_gw/worker/publish.py
nathanegillett: Doesn't look to be an option on brokers. I think we would need to implement class-based actors for that, https://dramatiq.io/reference.html#class-based-actors.

PR #197, opened 2021-02-16
Add "FAILED" task state

nathanegillett:
Tasks are now set to this state if an exception is raised during the
DynamoDB writing process. Previously in this scenario, the task would
remain "IN_PROGRESS" indefinitely.


nathanegillett:
@rohanpm should this be changed to "FAILED" as seen in 5083?

exodus_gw/worker/publish.py
```diff
@@ -27,8 +27,8 @@ def commit(publish_id: str, env: str):
 
     if task.state == "COMPLETE":
```
nathanegillett, line 28: We still permit failed commits to be retried. Is that okay?

exodus_gw/worker/publish.py
```diff
@@ -48,19 +48,26 @@ def commit(publish_id: str, env: str):
...
+            last_items_written = write_batches(env, last_items)
 
-        if not last_items_written:
-            # Delete everything if failed to write last_items.
-            write_batches(env, items + last_items, delete=True)
+            if not last_items_written:
+                # Delete everything if failed to write last_items.
+                write_batches(env, items + last_items, delete=True)
+    except Exception:
+        LOG.warning("Task %s encountered error\nFailing commit", str(task.id))
```
rohanpm, line 66: I'm having a hard time reconciling this with above code which deletes items in case of a partial write.  Something seems a bit broken here, like not all error cases are handled consistently.

Code like this:

```
            if not last_items_written:
                # Delete everything if failed to write last_items.
                write_batches(env, items + last_items, delete=True)
```

...suggests that the expected behavior is for write_batches to return False on failure.  But if that's the case then is this possible:

- write_batches on last_items returns False, doesn't raise
- write_batches with delete=True is called, deletes everything successfully, doesn't raise.
- as no exception is raised, the 'state = failed' block doesn't get entered and the task is marked as complete.

So a complete task, from the caller's point of view, could mean either "everything was committed" or "nothing was committed"?  I don't think we'd want that.  If the commit was rolled back due to an error then we would want the client to see that as a failed task.

What is the desired behavior of write_batches actually?  Is it supposed to raise on failure, or return False on failure, or raise for some kinds of failures and return False for others?

exodus_gw/worker/publish.py
```diff
@@ -48,19 +48,26 @@ def commit(publish_id: str, env: str):
...
+        # Change task state to FAILED.
+        task.state = schemas.TaskStates.failed
+        db.commit()
+        raise
```
rohanpm, line 70: I would expect the default actor retry to kick in at this point, so as currently written the behavior might be like this:

- failure occurs
- actor sets task state to FAILED
- actor raises
- some time later, actor is called again on same task due to retry and runs again

That would mean the task state flips between 'failed' and 'in progress' during retries.  I don't think we'd want that, as a client of exodus-gw would have no way of knowing if 'failed' meant 'permanently failed', or meant 'failed right now and I'm going to retry'.

I suggest that FAILED state should only be set after all retries have been exhausted, or alternatively that if we get here, we don't allow any further retries (as batch_write itself already did some retries).  The latter could be accomplished by replacing the "raise" with just logging the exception and returning.

Basically we should return if we are done processing this task, or leave task IN_PROGRESS and allow an exception to be raised if we would like to be retried.

tests/worker/test_actors.py
```diff
@@ -133,4 +134,31 @@ def test_commit_completed_task(
...
+):
+    # Construct task that would be generated by caller.
+    task = _task(mock_publish.id)
+    # Construct dramatiq message that would be generated by caller.
+    mock_get_message.return_value = mock.MagicMock(
+        message_id=task.id, kwargs={"publish_id": mock_publish.id}
+    )
+    # Simulate successful write of items, failed write of entry point items
+    # and then successful deletion of items.
+    mock_write_batches.side_effect = ValueError()
```
rohanpm, line 153: Doesn't this raise a ValueError every time?  I don't get how this implements what the comment above says.  Successful write of items shouldn't raise ValueError, right?

rohanpm (changes requested):

exodus_gw/worker/publish.py
rohanpm: No, we should leave it IN_PROGRESS if we're going to retry.  Otherwise it's not clear what the client is supposed to do when they see FAILED, as the commit still *might* succeed later despite us telling the client it failed, so are they just supposed to keep polling?

tests/worker/test_actors.py
```diff
@@ -133,4 +134,31 @@ def test_commit_completed_task(
...
+
+    with TestClient(app):
+        db.add(mock_publish)
```
rohanpm, line 156: It's not introduced by this commit, but I see the fixture naming here is quite confusing.  We seem to be using `mock_*` to mean three different things:

- a unittest.mock object (Mock or MagicMock)
- real objects from 'schemas' module
- real objects from 'models' module

At first I thought this code was wrong as adding a Mock or MagicMock to db doesn't make sense, but I see now mock_publish comes from the third category so it's right.  The naming is an absolute mess here though and would be nice to clean up.

tests/worker/test_actors.py
nathanegillett: This test was copy/pasted from one of the above, I just failed to change the comments.

tests/worker/test_actors.py
nathanegillett: Yeah, I struggled with this myself recently. Created another PR to fix this: https://github.com/release-engineering/exodus-gw/pull/198

exodus_gw/worker/publish.py
nathanegillett: Currently, `write_batches` raises if a delete request returns unprocessed items but returns false if a put request returns unprocessed items. This is because a delete request should be submitted for a put failure but a delete failure will have to be handled (cleaned up) manually.

So, if an exception is caught here it means that a delete request failed (after retries), the batch of items was somehow too large, or there was an exception within boto. In all of those cases, I think it's appropriate to say the task has failed.

exodus_gw/worker/publish.py
nathanegillett: I think this raises a larger question of whether or not we want the actor to retry at all. The actual writing of the items to DynamoDB is already retried on failure as defined in settings. So, potential retries are multiplied by the number of retries we allow the actor to attempt.

nathanegillett: I actually had an issue with this while testing publish times and am of the opinion that we should hardcode the value of this actor's `max_retries` as 1.

exodus_gw/worker/publish.py
nathanegillett: With the change to the task.state check at the beginning of this function, a failed task wouldn't be allowed. Should we still `return` rather than `raise` when marking a task as failed?

exodus_gw/worker/publish.py
rohanpm: I think we should leave retries enabled in general, due to the fact that the DynamoDB steps aren't the only ones which can fail.

For example: the db.commit which sets the task state to IN_PROGRESS at beginning of the actor can fail.  What do we want to happen in that case:

- task stays NOT_STARTED forever, or
- task stays NOT_STARTED, our scheduled cleanup task later sees a task which has been NOT_STARTED for e.g. one day and assumes it will therefore never be run, so marks it FAILED, or
- task stays NOT_STARTED until a retry a minute later or so which manages to then process the task as normal

To me the third option seems best, so I think high-level retries still makes sense here.

However since the batch writes themselves do lower-level retries, it's also reasonable to say "those retries are sufficient here" and wrap that part of the actor in a try/except block which doesn't retry if *those* steps failed, since we know any failure already comes after several retries.

exodus_gw/worker/publish.py
rohanpm: Yeah, I think so.  Setting state to FAILED, then raising, then expecting the actor to be retried and find that the task is FAILED and just return, would technically work.  But it's more complicated than just setting state to FAILED and then returning, for no clear reason.

exodus_gw/worker/publish.py
nathanegillett: Okay. Actually, I think just having the `write_batches` calls in the try/except block as they are now will achieve this. However, `write_batches` will return True if it successfully deletes too. I don't think we want to call the task "COMPLETE" in those cases. So, I've added a couple more state changes.

exodus_gw/worker/publish.py
```diff
@@ -48,19 +47,25 @@ def commit(publish_id: str, env: str):
...
-            # Delete everything if failed to write last_items.
-            write_batches(env, items + last_items, delete=True)
+        if not items_written or not last_items_written:
+            items = items + last_items if last_items else items
+            write_batches(env, items, delete=True)
+            # Change task state to FAILED.
+            task.state = schemas.TaskStates.failed
+            db.commit()
+    except Exception:
+        LOG.exception("Task %s encountered an error", str(task.id))
```
rohanpm, line 64: Minor: %s already implies "convert to string", so the str() should be redundant.

exodus_gw/worker/publish.py
```diff
@@ -48,19 +47,25 @@ def commit(publish_id: str, env: str):
...
 
-        if not last_items_written:
-            # Delete everything if failed to write last_items.
-            write_batches(env, items + last_items, delete=True)
+        if not items_written or not last_items_written:
+            items = items + last_items if last_items else items
+            write_batches(env, items, delete=True)
+            # Change task state to FAILED.
+            task.state = schemas.TaskStates.failed
+            db.commit()
```
rohanpm, line 62: I think we are missing a 'return' here as well, aren't we?  Otherwise it will continue on after the 'except' block and then immediately set the task state back to COMPLETE.

If I'm right and this is buggy then it would also imply that the tests aren't checking the state updates on the task object very well.

rohanpm (changes requested):

exodus_gw/worker/publish.py
nathanegillett: You're right. No checks on the task's end state were made in test_actors.py. Fixed.

crungehottman (approved):

rohanpm (approved):

PR #198, opened 2021-02-17
Fix pytest fixture naming

nathanegillett:
Some pytest fixtures were named with "mock" even though they weren't
Mock objects, so change the names for accuracy.


rohanpm (approved):
This is definitely an improvement.  It's still a little broken that "fake_publish" returns a *model* object and "fake_item_list" returns *schema* objects - how is anyone meant to remember that?

nathanegillett:
> It's still a little broken that "fake_publish" returns a _model_ object and "fake_item_list" returns _schema_ objects

Hmm. As far as I can tell, that item list should hold models, not schemas, as it's being fed to `update_publish_items`. I think this is a leftover from schema/model confusion a while back.

rohanpm (approved):

PR #199, opened 2021-02-17
Fix usage of blocking DB session from async functions [RHELDST-4857]

rohanpm:
sqlalchemy's session is not designed for use with async. Using any
of its blocking methods from within an async function could block
the entire service for the duration (e.g. if a db.commit() in one
request were to take five seconds, the service would effectively
hang and not process any other requests for those five seconds).

Fix it so that usage of the session never blocks the main thread.
Achieved in two ways:

- for endpoints using the DB, they simply shouldn't be defined with
  async. This will make FastAPI run them via a threadpool.

- for the DB usage within middleware, which has to be async,
  explicitly call the close/commit functions via a threadpool.

This kind of bug would be easy to reintroduce, as the symptoms
are only some mysterious slowness which is probably too slight
to notice during casual testing. To help protect against further
mistakes, a fixture was added which wraps DB sessions during tests
using TestClient. The wrapper will enforce that the DB session
is not being used from the main thread.

### DIFF
--- a/exodus_gw/main.py
+++ b/exodus_gw/main.py
@@ -67,6 +67,7 @@ import dramatiq
 from fastapi import FastAPI, Request
 from fastapi.exception_handlers import http_exception_handler
 from sqlalchemy.orm import Session
+from starlette.concurrency import run_in_threadpool
 from starlette.exceptions import HTTPException as StarletteHTTPException
 
 from .aws.util import xml_response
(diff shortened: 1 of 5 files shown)


rohanpm:
https://github.com/release-engineering/exodus-gw/pull/200 is related.  This PR removes async where it should not be, the other PR adds async where it was missing.

nathanegillett (approved):

crungehottman (approved):

PR #200, opened 2021-02-17
Add missing 'async' in various places, reduce unneeded threads [RHELDST-4857]

rohanpm:
In the case of both endpoints, and functions used with Depends, if they
don't use any blocking functionality then we should mark them as
async.

We should do this because otherwise FastAPI is pessimistic: it assumes
any non-async function might block, and so runs it in a threadpool.
This works but is rather wasteful to do this for every dependency
when not needed.

### DIFF
--- a/exodus_gw/auth.py
+++ b/exodus_gw/auth.py
@@ -36,7 +36,7 @@ class CallContext(BaseModel):
     user: UserContext = UserContext()
 
 
-def call_context(request: Request) -> CallContext:
+async def call_context(request: Request) -> CallContext:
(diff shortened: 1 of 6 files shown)


rohanpm:
Related to https://github.com/release-engineering/exodus-gw/pull/199.  They're kind of opposites, one being removing async where it shouldn't be and another being to add it where it was missing.

Unfortunately for this one, unlike #199, I can't come up with any way to enforce from the tests that we don't make similar mistakes again.

nathanegillett (approved):

crungehottman (approved):

PR #201, opened 2021-02-17
examples: start using https & certs in exodus-sync

rohanpm:
Since we have resolved the problem with https requests hanging in
boto, we should now get into the habit of using https and client
certs/keys by default. Let's also exercise a few of the other APIs
in this example, particularly 'whoami' so the cert/key can be
easily confirmed as working (or not).

Right now, this doesn't really affect anything, but we will soon
start enforcing appropriate roles on each endpoint. Once that
happens, usage of https will become more or less mandatory.

### DIFF
--- a/examples/exodus-sync
+++ b/examples/exodus-sync
@@ -30,6 +30,7 @@ import backoff
 import boto3
 import boto3.session
 import requests
+from botocore.config import Config
 from botocore.exceptions import ClientError
 
 # How many items we'll add to a publish per request.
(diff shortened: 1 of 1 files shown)


examples/exodus-sync
```diff
@@ -202,20 +210,79 @@ def publish_items(args, items):
...
     parser.add_argument(
         "--debug", action="store_true", help="Enable verbose logging"
     )
 
-    # TODO: we should point this at https by default, but that won't work
-    # well until RHELDST-3478 is fixed
-    parser.add_argument("--exodus-gw-url", default="http://localhost:8000")
+    parser.add_argument(
+        "--cert",
+        default=os.path.expandvars("${HOME}/certs/${USER}.crt"),
```
rohanpm, line 277: These paths are essentially arbitrary, AFAIK there is no standard for this kind of thing.

There is a bit of a convention in some existing tools (rhsm-tools) for the ~/certs directory at least, but not for user certs, so I've just made this up.

nathanegillett (approved):

crungehottman (approved):

PR #202, opened 2021-02-17
Add state field to publishes [RHELDST-5083]

nathanegillett:
Publish states are now updated at appropriate points in the commit
process and are checked before adding items to or committing publishes.

The get_publish_by_id CRUD function was removed in favor of directly
querying as needed.

Additionally, a bug wherein delete requests to DynamoDB contained an
invalid number of keys was found and fixed.


exodus_gw/models/service.py
```diff
@@ -10,4 +10,4 @@ class Task(Base):
 
     id = Column(UUID(as_uuid=True), primary_key=True)
     publish_id = Column(UUID(as_uuid=True), nullable=False)
-    state = Column(String)
+    state = Column(String, nullable=False)
```
nathanegillett, line 13: I've added `nullable=False` back here because migrations knew nothing of its removal, resulting in some testing issues. I think it should be required, anyway. A Task (or Publish) without a state is DOA now that we're checking for them when committing (and/or updating).

nathanegillett:
@rohanpm, I have no clue as to why the testing db is broken here. Locally, I have one test failing due to a similar issue which I had planned to ask about here but it's now obscured by all of this.
https://travis-ci.com/github/release-engineering/exodus-gw/jobs/483997603

nathanegillett (commented):
@rohanpm, I have no clue as to why the testing db is broken here. Locally, I have one test failing due to a similar issue which I had planned to ask about here but it's now obscured by all of this.
https://travis-ci.com/github/release-engineering/exodus-gw/jobs/483997603

It's complaining about a null value for Publish.state but I don't see where it could be missing.

rohanpm:
> @rohanpm, I have no clue as to why the testing db is broken here. Locally, I have one test failing due to a similar issue which I had planned to ask about here but it's now obscured by all of this.
> https://travis-ci.com/github/release-engineering/exodus-gw/jobs/483997603
> 
> It's complaining about a null value for Publish.state but I don't see where it could be missing.

For "sqlite3.OperationalError: Cannot add a NOT NULL column with default value NULL", I discovered the same thing on another PR of mine.  If you have a look at exodus_gw/migrations/versions/a60131dd10c4_.py in latest master, the solution is there: you have to adjust your migration to use a 'batch op' and to set recreate='always'.

The background is that sqlite versions older than 3.32.0 have a weird limitation whereby you aren't allowed to add a NOT NULL column with no default value using the ALTER COLUMN statement.  This is despite the fact that there shouldn't be any problem with this if the table is empty, and that there's no problem defining a column in this way when the table is initially created.

batch op and recreate fixes it because this tells sqlalchemy to add the column in a different way, essentially creating an entirely new table and then renaming it over the old one.

The sqlite limitation was removed in 3.32.0, and an up-to-date Fedora env will have a version newer than that so it can't be reproduced locally, but I guess the Travis environment has something older.

nathanegillett:
> [...] batch op and recreate fixes it [...]

Thanks, @rohanpm.

exodus_gw/worker/publish.py
```diff
@@ -41,8 +53,8 @@ def commit(publish_id: str, env: str):
     items_written = False
     last_items_written = False
 
-    # Change task state to IN_PROGRESS.
     task.state = schemas.TaskStates.in_progress
+    publish.state = schemas.PublishStates.committing
```
rohanpm, line 57: I think this needs to happen in the web endpoint, commit_publish, not here.  Also with some locking.

If done here there are many scenarios where concurrent requests can mess it up, e.g.:

- commit_publish is invoked N times before this actor manages to set the state to COMMITTING.  With current code that would result in N tasks & calls to the actor being queued.  We want a maximum of 1.
- commit_publish is invoked, then update_publish_items is invoked after this task is enqueued but before it got up to here.  We'd want the update_publish_items to be rejected at that point.

I think the way it ought to work is something like this:

- in commit_publish, when you query for the publish object, that should be a "SELECT ... FOR UPDATE" query.  That can be achieved by adding a 'with_for_update' to the query call.  The point of this is that it both reads the object and obtains a lock on it for the current transaction - so you know nobody else is in the middle of code to update the state.
- set the publish to COMMITTING within the same transaction as sending the commit message to the actor.
- update_publish_items needs to use "SELECT ... FOR UPDATE" as well.  This is how we can guarantee that, for example, we don't have one request starting to commit the publish at exactly the same time as another request is adding more publish items onto it.

rohanpm (changes requested):

rohanpm (approved):

PR #208, opened 2021-02-22
Add UTC timestamp to publish and task objects [RHELDST-5084]

nathanegillett:
Timestamps are automatically updated when updates are made to the
respective object's columns.


nathanegillett:
@rohanpm, migration has broken somehow.

> sqlalchemy.exc.InternalError: (psycopg2.errors.DependentObjectsStillExist) cannot drop constraint publishes_pkey on table publishes because other objects depend on it
DETAIL:  constraint items_publish_id_fkey on table items depends on index publishes_pkey
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
>
> [SQL: ALTER TABLE publishes DROP CONSTRAINT publishes_pkey]
(Background on this error at: http://sqlalche.me/e/13/2j85)
ERROR: InvocationError for command /home/ngillett/src/exodus-gw/.tox/alembic-autogen/bin/python scripts/alembic-autogen (exited with code 1)

Looks like `meta.drop_all` in `db_reset` from migrate.py is trying to drop the publishes table before the items table without using `CASCADE`.

This problem is present on the master branch as well.
I've tried a few different things, altering `db_reset`, adding cascade options in various places, etc., with no luck.

exodus_gw/models/publish.py
```diff
@@ -1,7 +1,7 @@
 import uuid
 
 from sqlalchemy import Column, ForeignKey, String
-from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.dialects.postgresql import TIMESTAMP, UUID
```
rohanpm, line 4: Hmm, rather than using this type from postgresql dialect, can we just use the generic DateTime type which can be imported from sqlalchemy module?

From a quick reading of the doc string, the differences are probably something like:

- TIMESTAMP might support fractional seconds
- TIMESTAMP might support storing timezones

I don't think we need either of these, so it'd be preferable to stick with the generic type.  Note: I said something before about handling datetimes as UTC, but this doesn't mean we have to store timezone info in the DB (we would only have to do that if we wanted to store values using multiple different timezones).

exodus_gw/routers/publish.py
```diff
@@ -207,11 +223,13 @@ def commit_publish(
     msg = worker.commit.send(publish_id=str(db_publish.id), env=env.name)
     LOG.info("Enqueued commit for '%s'", msg.kwargs["publish_id"])
     db_publish.state = schemas.PublishStates.committing
+    db_publish.updated = datetime.now(timezone.utc)
 
     task = models.Task(
         id=msg.message_id,
         publish_id=msg.kwargs["publish_id"],
         state="NOT_STARTED",
+        updated=datetime.now(timezone.utc),
```
rohanpm, line 232: I think it'd be preferable to use exactly the same timestamp here as on the publish object as well, rather than possibly allowing them to be different.

Similar to the comment I made on the other PR about advantages of having "from_date" the same across an entire publish.  Even though either way works, it just makes it a bit easier to identify related data if they've got the same timestamp.

exodus_gw/worker/publish.py
```diff
@@ -13,6 +14,28 @@
 LOG = logging.getLogger("exodus-gw")
 
 
+def update_state(
```
rohanpm, line 17: I guess the intent of this is to help ensure that you won't forget to update timestamps when changing states.

There's another more direct way you might consider looking into.  sqlalchemy has an events system which allows to run bits of code when certain things happen.  If we want to implement e.g. "Whenever a Task is modified, its timestamp is updated" then I believe dropping something like this underneath the 'class Task' may do the trick:

```python
@event.listens_for(Task, 'before_update')
def task_before_update(_mapper, _connection, task):
    task.updated = datetime.utcnow()
```

So you might want to consider it in case it helps remove the need for a helper like this.

exodus_gw/worker/publish.py
```diff
@@ -13,6 +14,28 @@
...
+    if action == "start":
+        task.state = schemas.TaskStates.in_progress
+    elif action == "fail":
+        task.state = schemas.TaskStates.failed
+        publish.state = schemas.PublishStates.failed
+    elif action == "finish":
+        task.state = schemas.TaskStates.complete
+        publish.state = schemas.PublishStates.committed
+    else:
+        LOG.error("Unknown action, '%s'", action)
```
rohanpm, line 31: This would be just a bug in our code, it shouldn't be tolerated by returning since it means we can write code passing the wrong value to update_state and it will seem to work.  In a case like this I'd suggest raising an AssertionError, as there is just no legitimate way to get here, if we do then it's most likely a bug in the code the dev is currently working on and it should be detected ASAP rather than ignored.

tests/conftest.py
```diff
@@ -47,6 +48,14 @@ def mock_db_session():
     yield db_session
 
 
+@pytest.fixture()
+def mock_datetime(autouse=True):
```
rohanpm, line 52: To me this is a little too severe to enable as an autouse fixture, as it means every piece of code will always get the same value for the current datetime which is very unrealistic.

The way it is mocked, by patching datetime.datetime, is also very severe as it means even if you explicitly try to create a datetime.datetime of a particular year, month, day etc, you won't get it - instead you'll just get this hardcoded value.  This would apply across every library in use.

What's the goal here, and is there another way we can do it which won't make datetime.datetime behave so oddly across all tests?

rohanpm (changes requested):

exodus_gw/models/publish.py
nathanegillett: Sure. I just thought it would help anyone looking at these figure them out but I suppose it's not meant for human consumption anyway.

tests/conftest.py
nathanegillett: Thought it would be convenient to have datetimes in tested code return a static value, so we can assert about it, but we can just mock it as needed instead.

exodus_gw/worker/publish.py
nathanegillett: This is exactly what I was looking for. Thanks!

crungehottman (approved):

exodus_gw/migrations/versions/54f2120b9ab6_.py
```diff
@@ -0,0 +1,55 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "54f2120b9ab6"
+down_revision = "854e06069e65"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # clean all publishes and tasks first to avoid crashing on null state
```
rohanpm, line 21: This all looks wrong as the columns have been defined with nullable=True.

Which is it - are they intended to be nullable or not?  If nullable then why would it be needed to delete the current content of the tables and how can the comment about sqlite "NOT NULL column" be correct?

tests/test_models.py
```diff
@@ -12,3 +18,50 @@ def test_Item_init():
...
+        # Change state of publish and task
+        publish.state = "COMMITTING"
+        task.state = "IN_PROGRESS"
+        db.commit()
+
+        # Updated should now hold datetime objects
+        p_updated = publish.updated
+        assert isinstance(p_updated, datetime)
+
+        t_updated = publish.updated
```
rohanpm, line 54:
```suggestion
        t_updated = task.updated
```

rohanpm (changes requested):

exodus_gw/migrations/versions/54f2120b9ab6_.py
nathanegillett: Seems like emptying the tables wasn't necessary.

exodus_gw/migrations/versions/54f2120b9ab6_.py
```diff
@@ -0,0 +1,51 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "54f2120b9ab6"
+down_revision = "854e06069e65"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # recreate='always' to avoid "Cannot add a NOT NULL column with default value NULL"
```
rohanpm, line 21: This should still be wrong though, it's specifically about "NOT NULL" columns, and the columns being touched here aren't NOT NULL.  Can this be dropped too please and just don't pass a 'recreate' value if we don't have to?

(It might not even be required to use a batch op at all here, but I think it's reasonable to just use them by default anyway to keep all changes to a table grouped.)

rohanpm (changes requested):

exodus_gw/migrations/versions/54f2120b9ab6_.py
nathanegillett: Yeah, I just had another, closer look at the error raised by the autogen migration version. The stock upgrade function should be fine but the downgrade function is what needed some attention.

rohanpm (approved):

PR #210, opened 2021-02-22
Fix broken migrations - 'DependentObjectsStillExist'

nathanegillett:
This was caused by migration version 854e06069e65 in which
batch_alter_table() in upgrade() used recreate="always". This forced a
drop of the publishes table which failed because Item objects still
existed which depended on Publish ids. This commit removes the recreate
argument.

### DIFF
--- a/exodus_gw/migrations/versions/854e06069e65_.py
+++ b/exodus_gw/migrations/versions/854e06069e65_.py
@@ -21,9 +21,6 @@ def upgrade():
 
     with op.batch_alter_table(
         "publishes",
-        # recreate='always' to avoid "Cannot add a NOT NULL column with default value NULL"
-        # on sqlite < 3.32.0
-        recreate="always",
     ) as batch_op:
         batch_op.add_column(
             sa.Column(


nathanegillett:
Some background on the issue addressed by this PR: https://github.com/release-engineering/exodus-gw/pull/208#issuecomment-783481449

Also, I can't seem to get a recent version of sqlite on our builds. Travis doesn't support the only Ubuntu release that has access to any sqlite3 version >= 3.32, and my attempts at installing sqlite3 from source in the job weren't successful in affecting the tests.

I guess using this solution to the migrations problem means dealing with failing auto-tests for as long as we use Travis CI.

rohanpm (changes requested):
I don't think this is the correct fix.  But it might help to know a bit more how the problem can be reproduced.

Let me explain a bit more about these kinds of migrations:

This migration tries to add a new 'state' column to an existing table, and the column has a not-null constraint.

This raises an issue: what if that table already has some rows?  The 'state' column is not allowed to be null, so it's going to need some real value filled in.

If you're doing migrations 100% properly and aiming for no data loss, in such cases you have to do something like:

- first introduce the column *without* "not-null" constraint
- then iterate over all rows, calculate and write some appropriate value for the field
- then finally set the not-null constraint

However for this service, both because it's not in production yet but also due to the design of the service, data loss is acceptable.  So the simpler approach of "just delete the data I'm about to change" will sometimes be acceptable.

So that's why it's reasonable to "DELETE FROM publishes", but since there's a related table, I suppose the delete isn't going far enough.  What if we delete from the related table as well (first), does that fix it?

In other words just before `op.execute("DELETE FROM publishes")`, add an `op.execute("DELETE FROM items")` ?

rohanpm:
> In other words just before `op.execute("DELETE FROM publishes")`, add an `op.execute("DELETE FROM items")` ?

Having hit this now locally and reviewing the backtrace, I don't think the above will work because the error doesn't seem to be related to any *data*, but rather a constraint.

The current solution also wouldn't be acceptable though, we can't leave Travis broken like that.

rohanpm:
The fix in https://github.com/release-engineering/exodus-gw/pull/214 ought to work in all environments.

PR #211, opened 2021-02-22
Remove from_date Item attribute [RHELDST-4868]

nathanegillett:
This commit removes the from_date attribute from the Item model and
schema so users are unable to set this field themselves, as there is
currently no use-case for them to do so.

The from_date attribute is still required by the DynamoDB table,
however, so it is programmatically set to the datetime at which
the item's parent publish is committed.

### DIFF
--- a/exodus_gw/aws/dynamodb.py
+++ b/exodus_gw/aws/dynamodb.py
@@ -40,26 +40,47 @@ def batch_write(env: str, request: Dict[str, Any]):
     return response
 
 
-def create_request(env: str, items: List[models.Item], delete: bool = False):
+def create_request(
+    env: str, items: List[models.Item], from_date: str, delete: bool = False
+):
     """Create the dictionary structure expected by batch_write_item."""
 
-    table = get_environment(env).table
+    table_name = get_environment(env).table
 
     if delete:
-        req_type = "DeleteRequest"
-        item_type = "Key"
-    else:
-        req_type = "PutRequest"
-        item_type = "Item"
+        return {
+            table_name: [
+                {
+                    "DeleteRequest": {
+                        "Key": {
+                            "from_date": {"S": from_date},
+                            "web_uri": {"S": item.web_uri},
+                        }
+                    }
+                }
+                for item in items
+            ]
+        }
 
     return {
-        table: [
-            {req_type: {item_type: item.aws_fmt(delete)}} for item in items
+        table_name: [
+            {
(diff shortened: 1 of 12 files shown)


exodus_gw/aws/dynamodb.py
```diff
@@ -43,18 +43,32 @@ def batch_write(env: str, request: Dict[str, Any]):
...
-        table: [
-            {req_type: {item_type: item.aws_fmt(delete)}} for item in items
+        get_environment(env).table: [
+            {
+                req_type: {
+                    key_type: {
+                        # While required, the from_date key should not
+                        # be set by client users until a use-case for
+                        # them to do so emerges.
+                        "from_date": {"S": datetime.now(timezone.utc)},
```
rohanpm, line 65: I don't think this is the most appropriate place to initialize the value, mainly because it means every item or batch of items within the same publish can have a different from_date.

I think it's quite advantageous to have one from_date used for the entire publish.  Think of the scenario for instance of getting a ticket "hey, this one publish messed up a bunch of stuff, can you manually undo it somehow?"  The request is basically impossible if exodus-gw used a range of from_date values during the publish, but if it used the same value for all items, and that value was logged somewhere, then we have the ability to look up or delete all the items it wrote.

I suggest passing from_date as an argument into the dramatiq actor.  This also has another effect which I think is positive: in the event of a conflict, whichever commit is requested later "wins", even if some writes are delayed due to retries etc.  I think this is easier to reason about than having whichever write succeeds last "wins".

Also, unless I'm missing something, this is going to completely break deletes, since datetime.now at time of delete is not necessarily going to be the same as it was at the time of put.

rohanpm (changes requested):

nathanegillett:
Based on #208 to avoid migration conflicts. Holding as draft until parent commit is merged.

rohanpm (approved):

rohanpm:
Looked right, but it failed CI.  One of the revisions within the alembic migration might be wrong.

nathanegillett:
> Looked right, but it failed CI. One of the revisions within the alembic migration might be wrong.

Yep. I remade the revision for #208 so the revision from this commit was looking for a non-existent down_revision.
Fixed.

rohanpm (approved):

rohanpm:
I want to submit this because otherwise I still need to write "from_date" in some new autotests only to remove it later.

PR #216, opened 2021-02-25
Implement scheduled cleanup task [RHELDST-4843]

rohanpm:
Previously we had a stubbed out scheduled cleanup task, as we
needed some more columns on the model before it could be
implemented.

As those are now available, let's fill in the implementation of
the cleanup task. It covers steps:

- find & fix anything with missing timestamps
- mark anything which is obviously abandoned or stuck as failed
  (for 2 days by default)
- delete anything which has been in a terminal state for long
  enough (2 weeks by default)

The thresholds are configurable via settings.

### DIFF
--- a/exodus_gw/models/publish.py
+++ b/exodus_gw/models/publish.py
@@ -20,7 +20,9 @@ class Publish(Base):
     env = Column(String, nullable=False)
     state = Column(String, nullable=False)
     updated = Column(DateTime(timezone=True))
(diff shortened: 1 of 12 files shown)


rohanpm:
Note this pull request includes a refactor of "db" fixture, this is done in its own commit so it can be reviewed separately if desired.

nathanegillett (approved):

PR #217, opened 2021-02-25
Require authorization for upload and publish APIs [RHELDST-4729]

nathanegillett:
Users must now possess the appropriate role to request uploads or
publishes to a particular environment, e.g., "stage-blob-uploader" to
upload to the "stage" environment.

The development environment was updated to automatically authorizate
users for the "test" environment.

### DIFF
--- a/exodus_gw/auth.py
+++ b/exodus_gw/auth.py
@@ -78,10 +78,14 @@ def needs_role(rolename):
     >    "If caller does not have role xyz, they will never get here."
     """
 
-    async def check_roles(roles: Set[str] = Depends(caller_roles)):
-        if rolename not in roles:
+    async def check_roles(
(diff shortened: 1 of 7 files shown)


rohanpm:
Looks good to me, the only missing thing is docs:

Could we please have doc strings updated so that the endpoints which require authorization have the needed roles showing at https://release-engineering.github.io/exodus-gw/api.html ?

My suggestion for how to do this is to have it as the second paragraph of the doc string on the relevant endpoints, in a consistent style.  So, for example, update_publish_items doc string might look like:

```
Add publish items to an existing publish object.

**Required roles**: `{env}-publisher`

Publish items primarily are a mapping between a URI relative to the root of the CDN,
and the key of a binary object which should be exposed from that URI.
(...)
```

crungehottman (approved):

rohanpm (approved):

PR #238, opened 2021-03-11
Add auth support for s3-upload

dichen16:
It requires you to have a valid certificate and key produced by RHCS
for calling example/s3-upload.

### DIFF
--- a/examples/s3-upload
+++ b/examples/s3-upload
@@ -22,9 +22,11 @@
 import argparse
 import hashlib
 import logging
+import os
 import sys
 
 import boto3
+from botocore.config import Config
 
 
 def get_object_key(filename):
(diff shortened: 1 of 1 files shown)


examples/s3-upload
```diff
@@ -13,7 +13,7 @@
 #
 #  # Testing same thing via exodus-gw
 #  uvicorn exodus_gw.gateway:app &
-#  examples/s3-upload --endpoint-url http://localhost:8000/upload file1 [file2 [...]]
+#  examples/s3-upload --endpoint-url http://localhost:8000/upload --key my.key --cert my.cert file1 [file2 [...]]
```
rohanpm, line 16: Personally I would just leave this out of the example, since there are good defaults and we should encourage people to use them.

rohanpm (approved):

rohanpm (approved):

nathanegillett (approved):

crungehottman (approved):

PR #239, opened 2021-03-11
Set API as stable

rohanpm:
I think we've reached the point that we should commit to a stable
API, so let's set the version to 1.0.0 and remove the warning from
docs. The API has been stable for a little while now, and since
we now have code in exodus-rsync making use of it, it's time to
start taking backwards-compatibility seriously.


dichen16 (approved):
LGTM

nathanegillett (approved):

crungehottman (approved):

PR #240, opened 2021-03-11
Move CI to Github actions from Travis.ci

nathanegillett:


scripts/push-docs
```diff
@@ -5,7 +5,7 @@ ORG=release-engineering
 REPO=exodus-gw
 
 enabled(){
-  if [ "$TRAVIS_BRANCH" != "master" ] || [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+  if [ "$GITHUB_BASE_REF" != "master" ] || [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
```
nathanegillett, line 8: @rohanpm, I'm a little worried about this, as GITHUB_BASE_REF is only set on pull request events. But I'm also a little confused about when this script actually pushes the docs and the actual goal of this function.
Does this script run on merges? Should we be testing the name of the feature branch or the base branch here?

rohanpm: Travis runs the same jobs both when testing a pull request pre-merge, and also testing master branch *after* merge.  The goal here is to make sure that we *don't* push docs in the pre-merge tests (because then the published docs would just be randomly flipping between whatever PRs happened to be tested most recently rather than showing docs for the latest official version).

To make this work with actions, firstly I see that in the workflow we currently trigger on pull_request only and I think that should be changed to also trigger on "push" event (to master).

When that's done then probably the check here could become: `[ "$GITHUB_EVENT_NAME" != "pull_request" ]` and drop the GITHUB_BASE_REF part.

nathanegillett: > Travis runs the same jobs both when testing a pull request pre-merge, and also testing master branch after merge.

I thought this was the case but couldn't find evidence.

Anyway, thanks. I've made the changes you mention.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
+name: Continuous Integration
+
+on:
+  pull_request:
+    branches: [master]
+  push:
+    branches: [master]
+
+jobs:
```
nathanegillett, line 9: Note that I've intentionally left out the docker build job, as I don't believe the resulting image was serving any purpose.

rohanpm: The purpose of it was just to test that the image build works.  This is how, if a PR makes changes to Dockerfile, we can get some assurance that the image is able to build before merging the PR.  It was added after one PR broke the build of container images without being detected by CI.

I think it should be retained if practical.

af452cce83bfa578 is an example of a commit which becomes essentially untested if the CI isn't trying to build images.

scripts/push-docs
```diff
@@ -5,7 +5,7 @@ ORG=release-engineering
 REPO=exodus-gw
 
 enabled(){
-  if [ "$TRAVIS_BRANCH" != "master" ] || [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+  if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
     return 1
```
rohanpm, line 9: (just noting for myself that 0 means true in shell, opposite of most languages... I think I got confused and wrote `!=` in an earlier comment where it should have been `==`.)

.github/workflows/ci.yml
nathanegillett: > I think it should be retained if practical.

Done

rohanpm (approved):

dichen16 (approved):

PR #241, opened 2021-03-12
Replace push-docs' GITHUB_TOKEN with access token [RHELDST-5390]

nathanegillett:
The push-docs script, when passed 'github.token', reports "No
such device or address". This commit replaces the token with a personal
access token, which was the method used with Travis.ci.


nathanegillett:
Will merge this upon passing checks, without approval, as I'm troubleshooting push-docs. Will revert if unsuccessful.

PR #345, opened 2021-06-30
chore: update pip-compile workflow


rohanpm:
## Automated update

This is an automated update generated by ansible playbooks.
See [the playbook repo](https://url.corp.redhat.com/300a997) for more information.

### update pip-compile workflow

This GitHub Actions workflow will regularly invoke pip-compile over relevant files in this repo and submit pull requests when python dependencies need updating.

### DIFF
--- a/.github/workflows/pip-compile-merge.yml
+++ b/.github/workflows/pip-compile-merge.yml
@@ -1,0 +1,23 @@
+on:
+  workflow_dispatch: {}
+  schedule:
+  - cron: "20 20 * * 2"
+
+name: "pip-compile: merge PR"
+jobs:
+ pip_compile_merge:
+   name: pip-compile automerge
+   runs-on: ubuntu-latest
+   steps:
+    - name: Look up pull request
+      uses: juliangruber/find-pull-request-action@v1
+      id: find-pull-request
+      with:
+        branch: deps/pip-compile
+    - name: Merge Pull Request
+      uses: juliangruber/merge-pull-request-action@v1
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        number: ${{ steps.find-pull-request.outputs.number }}
+        method: merge
+      if: ${{ steps.find-pull-request.outputs.number }}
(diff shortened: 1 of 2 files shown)


rohanpm:
This might take a bit of explaining since I'm doing two experiments at once here...

### 1. ansible

As seen in the PR description, this change was created by ansible and not by myself directly.  This is just something I'm gradually putting together because I feel like the config in our repos has become complex enough and inconsistent enough that I need some better way to manage it and to push out changes to multiple repos at once.

See for example that the equivalent PR was created at https://github.com/release-engineering/pushsource/pull/153 .  But you can just review it as if I'd created it manually.

### 2. scheduled pip-compile

As for what the change is actually doing... the intent is to try out a new way of updating python deps in our repos.  Instead of relying on dependabot or a similar service like renovatebot, we "just" do the following:

- once a week, run pip-compile and then create a PR with whatever changed (if anything).
- a couple days later, submit it, if checks have passed.

So, if it's working well, there will be max of one "update dependencies" PR per week and it'll be merged without any human interaction.  Why I think this is worthwhile is due to a few problems with both dependabot and renovatebot, neither one seems to do the job quite right.

dependabot issues:
- no way to batch multiple updates together, so it seems too noisy to me
- no way to control the environment, so if you need extra native packages you're out of luck (e.g. it doesn't work in repos depending on rpm-py-installer)

renovatebot issues:
- it doesn't do pip-compatible dependency resolution. This means it creates updates with unresolvable situations (e.g. it may try to update requirements.txt to request pkg1==1.0.0 and pkg2==2.0.0, where pkg1 has a dependency of 'pkg2<2.0' so the combination is uninstallable)
- same issue with lack of control over environment

Since those services don't quite seem to do the trick, I want to try out this simpler approach.  It's crude but hopefully effective.  I've tried it in my own forks for a couple of weeks now and it seems to be working OK, e.g. an example at https://github.com/rohanpm/pushsource/pull/9 .

@nathanegillett I thought exodus-gw might be a good candidate for trying this, since you expressed interest in the dependency management stuff before and have been taking care of most of the dependabot PRs.

nathanegillett (approved):

PR #353, opened 2021-07-06
Fix CI issues with scheduled pip-compile

rohanpm:
Addresses the following issues in scheduled pip-compile (see also https://github.com/release-engineering/exodus-gw/commit/ce0e58a1a2a5d9976f0ef2e193de09c1ee58d94e for more explanation):

- when compiling test-requirements.txt, we really need to use both requirements.in and test-requirements.in together.  It's needed to ensure that the generated requirements end up being compatible with *both* of the inputs.
- we need to use a non-default token when creating the PR.  If not, it runs into the rule from https://docs.github.com/en/actions/reference/events-that-trigger-workflows: "When you use the repository's GITHUB_TOKEN to perform tasks on behalf of the GitHub Actions app, events triggered by the GITHUB_TOKEN will not create a new workflow run".  That means, if the repo's CI is based on actions (as it is here), then CI will never run over the pull request.
    - we now use a token associated with @cd-red-bot .

### DIFF
--- a/.github/workflows/pip-compile.yml
+++ b/.github/workflows/pip-compile.yml
@@ -22,11 +22,11 @@ jobs:
          EXECUTE_COMMANDS: |
            pip install pip-tools
            pip-compile -U --generate-hashes requirements.in
-           pip-compile -U --generate-hashes test-requirements.in
-           
+           pip-compile -U --generate-hashes requirements.in test-requirements.in -o test-requirements.txt
          COMMIT_MESSAGE: 'chore: scheduled pip-compile'
          COMMIT_NAME: 'GitHub Actions'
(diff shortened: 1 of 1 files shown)


crungehottman (approved):

nathanegillett (approved):
Do we not need to compile requirements.txt as well?

rohanpm:
> Do we not need to compile requirements.txt as well?

We do, that's still in there:

> pip-compile -U --generate-hashes requirements.in

This compiles requirements.txt (or to use more precise language... it compiles requirements.in and produces requirements.txt as output).

It might not be as obvious because pip-compile default behavior is to calculate the output .txt file based on the .in file, and that's used here, but then not on the next line for test-requirements.txt which has to be written out explicitly.

PR #356, opened 2021-07-06
Fix dependency mismatch breaking CI

rohanpm:
Automated pip-compile PR release-engineering/exodus-gw#347 had a few issues:

- outputs included idna 3.x in requirements.txt and idna 2.x in
  test-requirements.txt. This is broken because both need to be
  installable at the same time.

- It would have failed CI - but it wasn't covered by CI since one github
  action can't trigger anotheranother (normally).

- It was submitted despite lack of CI due to missing branch protection
  rules (now fixed).

The idna problem was triggered by:

- idna is in the dependency tree for both requirements.in and
  test-requirements.in

- requests library requires idna<3

- requests is in dependency tree for test-requirements.in only

- therefore, compiling requirements.in will get you idna 3 while
  compiling test-requirements.in will get you idna 2

requests & idna are one trigger for the problem, but in theory any
packages could get mismatches in this way.

It seems the approach to running pip-compile needs some rethinking:
we probably need to give both the .in files when compiling
test-requirements.txt, and in some cases might need to add some
constraints into requirements.in to make sure we can test and
release with the same dependencies.

A permanent fix would have to refactor the scheduled workflows.
For now, I've added an idna constraint and manually pip-compiled
to unbreak the repo.


nathanegillett (approved):

rohanpm:
Please ignore the title & body change in this PR from a couple of minutes ago.  That's apparently a bug in the repo update playbooks I've been testing.

The actual content of the PR didn't change since it was tested and approved (the commits).

PR #377, opened 2021-11-03
Accept publish items with link_to or object_key [RHELDST-8586]

nathanegillett:
Previously, exodus-gw accepted items that had a web_uri and a valid
object_key. Soon exodus-rsync will send items with a link_to value
instead of an object_key and the resolution of the object_key will be
handled within exodus-gw based on the link_to path. With this commit
exodus-gw accepts items that have either a link_to or an object_key.

### DIFF
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -59,6 +59,7 @@ Publish objects should be treated as ephemeral; they are not persisted indefinit
 """
 
 import logging
+import os
 from datetime import datetime, timezone
 from typing import Dict, List, Union
 from uuid import UUID, uuid4
@@ -123,7 +124,7 @@ def publish(
     dependencies=[auth.needs_role("publisher")],
 )
 def update_publish_items(
-    items: Union[schemas.ItemBase, List[schemas.ItemBase]] = Body(
+    items: Union[List[schemas.ItemBase], schemas.ItemBase] = Body(
         ...,
         example=[
             {
@@ -133,6 +134,10 @@ def update_publish_items(
             {
                 "web_uri": "/my/slightly-less-awesome/other-file.iso",
                 "object_key": "c06545d4e1a1c8e221d47e7d568c035fb32c6b6124881fd0bc17983bd9088ae0",
+            },
+            {
+                "web_uri": "/another/route/to/my/awsome/file.iso",
+                "link_to": "/my/awesome/file.iso",
             },
         ],
     ),
@@ -172,7 +177,19 @@ def update_publish_items(
         items = [items]
 
     for item in items:
-        validate_object_key(item.object_key)
+        if not item.object_key and not item.link_to:
+            raise HTTPException(
+                status_code=400,
+                detail="No object key or link target for '%s'" % item.web_uri,
+            )
+        if item.object_key:
+            validate_object_key(item.object_key)
+        elif item.link_to != os.path.abspath(item.link_to):
+            raise HTTPException(
+                status_code=400,
+                detail="Link target is not an absolute path: '%s'"
+                % item.link_to,
+            )
 
         db.add(models.Item(**item.dict(), publish_id=db_publish.id))
 
--- a/tests/routers/test_publish.py
+++ b/tests/routers/test_publish.py
@@ -89,6 +89,10 @@ def test_update_publish_items_typical(db, auth_header):
                     "web_uri": "/uri2",
                     "object_key": "2" * 64,
                 },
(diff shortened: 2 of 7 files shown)


exodus_gw/routers/publish.py
```diff
@@ -172,7 +179,13 @@ def update_publish_items(
         items = [items]
 
     for item in items:
-        validate_object_key(item.object_key)
+        if item.object_key:
+            validate_object_key(item.object_key)
+        elif not item.link_to:
```
rohanpm, line 184: May I suggest also validating (somewhere) that link_to is an absolute path?  I think it can be done by checking e.g. `os.path.abspath(item.link_to) == item.link_to`.

It was mentioned elsewhere that relative symlinks on the client-side at some point have to be converted to absolute, and both exodus-rsync and exodus-gw in theory could do that.  If we don't enforce that the incoming values are absolute, that means exodus-gw will need to support resolving them.  At the moment I'm under the impression that it will work better if this happens in exodus-rsync instead.

rohanpm (changes requested):

tests/routers/test_publish.py
```diff
@@ -84,10 +84,17 @@ def test_update_publish_items_typical(db, auth_header):
                 {
                     "web_uri": "/uri1",
                     "object_key": "1" * 64,
+                    "link_to": "",
```
rohanpm, line 87: As in the example, can we please not change this to pass empty strings?  The client shouldn't be forced to provide any link_to in the non-link case, so at least some tests should remain which don't provide any link_to.  There should be a mixture of items with object_key/link_to present and absent.

tests/routers/test_publish.py
```diff
@@ -89,6 +89,10 @@ def test_update_publish_items_typical(db, auth_header):
                     "web_uri": "/uri2",
                     "object_key": "2" * 64,
                 },
+                {
+                    "web_uri": "/uri3",
+                    "link_to": "/uri1",
+                },
```
nathanegillett, line 95: I'm at a total loss.
This should be fine but is somehow corrupted when sent/received. Instead of the three items listed, update_publish_items gets something like the following;
`[ItemBase(web_uri='link_to', object_key='', link_to='')]`

More oddities of note:
- Commenting out the link_to line (94) restores the web_uri to "/uri3" and causes the inclusion of the other two items
- Removing the item entirely from the request just shifts the issue: `web_uri='object_key' object_key='' link_to=''`

This is very strange and I've struggled to find a cause all day. I'm hoping it's something simple I missed, perhaps in the schemas.
@rohanpm, @crungehottman, think you could take a look?

rohanpm: @nathanegillett it's because of some very tolerant conversion from pydantic library combined with the type we set on the update_publish_items inputs.

A patch like this seems to fix it:

```
@@ -124,7 +124,7 @@ def publish(
     dependencies=[auth.needs_role("publisher")],
 )
 def update_publish_items(
-    items: Union[schemas.ItemBase, List[schemas.ItemBase]] = Body(
+    items: Union[List[schemas.ItemBase], schemas.ItemBase] = Body(
         ...,
         example=[
             {
```

When you have a Union, pydantic docs say that it'll try the types in the order listed, so it's best to put the most specific type first.  Well, it turns out that lists in some cases can be used to initialize an ItemBase instance, as in example:

```python
>>> from exodus_gw.schemas import ItemBase

# Parsing a valid dict
>>> ItemBase.parse_obj({'web_uri': 'aaa', 'object_key': '123'})
ItemBase(web_uri='aaa', object_key='123', link_to='')

# Also parseable if it's in a list, but wrongly
>>> ItemBase.parse_obj([{'web_uri': 'aaa', 'object_key': '123'}])
ItemBase(web_uri='object_key', object_key='', link_to='')
```

I haven't looked into the precise code path in pydantic which causes this but I believe it has something to do with both lists and dicts being iterable... in the "bad" case, it seems like it's iterating over dict keys.

Just doing that patch is probably enough, though it makes me wonder if we should be more strict in general and avoid Union on our inputs.

nathanegillett: I knew it'd be something dumb like this, lol. Thanks, @rohanpm.

I'd be open to removing Union. It's only used for update_publish_items to permit single publish items and I don't believe there's a real use case for that. It looks like exodus-rsync always sends a list.
Though, if that can wait for another time/PR, I'd like to get this in to unblock the next part of the symlink fix.

rohanpm (approved):

PR #378, opened 2021-11-05
Resolve item object_keys from link_to [RHELDST-8587]

nathanegillett:
Now that items with either object_key or link_to are accepted,
we can use link_to to find the correct object_key from among the other
publish items.

This commit adds the resolve_links function to the publish model.
If the function doesn't find an item with a web_uri that maches the
link_to of the target item, an error of code 400 is raised.

### DIFF
--- a/exodus_gw/models/publish.py
+++ b/exodus_gw/models/publish.py
@@ -1,6 +1,7 @@
 import uuid
 from datetime import datetime, timezone
 
+from fastapi import HTTPException
 from sqlalchemy import Column, DateTime, ForeignKey, String, event
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.orm import relationship
@@ -23,6 +24,28 @@ class Publish(Base):
     items = relationship(
         "Item", back_populates="publish", cascade="all, delete-orphan"
     )
+
+    def resolve_links(self):
+        ok_items = []
+        ln_items = []
+        for item in self.items:
+            if item.object_key:
+                ok_items.append(item)
+            else:
+                ln_items.append(item)
+
+        for ln_item in ln_items:
+            matched = [i for i in ok_items if i.web_uri == ln_item.link_to]
+            if not matched:
+                raise HTTPException(
+                    status_code=400,
+                    detail=(
+                        "Unable to resolve item object_key:"
+                        "\n\tURI: '%s'\n\tLink: '%s'"
+                    )
+                    % (ln_item.web_uri, ln_item.link_to),
+                )
+            ln_item.object_key = matched[0].object_key
 
 
 @event.listens_for(Publish, "before_update")
--- a/exodus_gw/models/publish.py
+++ b/exodus_gw/models/publish.py
@@ -26,17 +26,17 @@ class Publish(Base):
     )
 
     def resolve_links(self):
-        ok_items = []
+        ok_items = {}
         ln_items = []
         for item in self.items:
             if item.object_key:
-                ok_items.append(item)
+                ok_items[item.web_uri] = [item.object_key]
             else:
                 ln_items.append(item)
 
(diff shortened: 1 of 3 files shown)


exodus_gw/models/publish.py
```diff
@@ -24,6 +25,28 @@ class Publish(Base):
...
+        ok_items = []
+        ln_items = []
+        for item in self.items:
+            if item.object_key:
+                ok_items.append(item)
+            else:
+                ln_items.append(item)
+
+        for ln_item in ln_items:
+            matched = [i for i in ok_items if i.web_uri == ln_item.link_to]
```
rohanpm, line 38: I think this is probably a case where performance matters, and the lookup here should be indexed by web_uri (e.g. store ok_items as a dict, maybe even just `{uri => object key}` mappings rather than full objects).

The current approach is `O(N^2)` and I believe the numbers are going to be big enough for this to cause problems.  Our goal is to support 500,000 items in a single publish.  If we consider a 500,000 item publish and imagine it's a 50-50 split between links and non-links (which should be roughly accurate), we're looking at 250000*250000 =~ 60 billion iterations here which I think will be unusable.

Testing locally gives me the impression that, when we get into the hundreds of thousands of items, this will make the difference between the code working OK or not at all:

```python
>>> limit = 250000
>>> ok_items = [{'uri': '/' + str(i), 'key': str(i)} for i in range(0, limit)]
>>> ln_items = [{'uri': '/link' + str(i), 'link_to': '/' + str(i)} for i in range(0, limit)]

>>> def lookup_scan():
...   for ln_item in ln_items:
...     matched = [i for i in ok_items if i['uri'] == ln_item['link_to']]
... 

>>> def lookup_dict():
...   by_uri = {i['uri']: i['key'] for i in ok_items}
...   for ln_item in ln_items:
...     matched = by_uri.get(ln_item['link_to'])
... 

>>> from time import monotonic

# lookup using a dict can succeed in <1 second
>>> a=monotonic(); lookup_dict(); print(monotonic() - a)
0.08488436499965246

# don't know how long this takes, got tired of waiting after a couple of minutes
# and the openshift router would most likely time us out before this ever completes...
>>> a=monotonic(); lookup_scan(); print(monotonic() - a)
^CTraceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in lookup_scan
  File "<stdin>", line 3, in <listcomp>
KeyboardInterrupt
```

rohanpm (changes requested):

exodus_gw/models/publish.py
```diff
@@ -24,6 +25,27 @@ class Publish(Base):
         "Item", back_populates="publish", cascade="all, delete-orphan"
     )
 
+    def resolve_links(self):
+        ok_items = {}
+        ln_items = []
```
nathanegillett, line 30: I wonder if it would be worth it to fill ok_items and ln_items as they are added to the publish and store the publish items as a dict too. This would mean altering the db schema (and probably a few other things) but would make this resolution process more efficient. Something like...
```
# publish object would have the following;
# items = {
#     "/some/path": {"object_key": "a1b2c3", "link_to":  ""},
#     "/another/path": {"object_key": "", "link_to": "/some/path"},
# }
# ln_items = {"/another/path": "some/path", }
# ok_items = {"/some/path": "a1b2c3", }

for web_uri, link_to in self.ln_items.items():
    item = self.items.get(web_uri)
    item.object_key = self.ok_items.get(link_to)
```
wdyt, @rohanpm?

rohanpm: It'll add some complexity, so I think we should only pursue changes like that if the current approach isn't working.  From my local testing I got the impression that python code building up a dict of 250,000 items is plenty fast already.

rohanpm (approved):

PR #380, opened 2021-11-11
Resolve origin aliases before publishing [RHELDST-8582]

nathanegillett:
Because exodus-lambda maps uri aliases before looking them up on the
Exodus CDN, we must do the same when publishing to it. This commit
implements code which do just that while constructing DynamoDB write
requests.

Previously, exodus-gw would publish an item with a web_uri like
"/content/origin/rpms/some/file.iso". If exodus-lambda then receives a
request for "/content/origin/rpms/some/file.iso", it will resolve the
origin alias and futilely look for an item with a web_uri of
"/origin/rpms/some/file.iso".

### DIFF
--- a/exodus_gw/aws/dynamodb.py
+++ b/exodus_gw/aws/dynamodb.py
@@ -1,12 +1,14 @@
+import json
 import logging
 from itertools import islice
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 import backoff
 
 from .. import models
 from ..aws.client import DynamoDBClientWrapper as ddb_client
-from ..settings import Settings, get_environment
+from ..aws.util import uri_alias
+from ..settings import Environment, Settings, get_environment
 
 LOG = logging.getLogger("exodus-gw")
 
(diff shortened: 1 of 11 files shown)


exodus_gw/aws/util.py
```diff
@@ -123,3 +123,34 @@ def get_reader(cls, request):
         # a helper to make tests easier to write.
         # tests can patch over this to effectively disable streaming.
         return cls(request)
+
+
+def uri_alias(uri, aliases):
```
nathanegillett, line 128: This was copy/pasted from exodus-lambda :)

tests/async_utils.py
```diff
@@ -9,7 +9,7 @@ def assert_not_main_thread(message):
     It'll be included in the error message.
     """
 
-    thread = threading.currentThread()
+    thread = threading.current_thread()
```
nathanegillett, line 12: This sig change was raised as a deprecation warning. Unrelated to the RP but thought I'd sneak it in anyway.

exodus_gw/aws/dynamodb.py
```diff
@@ -21,86 +22,109 @@
...
+    env_obj = get_environment(env)
     settings = Settings()
 
     it = iter(items)
     batches = list(iter(lambda: tuple(islice(it, settings.batch_size)), ()))
     unprocessed_items = []
 
     for batch in batches:
-        request = create_request(env, list(batch), from_date, delete)
+        request = create_request(env_obj, list(batch), from_date, delete)
```
rohanpm, line 118: Seems like this is going to query definitions from the DB once per batch.

Should we pull that up instead so it only happens once per publish?  Reasons being:

- reduce the number of queries for performance/cost
- what if the definitions are changed in the middle of a publish?  I think it's preferable to execute the entire publish using one set of definitions, rather than potentially applying different definitions onto different items in the same publish.

exodus_gw/aws/dynamodb.py
```diff
@@ -21,86 +22,109 @@
...
+    env_obj: Environment,
+    items: List[models.Item],
+    from_date: str,
+    delete: bool = False,
 ):
     """Create the dictionary structure expected by batch_write_item."""
 
-    table_name = get_environment(env).table
+    table_name = env_obj.table
+    definitions = query_definitions(env_obj, from_date)
```
rohanpm, line 71: This doesn't seem right; what type do you expect "definitions" to be?

Later code implies it should be a dict.  But query_definitions returns whatever ddb.query returns, and I would expect that to be a list.  I know the query is run with Limit=1, but I would not expect that to change it into returning a dict, I would expect it to return a list with either 0 or 1 elements, so to get a dict you'd have to extract the first element.

Can you please check it?  And if I'm right that this is currently wrong, it would be nice to figure out why tests pass and address that as well.  For example, perhaps MagicMock is being overused, so that definitions.get will work in the tests (since MagicMock would create a 'get' method on demand) even though it would always crash in reality.

We do use mypy, so I wonder if declaring the return type of query_definitions would have helped here.

exodus_gw/aws/util.py
```diff
@@ -123,3 +123,34 @@ def get_reader(cls, request):
...
+        if not processed:
+            # We didn't resolve any alias, then we're done processing.
+            break
+
+        # We resolved at least one alias, so we need another round
+        # in case others apply now. But take out anything we've already
+        # processed, so it is not possible to recurse.
+        remaining = [r for r in remaining if r not in processed]
+
+    return uri
```
rohanpm, line 156: May I suggest adding a DEBUG log if the output is not equal to the input?  It seems difficult otherwise to judge whether aliases are having the intended effect.

rohanpm (changes requested):

exodus_gw/aws/dynamodb.py
nathanegillett: Ah, yep, good catch.

exodus_gw/aws/dynamodb.py
nathanegillett: Right, yeah, it works out to `mock_s3_client.query` returning a mock which just passes.
I'll make it return a realistic structure (with no items) by default that can then be replaced in individual tests as needed.

nathanegillett: Correcting this revealed quite a bit wrong with the handling of the query result, so thanks for raising this.

crungehottman (approved):

tests/conftest.py
```diff
@@ -29,6 +49,7 @@ def mock_aws_client():
 def mock_boto3_client():
     with mock.patch("boto3.session.Session") as mock_session:
         client = mock.MagicMock()
+        client.query.return_value = BASE_QUERY_RESPONSE
```
rohanpm, line 52: Good idea to put this in the autouse fixture so you get a realistic response by default.

rohanpm (approved):

PR #387, opened 2021-11-30
tests: add a test covering common behaviors for every endpoint

rohanpm:
There are a few invariants which all endpoints should try to satisfy in
order for the API to work correctly and consistently. Add a test which
will check these on all endpoints, to protect against mistakes when
introducing a new endpoint.

One existing test needed some minor refactoring, as it would otherwise
leave a test route installed on app which would fail test_common.

### DIFF
--- a/tests/routers/test_common.py
+++ b/tests/routers/test_common.py
@@ -1,0 +1,59 @@
+"""Tests some invariants which are common to most/all endpoints."""
+
+import pytest
+from fastapi.routing import APIRoute
+
+from exodus_gw.main import app
+
+
+def all_api_routes():
+    return [r for r in app.routes if isinstance(r, APIRoute)]
+
+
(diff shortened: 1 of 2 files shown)


tests/routers/test_common.py
```diff
@@ -0,0 +1,59 @@
...
+        # healthchecks should be usable without auth so anyone can verify the
+        # system is up
+        "/healthcheck",
+        "/healthcheck-worker",
+        # this should not need auth as the endpoint is designed to tell you
+        # whether or not you're authorized
+        "/whoami",
+        # FIXME? We do not require auth for this, but maybe we should.
+        # Unauthenticated users have no way to find these tasks or to know what
+        # they represent. It'd be safest to lock this down too.
```
nathanegillett, line 52: +1

nathanegillett (approved):

crungehottman (approved):

PR #391, opened 2021-12-10
Fix: abspath check on empty/None link_to values [RHELDST-9189]

negillett:
It was detected that non-existent link_to values were wrongly being
validated. This commit corrects that by first checking the value's
existence and refactoring Item schema validation to favor simplicity.

### DIFF
--- a/exodus_gw/schemas.py
+++ b/exodus_gw/schemas.py
@@ -1,13 +1,12 @@
+import re
 from datetime import datetime
 from enum import Enum
-from os.path import abspath, join, normpath
+from os.path import join, normpath
 from typing import Dict, List, Optional
 from uuid import UUID
 
 from fastapi import Path
-from pydantic import BaseModel, Field, root_validator, validator
-
-from .aws.util import validate_object_key
+from pydantic import BaseModel, Field, root_validator
 
 PathPublishId = Path(
     ...,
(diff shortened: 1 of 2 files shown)


nathanegillett:
Will push again for 100% coverage despite passing checks.

exodus_gw/schemas.py
```diff
@@ -41,36 +40,31 @@ class ItemBase(BaseModel):
...
-        return v
+    def validate_item(cls, values):
+        web_uri = values.get("web_uri")
+        object_key = values.get("object_key")
+        link_to = values.get("link_to")
+
+        if not web_uri:
+            raise ValueError("No URI: %s" % values)
+        values["web_uri"] = normalize_path(web_uri)
+
```
rohanpm, line 53: This seems like an ideal place to also complain if the user has passed link_to *and* object_key.

exodus_gw/schemas.py
```diff
@@ -41,36 +40,31 @@ class ItemBase(BaseModel):
...
+        if link_to:
+            if link_to != abspath(link_to):
```
rohanpm, line 55: With the introduction of normalize_path this doesn't seem right to me, because we seem to be applying different behavior arbitrarily between the web_uri and link_to fields.

I mean if you pass a web_uri of "content/foo", this is accepted and saved as "/content/foo".  But if you pass a link_to of "content/foo", you get a fatal error.

The difference in handling seems confusing.  Can we let the path normalization *replace* the absolute path check instead so that web_uri and link_to are handled similarly?

PR #400, opened 2022-01-26
Add support for setting item content type [RHELDST-8907]

negillett:
When adding items to a publish, content_type may now be specified.
Content type value must adhere to MIME type structure:
TYPE/SUBTYPE[+SUFFIX][;PARAMETER=VALUE]


exodus_gw/schemas.py
```diff
@@ -67,6 +72,15 @@ def validate_item(cls, values):
         else:
             raise ValueError("No object key or link target: %s" % values)
 
+        if content_type:
```
rohanpm, line 75: Could you please clarify the relationship between content_type and link_to?

What's supposed to happen in the following case:

- an item has web_uri = /link-src,  link_to = /link-dest, content_type = A
- the item with web_uri /link-dest has content_type = B

If the user requests /link-src, do they get a Content-Type of A or B?

I suspect the correct answer is that it should be B.  In that case we can say that including both content_type and link_to together doesn't make sense, as the content type always comes from the link destination.

If that sounds right then can we please make the validation here reject any mixing of content_type and link_to ?

tests/routers/test_publish.py
```diff
@@ -354,6 +366,45 @@ def test_update_publish_items_invalid_object_key(db, auth_header):
     }
 
 
+def test_update_publish_items_invalid_content_type(db, auth_header):
```
rohanpm, line 369: I can't find any test which uses a *valid* non-empty content-type.  Can we please have one?

rohanpm (changes requested):

exodus_gw/migrations/versions/48cfe99f5c21_.py
```diff
@@ -0,0 +1,47 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "48cfe99f5c21"
+down_revision = "be804e93d51d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # clean all publishes first to avoid crashing on null state
```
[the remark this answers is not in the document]
negillett: Yeah, I initially left it out and it passed locally. However, it failed the ci workflow. There should be records of the failed runs.

dichn (changes requested):
Could you upgrade the example in func `update_publish_items` in `exodus_gw/routers/publish.py` for setting content type support?
```
        example=[
            {
                "web_uri": "/my/awesome/file.iso",
                "object_key": "aec070645fe53ee3b3763059376134f058cc337247c978add178b6ccdfb0019f",
            },
            {
                "web_uri": "/my/slightly-less-awesome/other-file.iso",
                "object_key": "c06545d4e1a1c8e221d47e7d568c035fb32c6b6124881fd0bc17983bd9088ae0",
            },
            {
                "web_uri": "/another/route/to/my/awsome/file.iso",
                "link_to": "/my/awesome/file.iso",
            },

```

exodus_gw/migrations/versions/48cfe99f5c21_.py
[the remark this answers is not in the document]
negillett: Looks like the downgrade failed. I'll try reverting the upgrade portion.

exodus_gw/schemas.py
negillett: Hmm. Yeah, that makes sense to me.

dichn (approved):

crungehottman (approved):

PR #403, opened 2022-02-11
Support publish of 'absent' items [RHELDST-9953]

dichn:
In order to support the explicit deletion/hiding of content on exodus
CDN, it should make exodus-gw allow setting the object_key to the string
"absent" on uploaded publish items.

### DIFF
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -136,8 +136,12 @@ def update_publish_items(
                 "content_type": "application/octet-stream",
             },
             {
-                "web_uri": "/another/route/to/my/awsome/file.iso",
+                "web_uri": "/another/route/to/my/awesome/file.iso",
                 "link_to": "/my/awesome/file.iso",
+            },
+            {
+                "web_uri": "/my/awesome/deletion.iso",
+                "object_key": "absent",
             },
         ],
     ),
--- a/exodus_gw/schemas.py
+++ b/exodus_gw/schemas.py
@@ -35,7 +35,10 @@ class ItemBase(BaseModel):
         "",
         description=(
             "Key of blob to be exposed; should be the SHA256 checksum of a previously uploaded "
-            "piece of content, in lowercase hex-digest form."
+            "piece of content, in lowercase hex-digest form. \n\n"
+            "Alternatively, the string 'absent' to indicate that no content shall be exposed at the given URI. "
+            "Publishing an item with key 'absent' can be used to effectively delete formerly published "
+            "content from the point of view of a CDN consumer."
         ),
     )
     content_type: str = Field(
@@ -67,7 +70,13 @@ class ItemBase(BaseModel):
             values["link_to"] = normalize_path(link_to)
         elif object_key:
             pattern = re.compile(r"[0-9a-f]{64}")
-            if not re.match(pattern, object_key):
+            if object_key == "absent":
+                if content_type:
+                    raise ValueError(
+                        "Cannot set content type when object_key is 'absent': %s"
+                        % values
+                    )
+            elif not re.match(pattern, object_key):
                 raise ValueError(
                     "Invalid object key; must be sha256sum: %s" % values
                 )
--- a/tests/routers/test_publish.py
+++ b/tests/routers/test_publish.py
@@ -101,6 +101,10 @@ def test_update_publish_items_typical(db, auth_header):
                     "web_uri": "/uri3",
                     "link_to": "/uri1",
                 },
+                {
+                    "web_uri": "/uri4",
+                    "object_key": "absent",
+                },
             ],
             headers=auth_header(roles=["test-publisher"]),
         )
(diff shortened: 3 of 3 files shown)


exodus_gw/schemas.py
```diff
@@ -35,7 +35,8 @@ class ItemBase(BaseModel):
         "",
         description=(
             "Key of blob to be exposed; should be the SHA256 checksum of a previously uploaded "
-            "piece of content, in lowercase hex-digest form."
+            "piece of content (in lowercase hex-digest form) or the string 'absent' "
+            "(it indicates that content unable to be fetched from the CDN.)"
```
negillett, line 39: I think a little more detail would be good here. Something like...
```suggestion
            "piece of content (in lowercase hex-digest form) or the string 'absent'."
            "An 'absent' object_key allows publishing of items not uploaded."
```

crungehottman: Maybe I've misread, but from the Jira tickets in the content deletion epic, it sounds like 'absent' doesn't suggest that the content doesn't exist in the s3 bucket, just that it shouldn't be accessed from some uri. (So, I don't think "allows publishing of items not uploaded" is correct.)

I'd suggest to either remove the explanation for what 'absent' means in the context of exodus-lambda,

```suggestion
            "piece of content (in lowercase hex-digest form) or the string 'absent'."
```
or to maybe fix up the sentence like so:
```suggestion
            "piece of content (in lowercase hex-digest form) or the string 'absent'"
            "(indicates that content shouldn't be fetched from the web_uri)."
```

exodus_gw/routers/publish.py
```diff
@@ -136,9 +136,14 @@ def update_publish_items(
                 "content_type": "application/octet-stream",
             },
             {
-                "web_uri": "/another/route/to/my/awsome/file.iso",
+                "web_uri": "/another/route/to/my/awesome/file.iso",
                 "link_to": "/my/awesome/file.iso",
             },
+            {
+                "web_uri": "/another/route/to/my/awesome/deletion.iso",
```
rohanpm, line 143: The chosen URI here doesn't make sense, it should be confusing to anyone paying attention to the values.

The previous item had a uri of "/another/route/to/..." because it was a link_to file.iso, i.e. file.iso could be accessed by /my/awesome/file.iso or by /another/route/to/... , two ways to get to the same content.

For this item it's not the case, this is not "another route to" anything, because there's no link and a key of "absent" means nothing can be accessed.  I guess this was just copy-pasted from the item above.  Can you please give it a URI which makes more sense for someone reading the docs?

exodus_gw/routers/publish.py
```diff
@@ -136,9 +136,14 @@ def update_publish_items(
...
+                "object_key": "absent",
+                "content_type": "application/octet-stream",
```
rohanpm, line 145: I don't think content_type should be provided when object_key is absent.

exodus_gw/schemas.py
```diff
@@ -67,7 +68,9 @@ def validate_item(cls, values):
             values["link_to"] = normalize_path(link_to)
         elif object_key:
             pattern = re.compile(r"[0-9a-f]{64}")
-            if not re.match(pattern, object_key):
+            if object_key == "absent":
+                pass
```
rohanpm, line 72: Could we please add a validation somewhere (could be here) that content_type is not provided if object_key is "absent" ?

exodus_gw/schemas.py
rohanpm: I'll pile on with my own suggestion:

> Key of blob to be exposed; should be the SHA256 checksum of a previously uploaded
> piece of content, in lowercase hex-digest form.
> Alternatively, the string 'absent' to indicate that no content shall be exposed at the given URI.
> Publishing an item with key 'absent' can be used to effectively delete formerly published
> content from the point of view of a CDN consumer.

The main points being:

- don't try to cram too much in a single sentence
- I do think the docs should explain *somewhere* why someone might want to do this.  Hence the last sentence.  I'm borderline whether 'schemas' is really the right place for that though.  The prime alternative would be in the doc string of the update_publish_items function, e.g. right after "Publish items primarily are ...", another sentence mentioning something like "Publish items can also hide previously published items by ..."

rohanpm (changes requested):

exodus_gw/schemas.py
dichn: Gotcha, plan to update content_type validation like this:
```
        elif object_key:
            pattern = re.compile(r"[0-9a-f]{64}")
            if object_key == "absent":
                if content_type:
                    raise ValueError(
                        "content_type should not be provided if object_key is 'absent'"
                    )
```

dichn:
(1) object_key description updated
(2) example modified `"web_uri": "/my/awesome/deletion.iso"`
(3) add validation/unittest for absent items with content_type

rohanpm (changes requested):

dichn:
(1) `<br>` change to `\n\n`
(2) error message updated

rohanpm (approved):

crungehottman (approved):

PR #408, opened 2022-02-21
dev env: tweak sscg execution conditions

rohanpm:
The dev env runs sscg in order to create certs.
It previously worked by always running sscg but tolerating an exit code
of 17, which was sscg's documented code for "output already exists".

Let's change it to instead only run sscg if one of the outputs doesn't
exist or is empty, because:

- the sscg "ERROR: File exists" message can be confusing and lead the
  developer to think that something has gone wrong.

- it is potentially useful to be able to recover from "cert exists but
  is empty", as this has been observed to happen at least once (though I
  don't know why).

Additionally, the duplication of sscg between exodus-gw-cert and
exodus-gw-sidecar has been removed so that this logic is only in one
place.

### DIFF
--- a/scripts/systemd/exodus-gw-cert.service
+++ b/scripts/systemd/exodus-gw-cert.service
@@ -6,14 +6,20 @@ Description=exodus-gw development cert
 [Service]
 Type=oneshot
 ExecStartPre=mkdir -p %S/exodus-gw-dev
+
+# Only run if one of the outputs is missing or empty.
+ExecCondition=sh -c '! ( \
+  test -s %S/exodus-gw-dev/ca.crt && \
+  test -s %S/exodus-gw-dev/service.pem && \
+  test -s %S/exodus-gw-dev/service-key.pem \
+)'
+
 ExecStart=sscg \
   --cert-file=%S/exodus-gw-dev/service.pem \
   --cert-key-file=%S/exodus-gw-dev/service-key.pem \
   --ca-file=%S/exodus-gw-dev/ca.crt \
-  --subject-alt-name localhost
-
-# This exit code means the cert already exists, which is OK.
-SuccessExitStatus=17
+  --subject-alt-name localhost \
+  --force
 
 [Install]
 WantedBy=exodus-gw.target
--- a/scripts/systemd/exodus-gw-sidecar.service
+++ b/scripts/systemd/exodus-gw-sidecar.service
@@ -18,14 +18,6 @@ Environment=EXODUS_GW_SIDECAR_IMAGE=images.paas.redhat.com/it-platform/platform-
 Environment=PODMAN_SYSTEMD_UNIT=%n
 Restart=on-failure
 
-# Make self-signed cert
-ExecStartPre=mkdir -p %S/exodus-gw-dev
-ExecStartPre=-sscg \
-  --cert-file=%S/exodus-gw-dev/service.pem \
-  --cert-key-file=%S/exodus-gw-dev/service-key.pem \
-  --ca-file=%S/exodus-gw-dev/ca.crt \
-  --subject-alt-name localhost
-
 # Make properties file
 ExecStartPre=sh -c "cat >%S/exodus-gw-dev/platform-sidecar.yaml <<END\n\
 server.ssl.client-auth: want\n\


negillett (approved):

negillett:
Pulled this PR, installed and managed to start services.
The `check_prereqs` function of systemd/install is problematic. It enforces dnf install of tox, which doesn't bring dependencies and causes gw worker startup failures (missing pluggy in my experience). The real solution for the missing command would be to install tox via pip (and ensure it's installed on root, not local or in virtualenv). So I had to remove 'tox' from that list of commands to work around it.

Also, `aws configure --profile test` doesn't seem to work. I go through the steps but `scripts/localstack-init` doesn't recognize the config. I get around this by setting up my saml config (`export AWS_PROFILE=saml && export AWS_DEFAULT_REGION=us-east-1 && aws-saml.py`) and changing exodus-gw.ini env(s) to 'saml'.

rohanpm:
@negillett about those two problems:

### dnf install tox

Though I agree in general to preferring a pip-installed tox, are you sure having it installed by dnf really breaks things for you?  It's not clear why it should, and it might indicate something else wrong in your environment.

I don't think tox should be installed using pip as root and not using a virtualenv.  I think everything you install should be either installed to the system with dnf, or installed to a non-root virtualenv.  If you've been using pip to install things as root, that may in fact be the cause of some of your issues as it'll be overwriting files installed by RPM.

I see "tox installed by dnf" and "tox installed to a virtualenv" as the only two acceptable approaches.  The problem with the latter is that, as we have not standardized the virtualenv setup, the systemd units have no obvious way to figure out where your virtualenv-installed tox is located.  There would be some ways to make that possible, but I don't want to do that unless I know there's a genuine reason dnf-installed tox doesn't work.

### localstack-init

I could look into that further if you elaborate what you mean by `scripts/localstack-init doesn't recognize the config` (e.g. what's the output of the command when it doesn't work?)

negillett:
**tox:**
Using a fresh fedora35 install (to remove the possibility of corrupt packages), `dnf install tox` seems okay, but I found that gw worker needs `pg_config` command which is provided by `libpq-devel`. So we should probably add that to the install prereq check.

**localstack**
```
$ rm -rf ~/.aws
$ 
$ aws configure --profile test
AWS Access Key ID [None]: x
AWS Secret Access Key [None]: x
Default region name [None]: us-east-1
Default output format [None]: None
$
$ scripts/localstack-init
+ scripts/localstack-s3-init
+ exec aws --endpoint-url=https://localhost:3377 s3api create-bucket --bucket my-bucket

The config profile (saml) could not be found
+ scripts/localstack-dynamodb-init
+ exec aws --endpoint-url=https://localhost:3377 dynamodb create-table --table-name my-table --attribute-definitions AttributeName=web_uri,AttributeType=S AttributeName=from_date,AttributeType=S --key-schema AttributeName=web_uri,KeyType=HASH AttributeName=from_date,KeyType=RANGE --provisioned-throughput ReadCapacityUnits=10,WriteCapacityUnits=10

The config profile (saml) could not be found
+ scripts/localstack-dynamodb-config-init
+ exec aws --endpoint-url=https://localhost:3377 dynamodb create-table --table-name my-config --attribute-definitions AttributeName=config_id,AttributeType=S AttributeName=from_date,AttributeType=S --key-schema AttributeName=config_id,KeyType=HASH AttributeName=from_date,KeyType=RANGE --provisioned-throughput ReadCapacityUnits=10,WriteCapacityUnits=10

The config profile (saml) could not be found
```
This output seems different than I remember but idea seems to hold. Not sure why it's defaulting to saml (or why profile=test isn't specified anywhere).

rohanpm:
> I found that gw worker needs pg_config command

That'd be something needed in order to compile a python module with a native component.  Unfortunately I don't think we can reliably document those since anyone in our dependency tree can add/remove deps on native components at any point.  I think devs just need to have "if pip install fails with a compile error, figure out which package to install" as part of their skill set and we can't precisely spell out what has to be done.

> The config profile (saml) could not be found

This sounds like something you'd get if you had set AWS_PROFILE=saml in your environment while you didn't have a profile named "saml" in config files.  Could that be the case?

Might be work checking like `env | grep AWS` to see if you set any AWS-related environment variables you've forgotten about.

negillett:
I think you were right about the AWS_PROFILE var.
Retrying now, I'm getting the original problem I had with `configure`;
```
$ rm -rf ~/.aws
$
$ aws configure --profile test
AWS Access Key ID [None]: x
AWS Secret Access Key [None]: x
Default region name [None]: us-east-1
Default output format [None]: None
$
$ scripts/localstack-init
+ scripts/localstack-s3-init
+ exec aws --endpoint-url=https://localhost:3377 s3api create-bucket --bucket my-bucket
Unable to locate credentials. You can configure credentials by running "aws configure".
+ scripts/localstack-dynamodb-init
+ exec aws --endpoint-url=https://localhost:3377 dynamodb create-table --table-name my-table --attribute-definitions AttributeName=web_uri,AttributeType=S AttributeName=from_date,AttributeType=S --key-schema AttributeName=web_uri,KeyType=HASH AttributeName=from_date,KeyType=RANGE --provisioned-throughput ReadCapacityUnits=10,WriteCapacityUnits=10
You must specify a region. You can also configure your region by running "aws configure".
```
So, it seems setting region via configure isn't enough. As there's no `--profile` option on the create commands, maybe we'll need to set AWS_PROFILE=test as part of the setup.
After setting AWS_PROFILE I got "Unknown output type: None" from every aws command and found that "Default output format" during `aws configure` must be set to "json" (must be all lower).
So, https://release-engineering.github.io/exodus-gw/development.html#systemd-based-development-environment states that all but region, key and key ID can be ignored, but apparently that's incorrect.

rohanpm:
@negillett yeah, there was indeed a gap in the AWS config stuff.  Should be fixed by https://github.com/release-engineering/exodus-gw/pull/410 .

PR #410, opened 2022-02-25
dev env: create & enforce 'test' aws profile during localstack setup

rohanpm:
It seems that the aws commands used in the localstack-* init scripts
were using the environment's default profile (or the profile set by the
user in AWS_PROFILE).

In a clean environment, that's a problem since no profile will exist;
and even though our docs suggested creating a "test" profile, these
scripts didn't attempt to use that profile, so it wouldn't help unless
the caller explicitly set AWS_PROFILE=test.

Let's make it a bit more robust:

- localstack-init will check for a 'test' aws profile and create it if
  missing.
- localstack-init always attempts to use the 'test' aws profile.

### DIFF
--- a/scripts/localstack-init
+++ b/scripts/localstack-init
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Populate localstack with basic resources needed for exodus-gw to run.
 #
@@ -12,9 +12,29 @@
 #
 thisdir=$(dirname $0)
 
+ensure_aws_profile()(
+  # Ensure the user has an AWS profile named 'test' (name matching
+  # the profile expected in exodus-gw.ini).
+  #
+  # If the profile already exists, we do nothing since the user might
+  # have configured it a certain way and we shouldn't mess with it.
+  # However if it doesn't exist, we define a minimal one.
+  if ! aws configure list-profiles | egrep --silent "^test$"; then
+    aws configure set profile.test.region us-east-1
+    aws configure set profile.test.aws_access_key_id fake-key-id
+    aws configure set profile.test.aws_secret_access_key fake-key
+  fi
+)
+
+set -e
+ensure_aws_profile
+export AWS_PROFILE=test
+
+set -x
+
 # Note: we continue on error here because the DDB steps will complain if
 # tables already exist, which is inconvenient
-set -x
+set +e
 
 $thisdir/localstack-s3-init
 $thisdir/localstack-dynamodb-init
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -166,22 +166,6 @@ the development environment.
    * - ``curl https://localhost:3377``
      - Sanity check for localstack
 
-   * - ``aws configure --profile test``
-     - Create a ``test`` profile for AWS CLI.
-
-       In order for exodus-gw to work, an AWS profile must exist in ``~/.aws/credentials``
-       matching the ``aws_profile`` in ``exodus-gw.ini``.  This command will help to create
-       a profile if you don't already have one.
-
-       It'll prompt for a few parameters:
-
-       - For region, use ``us-east-1``.
-
-       - For access keys, use any dummy value (e.g. ``"x"``) - these won't be used,
-         but must be present.
-
-       - Other parameters can be omitted.
-
    * - ``scripts/localstack-init``
 
      - Create resources in localstack.


crungehottman (approved):

scripts/localstack-init
```diff
@@ -12,9 +12,29 @@
...
+  # the profile expected in exodus-gw.ini).
+  #
+  # If the profile already exists, we do nothing since the user might
+  # have configured it a certain way and we shouldn't mess with it.
+  # However if it doesn't exist, we define a minimal one.
+  if ! aws configure list-profiles | egrep --silent "^test$"; then
+    aws configure set profile.test.region us-east-1
+    aws configure set profile.test.aws_access_key_id fake-key-id
+    aws configure set profile.test.aws_secret_access_key fake-key
+  fi
```
negillett, line 26: We also need to set default output format to "json" here. Otherwise, only "Unknown output type: None" will be returned from `aws` commands.

negillett (changes requested):

scripts/localstack-init
negillett: I do have output set in my AWS config:
```
$ env | grep AWS
$ 
$ for file in ~/.aws/*; do echo "==== $file"; cat $file; done
==== /home/ngillett/.aws/config
[profile test]
region = us-east-1
output = json
==== /home/ngillett/.aws/credentials
[test]
aws_access_key_id = x
aws_secret_access_key = x
(py3) [ngillett@ngillett exodus-cdn-playbooks]$ aws --version
aws-cli/1.22.61 Python/3.10.2 Linux/5.16.11-200.fc35.x86_64 botocore/1.24.6
```
I can delete the dir, though, and, when reconfiguring, it prompts for an output format which defaults to None.
```
$ rm -rf ~/.aws
$ aws configure --profile test
AWS Access Key ID [None]: x
AWS Secret Access Key [None]: x
Default region name [None]: us-east-1
Default output format [None]: 
```
I see that this is a documented behavior: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-profiles
Though nothing explicitly states, using --profile seems to override the `[default]` profile.

negillett: What interesting is that, when running `aws configure --profile saml` with an existing credentials for said profile, no config file is created (and therefore output format not overridden):
```
$ aws configure --profile saml
AWS Access Key ID [****************TVDQ]: 
AWS Secret Access Key [****************ZE2N]: 
Default region name [us-west-2]: 
Default output format [None]: 
$ aws s3 ls s3://exodus-cdn-prod
2021-12-17 11:26:48     196288 06316538b90d3aab20f4787132307f6d330da9a48bab11a13e0a616fcf622ce5
  [...]
$ for file in ~/.aws/*; do echo "==== $file"; cat $file; done
==== /home/ngillett/.aws/credentials
[test]
aws_access_key_id = x
aws_secret_access_key = x

[saml]
region = us-west-2
aws_access_key_id = REDACTED
aws_secret_access_key = REDACTED
aws_security_token = REDACTED
``` 
The same happens for `[test]`:
```
$ aws configure --profile test
AWS Access Key ID [****************x]: 
AWS Secret Access Key [****************x]: 
Default region name [None]: us-east-1
Default output format [None]:
$
$ for file in ~/.aws/*; do echo "==== $file"; cat $file; done
==== /home/ngillett/.aws/config
[profile test]
region = us-east-1
==== /home/ngillett/.aws/credentials
[...]
```

negillett: And now, for some reason, the output isn't  being set to None by `aws configure`:
```
$ rm -rf ~/.aws
$ aws configure --profile test
AWS Access Key ID [None]: x
AWS Secret Access Key [None]: x
Default region name [None]: us-east-1
Default output format [None]: 
$ for file in ~/.aws/*; do echo "==== $file"; cat $file; done
==== /home/ngillett/.aws/config
[profile test]
region = us-east-1
==== /home/ngillett/.aws/credentials
[test]
aws_access_key_id = x
aws_secret_access_key = x
```
Same versions as before:
```
$ aws --version
aws-cli/1.22.61 Python/3.10.2 Linux/5.16.11-200.fc35.x86_64 botocore/1.24.6
```
So, I guess this _was_ something local, magically fixed by rerunning the command under different conditions... 
We can resolve this, then, and revisit if/when this crops up again.

negillett (approved):

scripts/localstack-init
rohanpm: I see, I note the instruction to run `aws configure` in the guide here was removed though since the idea is that the updated logic here replaces it.

PR #422, opened 2022-04-27
Reuse S3 connections [RHELDST-4926]

negillett:
This commit introduces caching for AWS S3 clients to reduce overhead
of the upload API.

### DIFF
--- a/exodus_gw/deps.py
+++ b/exodus_gw/deps.py
@@ -1,9 +1,12 @@
 """Functions intended for use with fastapi.Depends."""
 
+from asyncio import LifoQueue
+
 from fastapi import Depends, Path, Request
 
 from .auth import call_context as get_call_context
-from .settings import Settings, get_environment
+from .aws.client import S3ClientWrapper
+from .settings import Environment, Settings, get_environment
 
 # Because we cannot rename arguments to silence this pylint warning
 # without breaking the dependency injection system...
@@ -30,6 +33,49 @@ async def get_environment_from_path(
     return get_environment(env, settings)
 
 
+async def queue_for_profile(profile: str, maxsize: int):
+    # Create a queue of duplicate s3 clients for the given AWS profile.
+    # Client duplication is a procautionary measure against potential
+    # overwriting at runtime.
+
+    queue: LifoQueue[None] = LifoQueue(maxsize=maxsize)
+
+    while not queue.full():
+        client = await S3ClientWrapper(profile=profile).__aenter__()
+        queue.put_nowait(client)
+
+    return queue
+
+
+async def get_s3_client(
+    request: Request,
+    env: Environment = Depends(get_environment_from_path),
+    settings: Settings = Depends(get_settings),
+):
+    # Produce an active s3 client from the queue for the given environment.
+
+    profile = env.aws_profile
+    queue_size = settings.s3_pool_size
+    queues = request.app.state.s3_queues
+
+    if profile not in queues:
+        queues[profile] = await queue_for_profile(profile, queue_size)
+
+    queue = queues[profile]
+    client = await queue.get()
+
+    try:
+        yield client
+    except Exception as exc_info:
+        # When an exception is raised, assume the client broke.
+        # Close the client and replace it before raising.
+        client.__aexit__()
+        client = await S3ClientWrapper(profile=profile).__aenter__()
+        raise exc_info
+    finally:
+        await queue.put(client)
+
+
 # These are the preferred objects for use in endpoints,
 # e.g.
 #
@@ -39,3 +85,4 @@ db = Depends(get_db)
 call_context = Depends(get_call_context)
 env = Depends(get_environment_from_path)
 settings = Depends(get_settings)
+s3_client = Depends(get_s3_client)
--- a/exodus_gw/main.py
+++ b/exodus_gw/main.py
@@ -139,8 +139,24 @@ def db_init() -> None:
     db_migrate(app.state.db_engine, app.state.settings)
 
 
+def db_shutdown() -> None:
+    app.state.db_engine.dispose()
+    del app.state.db_engine
+
+
 def settings_init() -> None:
     app.state.settings = load_settings()
+
+
+def s3_queues_init() -> None:
+    app.state.s3_queues = {}
+
+
+def s3_queues_shutdown() -> None:
+    for q in app.state.s3_queues.values():
+        while not q.empty():
+            client = q.get_nowait()
+            client.__aexit__()
 
 
 @app.on_event("startup")
@@ -148,12 +164,13 @@ def on_startup() -> None:
     settings_init()
     loggers_init()
     db_init()
+    s3_queues_init()
 
 
 @app.on_event("shutdown")
-def db_shutdown() -> None:
-    app.state.db_engine.dispose()
-    del app.state.db_engine
+def on_shutdown() -> None:
+    db_shutdown()
+    s3_queues_shutdown()
 
 
 def new_db_session(engine):
--- a/exodus_gw/routers/upload.py
+++ b/exodus_gw/routers/upload.py
@@ -69,7 +69,7 @@ from botocore.exceptions import ClientError
 from fastapi import APIRouter, HTTPException, Path, Query, Request, Response
 
 from .. import auth, deps
-from ..aws.client import S3ClientWrapper as s3_client
+from ..aws.client import S3ClientWrapper
 from ..aws.util import (
     RequestReader,
     content_md5,
@@ -95,6 +95,7 @@ router = APIRouter(tags=[openapi_tag["name"]])
 async def multipart_upload(
     request: Request,
     env: Environment = deps.env,
+    s3: S3ClientWrapper = deps.s3_client,
     key: str = Path(..., description="S3 object key"),
     uploadId: Optional[str] = Query(
         None,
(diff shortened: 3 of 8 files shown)


exodus_gw/deps.py
```diff
@@ -30,6 +38,35 @@ async def get_environment_from_path(
...
 
 
+async def queue_for_profile(profile: str, maxsize: int = 3):
+    # Create a queue of duplicate s3 clients for the given AWS profile.
+
+    queue = LifoQueue(maxsize=maxsize)
+
+    while not queue.full():
+        client = await S3ClientWrapper(profile=profile).__aenter__()
+        queue.put_nowait(client)
```
negillett, line 48: @rohanpm, I took this from your prototype but I'm not sure I understand correctly why we'd need to duplicate clients like this. It's unclear to me how exactly this might be better than a single, persistent s3_client.

negillett (commented):
Almost there. Just working through test issues.
https://github.com/release-engineering/exodus-gw/actions/runs/2235576570
Having trouble figuring out why some calls are seemingly retrying. I've considered that multiple clients may be the cause but it doesn't seem possible for `get_s3_client` to return, and then for the test to use, more than one client from the queue.

exodus_gw/deps.py
rohanpm: Because a single client can only do one thing safely at a time... or at least this is what I have assumed when writing the code.  It's a defensive assumption when you don't know otherwise.

In case it's not clear why it might be unsafe.  Note that when we reach code like this:

```
await s3.put_object(...)
```

At that 'await' statement we might jump into the request handler for some other incoming HTTP request to exodus-gw, and that request handler itself might again reach some code of the form `await s3.put_object(...)`.

Now imagine the `s3` object is the same object in both cases, and imagine the implementer of the client wrote the code along the lines of:

```python
self._current_request = prepare_request_from_args(...)
self._current_response = await self._send_current_request()
```

That would clearly be broken as our second call to put_object can overwrite the current_request, current_response fields which are still needed to finish the first call.

I'm not saying the client is necessary implemented this way, but I am saying that if you want your object to work OK when shared across multiple threads or coroutines, you need to write the code in a certain way.  It's best to assume the authors of a third-party class haven't done this, unless they've explicitly documented that they have.

exodus_gw/deps.py
```diff
@@ -30,6 +38,35 @@ async def get_environment_from_path(
     return get_environment(env, settings)
 
 
+async def queue_for_profile(profile: str, maxsize: int = 3):
```
rohanpm, line 41: We'll want the maxsize to be configurable, as the correct value is not obvious.  Could be derived from settings.

exodus_gw/main.py
```diff
@@ -156,6 +157,16 @@ def db_shutdown() -> None:
     del app.state.db_engine
 
 
+@app.on_event("shutdown")
+async def s3_client_cleanup() -> None:
+    for queue in deps.S3_QUEUES.values():
```
rohanpm, line 162: Can the S3_QUEUES object please belong to app.state so it's associated with the app, rather than a global?

exodus_gw/deps.py
```diff
@@ -30,6 +38,35 @@ async def get_environment_from_path(
...
+
+    if profile not in S3_QUEUES:
+        S3_QUEUES[profile] = await queue_for_profile(profile)
+
+    queue = S3_QUEUES[profile]
+    client = await queue.get()
+
+    try:
+        return client
+    finally:
```
rohanpm, line 66: I think we should extend the behavior a bit like this:

- if we are returning normally here, the current logic in finally is ok (put the current client back on queue)
- if we are returning due to an exception, drop this current client and put a new one back on the queue instead

This is to defend against the client object entering a broken unrecoverable state.  For example, if the connection between the client and AWS becomes broken, are we certain that the client is designed to automatically recreate a connection in that case?  If not, then once a client becomes broken in this way, many later requests to exodus-gw will fail because we keep trying to use the broken object.

The safest thing to do is, if our request handler fails for any reason, assume the reason for failure might have something to do with the client.  So throw it away and create a new one just in case.

rohanpm:
> Almost there. Just working through test issues. https://github.com/release-engineering/exodus-gw/actions/runs/2235576570 Having trouble figuring out why some calls are seemingly retrying. I've considered that multiple clients may be the cause but it doesn't seem possible for `get_s3_client` to return, and then for the test to use, more than one client from the queue.

When exactly do you expect S3_QUEUES to be cleared out during the tests?  If not between each test, then it is possible for one test to get a client which has already been used by a previous test, isn't it?

I think it should be fixed in the following way:

- S3_QUEUES should belong to the app object and not be a global, and then...
- either: the tests currently using s3 client but not using an app object need to be refactored to use an app
- or: refactor get_s3_client so it can look somewhere else for the queues when needed, and pass empty queues to it in those tests

exodus_gw/deps.py
negillett: Thanks, @rohanpm! It's enough to know that it's precautionary.

exodus_gw/deps.py
negillett: That sounds good. However, this function is only serving up the clients. How should it know if the user of the client experienced any issues or the request handler eventually failed?

rohanpm: In https://github.com/rohanpm/exodus-gw/commit/fc710a0a8f4a56c167da2f3ae6e83c980b8b5dd2 this function had "yield client", it looks like this was changed to "return client" somewhere along the way.  That's actually broken since it means the client is going to be immediately put back on the queue in the "finally" block below which is going to run immediately after the return, so that's going to result in the clients being shared.

If you put it back to "yield", then the control flow ought to be something like this:

```python
# <we have not yet entered the request handler>
yield client
# <the request handler is now complete>
```

Furthermore if the request handler raises an exception, it should be possible for this function to get that exception.  So the overall control flow taking care of both failing and passing cases should be:

```python
try:
  # <we have not yet entered the request handler>
  yield client
  # <the request handler completed normally>
except Exception:
  # <the request handler raised an exception>
finally:
  # <the request handler completed, either normally or by raising an exception>
```

The "throw away the client and make a new one" case can be fit into the except block (which should then re-raise afterward).

Dependencies with yield & try are documented at https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/#a-dependency-with-yield-and-try .

negillett: Ah, okay. Well, we'll have to narrow the exceptions down--probably by a lot--since any exception raised would then start a potentially infinite loop of s3 client generation. Alternatively, we implement some sort of counter.
@rohanpm

negillett: Nevermind. I think I was seeing a different issue and assumed it was retrying.

exodus_gw/routers/upload.py
```diff
@@ -133,13 +134,13 @@ async def multipart_upload(
     - see also: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html
     """
 
-    if uploads == "":
+    if uploads == "" and uploadId is None:
```
negillett, line 137: This is worth pointing out, I think.
The test case ensuring you _can not_ provide both was coming up 200.

exodus_gw/deps.py
```diff
@@ -30,6 +33,47 @@ async def get_environment_from_path(
     return get_environment(env, settings)
 
 
+async def queue_for_profile(profile: str, maxsize: int):
+    # Create a queue of duplicate s3 clients for the given AWS profile.
+    # Client duplication is a procautionary measure against potential
+    # overwriting at runtime.
+
+    queue: LifoQueue[None] = LifoQueue(maxsize=maxsize)
```
rohanpm, line 41: Not that this is actually going to affect anything at runtime, but the type hint here doesn't look right since the queue isn't containing None values.

exodus_gw/deps.py
```diff
@@ -30,6 +33,47 @@ async def get_environment_from_path(
...
+        queues[profile] = await queue_for_profile(profile, queue_size)
+
+    queue = queues[profile]
+    client = await queue.get()
+
+    try:
+        yield client
+    except Exception as exc_info:
+        # Assume the client broke and replace it before raising.
+        client = await S3ClientWrapper(profile=profile).__aenter__()
```
rohanpm, line 71: I think we'd better call \_\_aexit\_\_ on the old client as well because we aren't going to do that later, since it'll no longer be processed by s3_queues_shutdown.

That'll have to be written in a way which doesn't fail to replace client if \_\_aexit\_\_ itself raises, so maybe something like this:

```python
old_client = client
client = await ...
await old_client.__aexit__()
```

exodus_gw/routers/upload.py
```diff
@@ -184,35 +186,40 @@ async def upload(
...
 
-    async with s3_client(profile=env.aws_profile) as s3:
-        response = await s3.put_object(
-            Bucket=env.bucket,
-            Key=key,
-            Body=reader,
-            ContentMD5=content_md5(request),
-            ContentLength=int(request.headers["Content-Length"]),
-        )
+    response = await s3.put_object(  # type: ignore
```
rohanpm, line 206: I'm curious what is the type: ignore suppressing here?  Seems like the old code didn't need that.

exodus_gw/settings.py
```diff
@@ -169,6 +169,9 @@ class Settings(BaseSettings):
     cdn_signature_timeout: int = 60 * 30
     """Time (in seconds) signed URLs remain valid."""
 
+    s3_queue_size: int = 3
+    """Number of S3 clients tocache in queues"""
```
rohanpm, line 173: Could we please tweak the name and doc string here so it doesn't emphasize the "queue" aspect?  Queues just happens to be how this is currently implemented.  From the point of view of a sysadmin, I don't think the setting name or text makes much sense.  The standard terminology for what we are doing here is a "pool" (as in https://en.wikipedia.org/wiki/Pool_(computer_science) ) and not something like "caching queues".

If you have a look at how connection pooling is documented for SQLAlchemy: https://docs.sqlalchemy.org/en/14/core/pooling.html

What we are doing here is the same thing but for S3 connections.   In the SQLAlchemy example, they used pool_size as the parameter controlling the size, so I think s3_pool_size would be a natural name for this setting.

rohanpm (changes requested):

exodus_gw/routers/upload.py
negillett: mypy complains about every s3 method call, saying S3ClientWrapper doesn't have those attributes; e.g., "S3 ClientWrapper has no attribute 'put_object'".
I could declare the methods in S3ClientWrapper but...it all seems silly to me.

rohanpm (approved):

PR #425, opened 2022-05-16
Improve Publish model's resolve_links [RHELDST-11634]

negillett:
Previously this method loaded every item on the publish, storing them in
memory for its duration. The number of these items can realistically
reach into many hundreds of thousands, approaching a million.

This commit refactors the method to use a querying approach to store
only the items that it needs to operate on -- those with non-empty
link_to fields -- and load only the URI and object key of the
corresponding items. These changes should significantly reduce the
impact this method has on system memory.

### DIFF
--- a/exodus_gw/models/publish.py
+++ b/exodus_gw/models/publish.py
@@ -2,9 +2,9 @@ import uuid
 from datetime import datetime, timezone
 
 from fastapi import HTTPException
-from sqlalchemy import Column, DateTime, ForeignKey, String, event
+from sqlalchemy import Column, DateTime, ForeignKey, String, event, inspect
 from sqlalchemy.dialects.postgresql import UUID
-from sqlalchemy.orm import relationship
+from sqlalchemy.orm import Bundle, relationship
 
 from .base import Base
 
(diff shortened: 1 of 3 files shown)


exodus_gw/models/publish.py
```diff
@@ -25,17 +25,20 @@ class Publish(Base):
...
+        # Store only necessary fields from matching items to conserve memory.
+        match = Bundle("match", Item.web_uri, Item.object_key)
+        matches = {
+            row.match["web_uri"]: row.match["object_key"]
+            for row in db.query(match).filter(Item.web_uri.in_(ln_item_paths))
+        }
 
         for ln_item in ln_items:
-            ln_item.object_key = ok_items.get(ln_item.link_to)
+            if matches:
```
rohanpm, line 40: This change seems unnecessary.  'matches' can't be None, the worst it can be in the case of no match is an empty dict, but in that case the .get code which is already here should work just fine.  Could we please have this back the way it was?

exodus_gw/models/publish.py
```diff
@@ -25,17 +25,20 @@ class Publish(Base):
...
 
-    def resolve_links(self):
-        ok_items = {}
-        ln_items = []
-        for item in self.items:
-            if item.object_key:
-                ok_items[item.web_uri] = item.object_key
-            else:
-                ln_items.append(item)
+    def resolve_links(self, db: Session):
```
rohanpm, line 28: It doesn't feel right that the DB session needs to be passed in here.  The object must already be connected to a session, seeing as the other methods for modifying or querying it don't need to have a session passed in (including the older version of resolve_links). This also raises the question, what happens if you pass a session in here which is different from the one originally used to obtain the Publish object?

I went poking around in sqlalchemy and I think this is the right way to get the session already attached to an object, which could allow you to drop 'db' argument here:

```python
from sqlalchemy import inspect

  # ...

  def resolve_links(self):
    # this is the session the object is connected to
    db = inspect(self).session
```

rohanpm (changes requested):

exodus_gw/models/publish.py
negillett: Agreed. This was an artifact of development, will remove.

exodus_gw/models/publish.py
negillett: I had the same feeling but couldn't find a practical solution. I'll give your suggestion a try. Thank you!

rohanpm (approved):

PR #429, opened 2022-05-19
Add log for the request routing to s3 object

dichn:

PR #430, opened 2022-05-19
tests: fix an inaccurate mock

rohanpm:
The mock request here was an AsyncMock.

The problem with this is that it causes *all* automatically generated
methods to behave as async. This is not realistic: on a real Request
object, the 'body' method is async, but most others are not. This could
cause a "coroutine ... is never awaited" error if the code under test
was changed to access any new attributes on request.

Fix it by making it a non-async Mock and making the body method
explicitly async.


rohanpm:
This is expected to fix the CI failure in https://github.com/release-engineering/exodus-gw/pull/429.

negillett (approved):

PR #431, opened 2022-05-19
Improve worker's publish efficiency [RHELDST-11635]

negillett:
Previously, when commiting a publish, the worker would load all publish
items at once, potentially hundreds of thousands.

This commit refactors the worker's commit functionality into a class
structure and changes the behavior as to load only a certain number of
items at a time from the database by yielding from select statements.
The number of items loaded at once is configurable via a new setting,
"item_yield_size", defaulting to 5000.


negillett:
The commit function is starting to look a bit large so if anyone sees an opportunity for reduction, let me know.

rohanpm (changes requested):
If I wanted to split `commit()` into smaller pieces I'd probably try something like this:

```python
@dramatiq.actor(time_limit=Settings().actor_time_limit)
def commit(publish_id: str, env: str, from_date: str):
    # Initialize stuff same as it works now
    settings = ...
    db = ...
    task = ...

    # Get all the 'return early' cases out of the way
    if task in wrong state:
      log and return early

    publish = ...
    if publish in wrong state:
      log and return early

    if publish is empty:
      log and return early

    # now we know we actually have to do something non-trivial.
    # put all that logic in a separate function.
    try:
        write_publish_items(publish)
        task.state = schemas.TaskStates.complete
        publish.state = schemas.PublishStates.committed
    except:
        LOG.exception("Task %s encountered an error", task.id)
        task.state = schemas.TaskStates.failed
        publish.state = schemas.PublishStates.failed
    db.commit()


# new function does the main 'write or rollback' logic
def write_publish_items(publish):
    db = inspect(publish).session
    final_items = []
    written_item_ids = []

    # get the items partitioned
    statement = ...
    partitions = db.execute(statement).partitions()

    try:
        # This function needs to:
        # - iterate through the partitions
        # - for any found final items (like repomd.xml), put them into final_items rather
        #   than writing them
        # - for any written items, put their IDs onto written_item_ids
        write_items_in_partitions(final_items, written_item_ids, partitions)

        # Same function can be used to write final_items, which acts as the very last partition.
        # The first argument being None indicates "don't try to detect final items".
        write_items_in_partitions(None, written_item_ids, [final_items])
    except Exception:
        # If anything goes wrong at all then we try to roll back.
        # We know that written_item_ids contains everything we need to roll back.
        LOG.warning("Rolling back %d item(s) due to error.", len(written_item_ids), exc_info=True)
        rollback_written_items(db, written_item_ids)

        # And re-raise whatever original error we ran into.
        raise

# implementations of write_items_in_partitions, rollback_written_items are left as
# an exercise for the reader
```

If splitting into multiple functions like that, perhaps putting them as instance methods on some Committer helper class would make sense too, it can hold references to db/task/publish/settings to avoid passing them as arguments to every function.

exodus_gw/worker/publish.py
```diff
@@ -18,70 +20,95 @@ def commit(publish_id: str, env: str, from_date: str):
...
+
+            # Flatten partition, saving any entry point items for later.
+            for row in partition:
+                item = row.Item
+                if basename(item.web_uri) in settings.entry_point_files:
+                    last_items.append(item)
+                else:
+                    items.append(item)
+
+            if write_batches(env, items, from_date):
```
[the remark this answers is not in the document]
rohanpm: Note this relates to my top-level comment about refactoring commit().

It will probably be easier to get this right if write_batches is changed to *just* raise an exception if something goes wrong.  Right now it has "return false" and "raise an exception" as two separate error cases.  I'm not sure that is needed.  Having just the "raise exception" case will probably work better.  You would naturally jump out of this loop as soon as the first error is encountered, rather than continuing the loop and doing unnecessary writes which only have to be rolled back later.

negillett:
Thanks, @rohanpm. It does make a lot of sense to convert this into a class.

exodus_gw/worker/publish.py
```diff
@@ -1,87 +1,171 @@
...
+        )
 
-            if items_written and last_items:
-                last_items_written = write_batches(env, last_items, from_date)
+        chunk_size = self.settings.item_yield_size
+        for index in range(0, len(self.written_item_ids), chunk_size):
+            item_ids = self.written_item_ids[index : index + chunk_size]
+            if item_ids:
+                statement = select(Item).where(Item.id.in_(item_ids))
+                for partition in self.db.execute(statement).partitions():
```
rohanpm, line 143: Minor: you already know that you are processing a max of chunk_size items, so in theory there can only ever be zero or one partitions here.  So it's not really needed to use partitions for this query.

exodus_gw/worker/publish.py
```diff
@@ -1,87 +1,171 @@
...
+        commit_obj.write_publish_items()
+        commit_obj.task.state = TaskStates.complete
+        commit_obj.publish.state = PublishStates.committed
+        commit_obj.db.commit()
+    except Exception as exc_info:
+        commit_obj.rollback_written_items(exc_info)
+        commit_obj.task.state = TaskStates.failed
+        commit_obj.publish.state = PublishStates.failed
+        commit_obj.db.commit()
+        raise
```
rohanpm, line 171: Hmm, I don't think we should raise here.  Note that if the actor raises, it's going to be retried by dramatiq a few times.  But that will definitely fail because the next try will see that the task state has already been moved to "failed" which will make should_write() return False.

I think the error handling logic needs to be like this:

- Either: do NOT mark task as failed, let exception be raised so we will be retried, and combine this with logic somewhere else to set task to failed if it is retried too many times.
- Or: mark task as failed and let the actor return successfully; implement desired retry logic inside of the actor.  This is the way it worked prior to this commit.

However, if you do drop the raise here, make sure to add appropriate logging.  As right now it seems that the exception won't be logged prior to raise if written_item_ids is empty.

exodus_gw/settings.py
```diff
@@ -105,8 +105,10 @@ class Settings(BaseSettings):
     the target revision when migrating the DB.
     """
 
+    item_yield_size: int = 5000
+    """Number of publish items to load from the DB at one time."""
     batch_size: int = 25
-    """Maximum number of items to write at one time"""
+    """Maximum number of items to write at one time."""
```
rohanpm, line 111: Minor: this doc string would do well to mention DynamoDB.

Right now there are two items next to each other:

- item_yield_size: Number of publish items to load from the DB at one time.
- batch_size: Maximum number of items to write at one time.

...and they are talking about two different databases, but nothing in the setting name or doc string mentions that.  A reasonable reader would wrongly conclude they are referring to the same thing.  Especially since the very next doc string explicitly calls out DynamoDB, and why would it make sense to do that for max_tries and not for batch_size.

rohanpm (changes requested):

exodus_gw/settings.py
negillett: Good point. I even grouped it with the DynamoDB settings. Perhaps even I was conflating the two DBs.

exodus_gw/worker/publish.py
```diff
@@ -1,87 +1,168 @@
...
+            items: List[Item] = []
+
+            for row in partition:
+                item = row.Item
+                if basename(item.web_uri) in self.settings.entry_point_files:
+                    final_items.append(item)
+                else:
+                    items.append(item)
+
+            write_batches(self.env, items, self.from_date)
```
rohanpm, line 116: I think the order of operations here is incorrect and can result in incomplete rollback.

You call write_batches, and then if and only if it succeeds, you add items to self.written_item_ids which controls the items being rolled back.

However: write_batches can raise an exception part way through.  In that case written_item_ids won't have the items added, and so when rollback proceeds it won't roll back those items.

I think the order needs to be swapped: put them on written_item_ids *before* doing the write.  The meaning of the variable is more like "*probably* written item ids" then: we can be certain that anything we wrote is contained in the list, but there is some chance that some ids we didn't write are also in the list.  So the fallback code needs to be capable of handling that.

Maybe it would be clearer to rename it to "rollback_item_ids" ?

rohanpm (changes requested):

exodus_gw/worker/publish.py
negillett: Yeah, I think that's a good idea.

rohanpm (approved):

PR #433, opened 2022-05-25
Filter null and empty strings in Publish's resolve_links [RHELDST-11634]

negillett:
When querying for publish items with link_to set, all others were
expected to have link_to=None. However, clients are able, and often do,
set link_to to an empty string rather than leaving it unset or None.

This commit improves the query to account for this by reading NULL/None
values as empty strings and filtering those instead.

### DIFF
--- a/exodus_gw/models/publish.py
+++ b/exodus_gw/models/publish.py
@@ -2,7 +2,15 @@ import uuid
 from datetime import datetime, timezone
 
 from fastapi import HTTPException
-from sqlalchemy import Column, DateTime, ForeignKey, String, event, inspect
+from sqlalchemy import (
+    Column,
+    DateTime,
+    ForeignKey,
+    String,
+    event,
+    func,
+    inspect,
+)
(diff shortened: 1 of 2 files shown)


rohanpm (approved):

rohanpm:
It would be ideal also to add a test for this - I guess it's as easy as changing a test or two to use items with link_to="" where they currently use link_to=None ?

negillett:
> It would be ideal also to add a test for this - I guess it's as easy as changing a test or two to use items with link_to="" where they currently use link_to=None ?

Yeah, that was my initial intention; set `link_to=""` on item(s) where it was previously unset. I don't think it makes sense anymore though. The query is now filtering on `""`, not `None`, so setting some item's `link_to` to `""` doesn't have the same meaning and `None` is equivalent to being unset.

I'm adding values of both `None` and `""` to an existing test anyway, just to guard against this potentially becoming broken in the future.

rohanpm (approved):

PR #434, opened 2022-05-26
Enable SSL for psycopg2 connections on devenv [RHELDST-11430]

dichn:


scripts/systemd/exodus-gw-db.service
```diff
@@ -28,10 +28,17 @@ ExecStart=/usr/bin/podman run --conmon-pidfile %t/%n-pid --cidfile %t/%n-cid --c
   --network host\
   --log-driver journald\
   -v %S/exodus-gw-dev/postgresql:/var/lib/postgresql:z\
+  -v %S/exodus-gw-dev/db-service.pem:/var/lib/postgresql/db-service.pem:z\
```
dichn, line 31: maybe should not use :z 
The 'z' option tells Podman that the volume content will be shared between containers.
maybe should use readonly
```
  -v %S/exodus-gw-dev/postgresql:/var/lib/postgresql:ro\
  -v %S/exodus-gw-dev/db-service.pem:/var/lib/postgresql/db-service.pem:ro\
```

dichn: Changed to use "ro" for the bind mount

negillett (dismissed):

scripts/systemd/exodus-gw-db.service
emanuelelevo: ok, nicely done

emanuelelevo:
LGTM

Just noticed that https://github.com/release-engineering/exodus-gw/blob/master/scripts/systemd/uninstall does not disable exodus-gw-cert.service and exodus-gw-db-cert.service. This leads to:
```
(exodus-gw-3.8.13) [elevorin@elevorin exodus-gw]$ scripts/systemd/install
...
Failed to enable unit: File /home/elevorin/.config/systemd/user/exodus-gw-cert.service already exists and is a symlink to /home/elevorin/repos/exodus/exodus-gw/scripts/systemd/exodus-gw-cert.service.
```
however such small issue could be addressed with another PR if preferred.

emanuelelevo (dismissed):

scripts/systemd/exodus-gw-db-cert.service
```diff
@@ -0,0 +1,30 @@
+# Service file for ensuring postgres development cert & key exists.
```
rohanpm, line 1: This should work, but did you consider using the existing cert, key and CA?  I don't see a technical reason why a different cert is required for the DB.

Note the same cert & key is already shared by exodus-gw-uvicorn and localstack, rather than requiring localstack to use a different cert.

The advantages of reusing the cert would be:

- you don't need to add this new unit (though you would still need to add an ExecStartPost on exodus-gw-cert)
- since it's generated by the existing CA, which users are instructed to import into system CA bundle, it means that any environment/tool should be able to connect to the local postgres with ssl enabled in the default way - without having to provide a path to CA cert.  (e.g. psql CLI)
- perhaps less confusing? I can see people getting confused by the fact that uvicorn and localstack are both using certs generated by ca.crt, but the DB for no clear reason uses a separate db-ca.crt.  Opens up some possibilities for error which don't need to be there.

rohanpm (approved):

scripts/systemd/exodus-gw-db-cert.service
dichn: @rohanpm The main reason for creating another cert, key and CA is for use `podman unshare` to change owner of the key to "postgres" in rootless podman namespace to avoid the error `"FATAL:  private key file "/var/lib/postgresql/db-service-key.pem" must be owned by the database user or root "`.
At that time, I did not consider reuse the existing certs.
```
diff --git a/scripts/systemd/exodus-gw-cert.service b/scripts/systemd/exodus-gw-cert.service
index a8c8ef2..292b8c8 100644
--- a/scripts/systemd/exodus-gw-cert.service
+++ b/scripts/systemd/exodus-gw-cert.service
@@ -21,5 +21,12 @@ ExecStart=sscg \
   --subject-alt-name localhost \
   --force
 
+ExecStartPost=cp \
+  %S/exodus-gw-dev/service-key.pem
+  %S/exodus-gw-dev/db-service-key.pem
+
+ExecStartPost=/usr/bin/podman unshare \
+  chown 999:999 %S/exodus-gw-dev/db-service-key.pem
+
 [Install]
 WantedBy=exodus-gw.target
```
Compare to cope the private key, and change the owner, I prefer create a new set of certs, mostly for making it straightforward.

dichn:
Re-pushed for reusing the existing certs.
> scripts/systemd/uninstall does not disable exodus-gw-cert.service and exodus-gw-db-cert.service.

@emanuelelevo I don't know exactly why exodus-gw-cert.service currently in `scripts/systemd/uninstall`, I think the intent maybe it prefer keep using the same key, cert, CA even reinstall the whole service. Adding disable behavior for exodus-gw-cert.service to `scripts/systemd/uninstall` will be confusing to user whether the certs are deleted or not.

rohanpm (approved):

emanuelelevo (approved):

PR #436, opened 2022-05-30
Fix exodus-gw worker DB errors on startup [RHELDST-11593]

dichn:
Currently, we consider the two exodus components as "web" and "workder".

* the web component is responsible for running DB migrations during the
  initialization of the web app
* the worker component does not run DB migrations, it simply assumes the
  DB is up-to-date.

When booting up worker, the web component initialization can be
unfinished. This patch adds DB check when booting up a worker.

### DIFF
--- a/exodus_gw/dramatiq/broker.py
+++ b/exodus_gw/dramatiq/broker.py
@@ -12,6 +12,7 @@ from sqlalchemy.orm import Session
 from exodus_gw.database import db_engine
 from exodus_gw.dramatiq.consumer import Consumer
 from exodus_gw.dramatiq.middleware import (
+    DatabaseReadyMiddleware,
     LocalNotifyMiddleware,
     PostgresNotifyMiddleware,
     SchedulerMiddleware,
@@ -43,6 +44,9 @@ class Broker(dramatiq.Broker):  # pylint: disable=abstract-method
         )
 
         self.add_middleware(LocalNotifyMiddleware())
+
+        # Enable Database readycheck when booting up a worker.
+        self.add_middleware(DatabaseReadyMiddleware(self.__db_engine))
 
         # This is postgres-specific, so...
         if "postgresql" in str(self.__db_engine.url):
--- a/exodus_gw/dramatiq/middleware/__init__.py
+++ b/exodus_gw/dramatiq/middleware/__init__.py
@@ -1,3 +1,4 @@
+from .db_ready import DatabaseReadyMiddleware
 from .local_notify import LocalNotifyMiddleware
 from .pg_notify import PostgresNotifyMiddleware
 from .scheduler import SchedulerMiddleware
@@ -6,4 +7,5 @@ __all__ = [
     "LocalNotifyMiddleware",
     "PostgresNotifyMiddleware",
     "SchedulerMiddleware",
+    "DatabaseReadyMiddleware",
 ]
--- a/exodus_gw/dramatiq/middleware/db_ready.py
+++ b/exodus_gw/dramatiq/middleware/db_ready.py
@@ -1,0 +1,22 @@
+import backoff
+from dramatiq import Middleware
+from sqlalchemy import inspect
+
+
+@backoff.on_predicate(
+    wait_gen=backoff.expo,
+    max_tries=24,
+    max_time=120,
+)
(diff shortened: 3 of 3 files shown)


dichn:
After this patch, the log will be like this:
```
$ journalctl -u exodus-gw-worker.service --user -f
...
May 30 17:08:28 arpeggio sh[337833]: dev-worker run-test: commands[0] | dramatiq --watch exodus_gw exodus_gw.worker -p 1
May 30 17:08:31 arpeggio sh[338175]: [2022-05-30 17:08:31,880] [PID 338175] [MainThread] [dramatiq.MainProcess] [INFO] Dramatiq '1.13.0' is booting up.
May 30 17:08:31 arpeggio sh[338175]: [2022-05-30 17:08:29,951] [PID 338178] [MainThread] [exodus-gw] [INFO] Scheduled actor cleanup uses message f630bd10-14b0-5f04-bda1-5642b9933dae
May 30 17:08:31 arpeggio sh[338175]: [2022-05-30 17:08:29,985] [PID 338178] [MainThread] [backoff] [INFO] Backing off db_table_check(...) for 0.1s (False)
May 30 17:08:31 arpeggio sh[338175]: [2022-05-30 17:08:30,096] [PID 338178] [MainThread] [backoff] [INFO] Backing off db_table_check(...) for 1.7s (False)
May 30 17:08:31 arpeggio sh[338175]: [2022-05-30 17:08:31,877] [PID 338178] [MainThread] [dramatiq.WorkerProcess(0)] [INFO] Worker process is ready for action.
May 30 17:08:31 arpeggio sh[338175]: [2022-05-30 17:08:31,881] [PID 338252] [MainThread] [dramatiq.ForkProcess(0)] [INFO] Fork process 'dramatiq.middleware.prometheus:_run_exposition_server' is ready for action.
May 30 17:08:31 arpeggio sh[338175]: [2022-05-30 17:08:31,890] [PID 338178] [Thread-3] [exodus-gw] [INFO] default.DQ-6c067240-ff98-4a28-b469-182e0ce12128: consumer is running
```

exodus_gw/dramatiq/middleware/scheduler.py
```diff
@@ -4,7 +4,9 @@
 from datetime import datetime
 from typing import Optional
 
+import backoff
 import pycron
+import sqlalchemy
```
emanuelelevo, line 9: could we have "from sqlalchemy import inspect", rather then importing the entire module?

emanuelelevo (changes requested):

dichn:
@emanuelelevo 
re-pushed for using "from sqlalchemy import inspect", instead of importing the entire module.
Re-request review.

emanuelelevo (approved):

exodus_gw/dramatiq/middleware/scheduler.py
```diff
@@ -144,6 +146,16 @@ def __ensure_enqueued(self, broker, actor, *args, **kwargs):
             broker.set_session(None)
 
     def after_process_boot(self, broker):
+        # Check if DB table exists when booting up worker
```
rohanpm, line 149: Small issue: why is the function declared nested here?  It doesn't access 'self' or 'broker'.  Unnecessary nesting of functions/methods tends to lead to unmaintainable code.

Bigger issue: Can you please not add it here (in SchedulerMiddleware)?

A well-designed class should focus on one responsibility.  Now with this change, the concept of "SchedulerMiddleware" class has changed from "Middleware allowing for scheduled actors with cron-style rules", to "Middleware allowing for scheduled actors with cron-style rules, and also waits for DB to become available at startup".  It's not good design, and I think that you only chose to add the code here because you found that an after_process_boot implementation exists here.

Could you please do it without changing behavior of SchedulerMiddleware class?  You could add another middleware, like DatabaseReadyMiddleware, which is dedicated purely to ensuring the DB is ready.

rohanpm (changes requested):

dichn:
@rohanpm 
Re-pushed for
1) split function: db_table_check() out of after_process_boot(), rather than making it a nested function.
2) Introduce new middleware: DatabaseReadyMiddleware, purely to ensuring the DB is ready.

rohanpm (approved):

PR #448, opened 2022-07-12
Upgrade python from 3.8 to 3.9 [RHELDST-12689]

dichn:
Bumps the python version used for GitHub CI to 3.9.


dichn:
Re-pushed for recreating the `requirements.txt` and `test-requirements.txt` with pip-compile under py39.

emanuelelevo (approved):

PR #450, opened 2022-07-28
Support "deadlines" for queued tasks [RHELDST-5506]

negillett:
This commit implements support for task deadlines, with a default of 2
hours. Tasks queued for longer than the deadline are abandoned/failed
to avoid potentially overloading systems in the event of a worker
outage.


exodus_gw/routers/publish.py
```diff
@@ -214,6 +215,14 @@ def commit_publish(
     objects from any of those publishes.
     """
 
+    if not deadline:
+        now = datetime.now(timezone.utc)
+        deadline_obj = now + timedelta(hours=settings.task_deadline)
+    else:
+        deadline_obj = datetime.strptime(
+            deadline, "2022-07-25 15:47:47.156005+00:00"
```
rohanpm, line 223: I don't think this works, does it?  strptime expects to have a format string passed, not an example timestamp.

Also, could we please do this in such a way that if the user passes an invalid timestamp, they get a 400 (client error) response rather than a 500 (internal server error) response?

And can we please have at least one test using a TestClient to invoke this endpoint with a deadline set?  This bug ought to have been caught by such a test.

exodus_gw/settings.py
```diff
@@ -127,6 +127,9 @@ class Settings(BaseSettings):
     passed. Defaults to two weeks.
     """
 
+    task_deadline: int = 2
+    """Maximum amount of time (in hours) a task should remain viable."""
```
rohanpm, line 131: I think this should mention that it's a default, as the caller can provide a larger value than this.

exodus_gw/routers/publish.py
```diff
@@ -193,6 +193,7 @@ def commit_publish(
     env: Environment = deps.env,
     db: Session = deps.db,
     settings: Settings = deps.settings,
+    deadline: str = None,
```
rohanpm, line 196: I'm not sure this is complete, does the new argument appear in the API docs at /redoc with a proper description?  It seems like you would have to add something to make sure it appears in the docs and specifies what the timestamp format is.

rohanpm (changes requested):

exodus_gw/routers/publish.py
negillett: Ah, yes, I forgot to create a test from that angle. Thanks!

exodus_gw/routers/publish.py
negillett: The new argument would appear in the response schema documentation but not the endpoint doc. It took a while but I found fastapi.Query which allows setting an example.

exodus_gw/routers/publish.py
```diff
@@ -193,6 +205,9 @@ def commit_publish(
     env: Environment = deps.env,
     db: Session = deps.db,
     settings: Settings = deps.settings,
+    deadline: Union[str, None] = Query(
+        default=None, example="2022-07-25T15:47:47.156005"
```
rohanpm, line 209: Did the format come from anywhere in particular?  I would rather use the same format as pulp if we can, example: `2022-08-01T22:54:21Z`

Differences:
- Z suffix indicates the timezone is UTC (the current example is ambiguous)
- no fractional part of second

The timezone one is the most important.

rohanpm (changes requested):

exodus_gw/routers/publish.py
negillett: Nope. I think it was just derived from the default datetime stamp. I did have to remove timezone due to Python translation conflicts but hard-coding a Z there is a trivial change. Thanks!

rohanpm (approved):

PR #454, opened 2022-08-15
Fix test_publish.py::test_commit fails at UTC+2 or higher [RHELDST-13…

dichn:
…133]

This failure is due to SQLAlchemy losing timezone information when
using SQLite for unittests. When comparing the deadline in function
task_ready(),the task.deadline is not the same as the previous one
when it's created.

This patch will make unittests use UTC time to avoid using timezone.

### DIFF
--- a/exodus_gw/migrations/versions/5bd0b38df850_.py
+++ b/exodus_gw/migrations/versions/5bd0b38df850_.py
@@ -1,0 +1,50 @@
+"""stop using timezone for column updated and deadline
+
+Revision ID: 5bd0b38df850
+Revises: 8b70b7e9c7fc
+Create Date: 2022-08-19 13:57:38.431351
+
+"""
+import os
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "5bd0b38df850"
+down_revision = "8b70b7e9c7fc"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("DELETE FROM tasks")
+    op.execute("DELETE FROM publishes")
+
+    # recreate='always' to avoid "Cannot add a NOT NULL column with default value NULL"
+    # on sqlite < 3.32.0
+    recreate = (
+        "always"
+        if "sqlite" in os.environ.get("EXODUS_GW_DB_URL", "")
+        else "auto"
+    )
+
+    with op.batch_alter_table("tasks", recreate=recreate) as batch_op:
+        batch_op.alter_column("updated", type_=sa.DateTime(), nullable=True)
+        batch_op.alter_column("deadline", type_=sa.DateTime(), nullable=True)
+    with op.batch_alter_table("publishes", recreate=recreate) as batch_op:
+        batch_op.alter_column("updated", type_=sa.DateTime(), nullable=True)
+
+
+def downgrade():
+    with op.batch_alter_table("tasks") as batch_op:
+        batch_op.alter_column(
+            "updated", type_=sa.DateTime(timezone=True), nullable=True
+        )
+        batch_op.alter_column(
+            "deadline", type_=sa.DateTime(timezone=True), nullable=True
+        )
+    with op.batch_alter_table("publishes") as batch_op:
+        batch_op.alter_column(
+            "updated", type_=sa.DateTime(timezone=True), nullable=True
+        )
--- a/tests/worker/test_publish.py
+++ b/tests/worker/test_publish.py
@@ -1,12 +1,12 @@
 import logging
 import uuid
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 
 import mock
 
 from exodus_gw import models, worker
 
-NOW_UTC = datetime.now(timezone.utc)
+NOW_UTC = datetime.utcnow()
 
 
 def _task(publish_id):
--- a/exodus_gw/models/publish.py
+++ b/exodus_gw/models/publish.py
@@ -1,5 +1,5 @@
 import uuid
-from datetime import datetime, timezone
+from datetime import datetime
 
 from fastapi import HTTPException
 from sqlalchemy import (
@@ -28,7 +28,7 @@ class Publish(Base):
     )
     env = Column(String, nullable=False)
     state = Column(String, nullable=False)
-    updated = Column(DateTime(timezone=True))
+    updated = Column(DateTime())
     items = relationship(
         "Item", back_populates="publish", cascade="all, delete-orphan"
     )
@@ -67,7 +67,7 @@ class Publish(Base):
 
 @event.listens_for(Publish, "before_update")
 def publish_before_update(_mapper, _connection, publish):
-    publish.updated = datetime.now(timezone.utc)
+    publish.updated = datetime.utcnow()
 
 
 class Item(Base):
--- a/exodus_gw/models/service.py
+++ b/exodus_gw/models/service.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timezone
+from datetime import datetime
 
 from sqlalchemy import Column, DateTime, String, event
 from sqlalchemy.dialects.postgresql import UUID
@@ -13,10 +13,10 @@ class Task(Base):
     id = Column(UUID(as_uuid=True), primary_key=True)
     publish_id = Column(UUID(as_uuid=True))
     state = Column(String, nullable=False)
-    updated = Column(DateTime(timezone=True))
-    deadline = Column(DateTime(timezone=True))
+    updated = Column(DateTime())
+    deadline = Column(DateTime())
 
 
 @event.listens_for(Task, "before_update")
 def task_before_update(_mapper, _connection, task):
-    task.updated = datetime.now(timezone.utc)
+    task.updated = datetime.utcnow()
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -71,7 +71,7 @@ indefinitely.
 """
 
 import logging
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 from typing import Dict, List, Union
 from uuid import UUID, uuid4
 
@@ -231,7 +231,7 @@ def commit_publish(
     objects from any of those publishes.
     """
 
-    now = datetime.now(timezone.utc)
+    now = datetime.utcnow()
 
     if isinstance(deadline, str):
         try:
--- a/exodus_gw/worker/publish.py
+++ b/exodus_gw/worker/publish.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from datetime import datetime, timezone
+from datetime import datetime
 from os.path import basename
 from typing import List
 
@@ -40,7 +40,7 @@ class Commit:
     @property
     def task_ready(self) -> bool:
         task = self.task
-        now = datetime.now(timezone.utc)
+        now = datetime.utcnow()
         if task.state in (TaskStates.complete, TaskStates.failed):
             LOG.warning(
                 "Task %s in unexpected state, '%s'", task.id, task.state
--- a/exodus_gw/worker/scheduled.py
+++ b/exodus_gw/worker/scheduled.py
@@ -1,5 +1,5 @@
 import logging
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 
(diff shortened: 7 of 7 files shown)


dichn:
I made a experiment to make unittests use dev-env's PostgreSQL by
```
diff --git a/tests/conftest.py b/tests/conftest.py
index b3ec280..827b137 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -141,8 +141,9 @@ def unmigrated_db():
     Note that this DB is likely to be empty. In the more common case that
     a test wants a DB with all tables in place, use 'db' instead.
     """
-
-    session = Session(bind=database.db_engine(settings.Settings()))
+    s = settings.Settings()
+    s.db_url = "postgresql://exodus-gw:exodus-gw@localhost:3355/exodus-gw?sslmode=verify-full&sslrootcert=/home/dichen/.config/exodus-gw-dev/ca.crt"
+    session = Session(bind=database.db_engine(s))
     try:
         yield session
     finally: 
```
And then ran the test at `2022-08-15T14:27:39+08:00` or `Aug 15 06:28:03 AM UTC 2022` (time check is little bit after the test run)
```
$ env TZ=Asia/Hong_Kong py.test -xv tests/worker/test_publish.py::test_commit 
```
It turns out SQLAlchemy **does not** lose timezone info when committing to PostgreSQL.
```
exodus-gw=# table tasks;
                  id                  |    state    |              publish_id              | updated |           deadli
ne            
--------------------------------------+-------------+--------------------------------------+---------+-----------------
--------------
...
 111a4692-c89b-4b57-840f-b3f0166148d2 | NOT_STARTED | 111e4567-e89b-12d3-a456-426614174000 |         | 2022-08-15 08:26
:46.386138+00
```

tests/worker/test_publish.py
```diff
@@ -1,12 +1,12 @@
...
 import uuid
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 
 import mock
 
 from exodus_gw import models, worker
 
-NOW_UTC = datetime.now(timezone.utc)
+NOW_UTC = datetime.now()
```
rohanpm, line 9: This fix will technically work, but I don't think it's a great fix because the way it works is to make the test less accurate.

I mean for example, this will cause all these tests to use a task.deadline having tzinfo=None.  However when the code is executed for real, task.deadline is always going to have a non-None tzinfo.  So this change makes the test further away from accurately exercising the code.  I see it as just hacking the test to make it pass, rather than having the test do the right thing.  It could cause other timezone-related bugs to be hidden.

Can we find a solution which doesn't have this problem?  I think the bug is telling us that we shouldn't have used timezone=True in the model definition for Task.  There is no reason to do that if your application code knows that the datetimes should always be UTC-only.  This is actually the answer given on the thread you linked to earlier at https://stackoverflow.com/a/6991536 , the "generally recommended solution" being that we just shouldn't be storing a timezone.

dichn:
A quick check
```
[dichen@arpeggio exodus-rsync]$ go run cmd/exodus-rsync/main.go -avviz --exodus-conf ~/.config/exodus-rsync.conf README.md exodus:/README.md | grep -e "Adding to publish object" -e "Completed successfully"
Thu Aug 18 04:43:18 UTC 2022 Adding to publish object {"batch":"1","item":"{/README.md c62af9d0fe652b83c22d171d0b52cd4f9adffc5613804d302fe7f0b7c51eb8c7 text/plain; charset=utf-8 }","url":"/test/publish/a686c1d8-e571-4a34-803f-2d2d2edd4f69"}
Thu Aug 18 04:43:23 UTC 2022 Completed successfully! {}
[dichen@arpeggio exodus-rsync]$ date -u
Thu Aug 18 04:43:31 AM UTC 2022
```
dateline looks good.
```
exodus-gw=# table tasks;
                  id                  |  state   |              publish_id              |            updated           
 |           deadline            
--------------------------------------+----------+--------------------------------------+------------------------------
-+-------------------------------
 e832a020-a2c8-4062-b899-f4de99b91b12 | COMPLETE | 9bee670f-94dd-4087-b98e-cd662177cfdc | 2022-08-18 04:24:58.179496+00
 | 2022-08-18 06:24:57.725715+00
 1468b951-2375-4bef-87e2-dbb3c30e0320 | COMPLETE | d3711e0d-df03-462f-9066-3619e9b796f3 | 2022-08-18 04:29:18.537177+00
 | 2022-08-18 06:29:18.110523+00
 9398b665-cac1-48a2-b768-af8fe6caa405 | COMPLETE | a686c1d8-e571-4a34-803f-2d2d2edd4f69 | 2022-08-18 04:43:19.402996+00
 | 2022-08-18 06:43:18.894075+00
(3 rows)

```

rohanpm (changes requested):
Except for the inline comments, what's here looks fine to me.

However, I'm a bit confused about one thing.  Doesn't this need an alembic migration as well?  Otherwise how does the current postgres DB react to removing timezone=True?

I checked my own DB just now:

```
exodus-gw=# \d publishes
                      Table "public.publishes"
 Column  |           Type           | Collation | Nullable | Default 
---------+--------------------------+-----------+----------+---------
 id      | uuid                     |           | not null | 
 env     | character varying        |           | not null | 
 state   | character varying        |           | not null | 
 updated | timestamp with time zone |           |          | 
Indexes:
    "publishes_pkey" PRIMARY KEY, btree (id)
Referenced by:
    TABLE "items" CONSTRAINT "items_publish_id_fkey" FOREIGN KEY (publish_id) REFERENCES publishes(id)
```

So postgres does specifically note the column as "timestamp with time zone", and I'd expect us to need a migration if we're going to change it.

rohanpm:
> ```
> 
> dateline looks good.
> 
> ```
> exodus-gw=# table tasks;
>                   id                  |  state   |              publish_id              |            updated           
>  |           deadline            
> --------------------------------------+----------+--------------------------------------+------------------------------
> -+-------------------------------
> ...
> 
>  9398b665-cac1-48a2-b768-af8fe6caa405 | COMPLETE | a686c1d8-e571-4a34-803f-2d2d2edd4f69 | 2022-08-18 04:43:19.402996+00
>  | 2022-08-18 06:43:18.894075+00
> (3 rows)
> ```

I think this actually *doesn't* look good, because if timezone isn't being stored in the DB any more as the model change suggests, then why do the displayed times have "+00" at the end?

dichn:
I have added alembic migration and restart exodus-gw with `scripts/systemd/restart`
```
    with op.batch_alter_table("tasks", recreate=recreate) as batch_op:
        batch_op.alter_column(
            "updated", existing_type=sa.DateTime(), nullable=True
        )
        batch_op.alter_column(
            "deadline", existing_type=sa.DateTime(), nullable=True
        )
    with op.batch_alter_table("publishes", recreate=recreate) as batch_op:
        batch_op.alter_column(
            "updated", existing_type=sa.DateTime(), nullable=True
        )
```
But updated column is still "timestamp with time zone".
```
exodus-gw=# table alembic_version;
 version_num  
--------------
 5bd0b38df850
(1 row)

exodus-gw=# \d tasks
                          Table "public.tasks"
   Column   |           Type           | Collation | Nullable | Default 
------------+--------------------------+-----------+----------+---------
 id         | uuid                     |           | not null | 
 state      | character varying        |           | not null | 
 publish_id | uuid                     |           |          | 
 updated    | timestamp with time zone |           |          | 
 deadline   | timestamp with time zone |           |          | 
Indexes:
    "tasks_pkey" PRIMARY KEY, btree (id)

```

exodus_gw/migrations/versions/5bd0b38df850_.py
```diff
@@ -0,0 +1,60 @@
...
+    # on sqlite < 3.32.0
+    recreate = (
+        "always"
+        if "sqlite" in os.environ.get("EXODUS_GW_DB_URL", "")
+        else "auto"
+    )
+
+    with op.batch_alter_table("tasks", recreate=recreate) as batch_op:
+        batch_op.alter_column(
+            "updated", existing_type=sa.DateTime(), nullable=True
```
rohanpm, line 35: This doesn't work because the correct argument name is `type_`, not `existing_type`.  See the docs at https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.alter_column .  If you change it like this it seems to work:

```suggestion
            "updated", type_=sa.DateTime(), nullable=True
```

I understand you would have copied this from an existing migration.  I believe all the current migrations using `existing_type` are also wrong.

rohanpm:
> 
> But updated column is still "timestamp with time zone".
> 

I added a comment explaining the reason it doesn't work.

dichn:
Sorry @rohanpm I should have looked into the argument explanation of function `alter_column`, rathan than just assuming it ("existing_type") works.
Here is a quick a check for the new commit.
```
exodus-gw=# \d publishes;
                        Table "public.publishes"
 Column  |            Type             | Collation | Nullable | Default 
---------+-----------------------------+-----------+----------+---------
 id      | uuid                        |           | not null | 
 env     | character varying           |           | not null | 
 state   | character varying           |           | not null | 
 updated | timestamp without time zone |           |          | 
Indexes:
    "publishes_pkey" PRIMARY KEY, btree (id)
Referenced by:
    TABLE "items" CONSTRAINT "items_publish_id_fkey" FOREIGN KEY (publish_id) REFERENCES publishes(id)

exodus-gw=# \d tasks;
                           Table "public.tasks"
   Column   |            Type             | Collation | Nullable | Default 
------------+-----------------------------+-----------+----------+---------
 id         | uuid                        |           | not null | 
 state      | character varying           |           | not null | 
 publish_id | uuid                        |           |          | 
 updated    | timestamp without time zone |           |          | 
 deadline   | timestamp without time zone |           |          | 
Indexes:
    "tasks_pkey" PRIMARY KEY, btree (id)

exodus-gw=# table tasks;
                  id                  |  state   |              publish_id              |          updated           | 
         deadline          
--------------------------------------+----------+--------------------------------------+----------------------------+-
---------------------------
 8d919b53-6715-4302-889a-f8752c0df194 | COMPLETE | 081fd31d-762f-40d7-b909-d38ec6127418 | 2022-08-22 00:09:39.176457 | 
2022-08-22 02:09:38.623666
(1 row)

[dichen@arpeggio exodus-rsync]$ date -u
Mon Aug 22 12:10:02 AM UTC 2022
```

dichn:
Pushed the new commit with a commit message change.
```
commit a8a8319b683532e028718ce73e17ce515a14ed37 (HEAD -> testing, origin/testing)
Date:   Fri Aug 19 15:57:21 2022 +0800

    Fix test_publish.py::test_commit fails at UTC+2 or higher [RHELDST-13133]
    
    This failure is due to SQLAlchemy losing timezone information when
    using SQLite for unittests. When comparing the deadline in function
    task_ready(),the task.deadline is not the same as the previous one
    when it's created.
    
    This patch will make it stop storing timezone information for "updated"
    and "deadline".

```

exodus_gw/migrations/versions/5bd0b38df850_.py
```diff
@@ -0,0 +1,54 @@
+"""empty message
```
rohanpm, line 1: Can we please have a proper message written?

exodus_gw/migrations/versions/5bd0b38df850_.py
```diff
@@ -0,0 +1,54 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "5bd0b38df850"
+down_revision = "8b70b7e9c7fc"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
rohanpm, line 21: Similar with the "empty message", we shouldn't really be leaving these placeholder comments here.  The code generated by alembic is meant to be rewritten.

exodus_gw/migrations/versions/5bd0b38df850_.py
dichn: Yes, changed to `stop using timezone for column updated and deadline`.

exodus_gw/migrations/versions/5bd0b38df850_.py
dichn: Removed the auto-generated comments.

rohanpm (approved):

negillett (approved):

emanuelelevo (approved):

crungehottman (approved):

PR #457, opened 2022-08-25
Fix the migration failure on startup [RHELDST-13372]

dichn:
When both the "items" and "publishes" tables contain some data, then
only deleting the existing records in "publishes" table without cleaning
"items" table, it will violate the item's foreign key constraint
pointing back at publishes.

This commit adds the "items" table cleaning.


dichn:
Check the migration log
```

dichn:
I was closing this by accident, hit the shortcut by mistake.

rohanpm (approved):

PR #459, opened 2022-08-29
Fix the missing Content-Type when using links during publish [RHELDST…

dichn:
…-13216]

Publishes using exodus-rsync & exodus-gw are supposed to automatically
populate a meaningful Content-Type on each published item.

Currently, if items are published via a symlink, no Content-Type is
assigned. This commit adds the copying of the link target's Contetn-Type
when resolving a link.

### DIFF
--- a/exodus_gw/models/publish.py
+++ b/exodus_gw/models/publish.py
@@ -46,15 +46,25 @@ class Publish(Base):
         ln_item_paths = [item.link_to for item in ln_items]
 
         # Store only necessary fields from matching items to conserve memory.
-        match = Bundle("match", Item.web_uri, Item.object_key)
+        match = Bundle(
+            "match", Item.web_uri, Item.object_key, Item.content_type
+        )
         matches = {
-            row.match["web_uri"]: row.match["object_key"]
+            row.match["web_uri"]: {
+                "object_key": row.match["object_key"],
+                "content_type": row.match["content_type"],
+            }
             for row in db.query(match).filter(Item.web_uri.in_(ln_item_paths))
         }
 
         for ln_item in ln_items:
-            ln_item.object_key = matches.get(ln_item.link_to)
-            if not ln_item.object_key:
+            match = matches.get(ln_item.link_to)
+
+            if (
+                not match
+                or not match.get("object_key")
+                or not match.get("content_type")
+            ):
                 raise HTTPException(
                     status_code=400,
                     detail=(
@@ -63,6 +73,9 @@ class Publish(Base):
                     )
                     % (ln_item.web_uri, ln_item.link_to),
                 )
+
+            ln_item.object_key = match.get("object_key")
+            ln_item.content_type = match.get("content_type")
 
 
 @event.listens_for(Publish, "before_update")
--- a/tests/routers/test_publish.py
+++ b/tests/routers/test_publish.py
@@ -656,12 +656,14 @@ def test_commit_publish_linked_items(mock_commit, fake_publish, db):
         object_key="1" * 64,
         publish_id=fake_publish.id,
         link_to=None,  # It should be able to handle None/NULL link_to values...
+        content_type="some type",
     )
     item2 = Item(
         web_uri="/another/path",
         object_key="2" * 64,
         publish_id=fake_publish.id,
         link_to="",  # ...and empty string link_to values...
+        content_type="another type",
     )
     item3 = Item(
         web_uri="/some/different/path",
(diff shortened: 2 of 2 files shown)


negillett:
@dichn, your commit subject line is too long, wrapping into the body.

exodus_gw/models/publish.py
```diff
@@ -46,15 +46,19 @@ def resolve_links(self):
...
+                "object_key": row.match["object_key"],
+                "content_type": row.match["content_type"],
+            }
             for row in db.query(match).filter(Item.web_uri.in_(ln_item_paths))
         }
 
         for ln_item in ln_items:
-            ln_item.object_key = matches.get(ln_item.link_to)
-            if not ln_item.object_key:
+            if not matches:
```
negillett, line 61: I think it still makes sense to look for the desired object_key here. Otherwise, this check will pass for any arbitrary object in `matches` with the potential to set ln_item values to NoneType. So, I think we'd want the code you've added below to be up here and then check if the object_key was actually set before raising, as was done before.
```suggestion
            match = matches.get(lin_item.link_to)
            ln_item.object_key = match.get("object_key")
            ln_item.content_type = match.get("content_type")
            if not ln_item.object_key or not ln_item.content_type:
```
Or at least find the correct match from `matches` and verify it has both an object_key and a content_type before passing the exception.

negillett (changes requested):

exodus_gw/models/publish.py
rohanpm, line 61: I find it unclear why this change is needed at all, what was wrong with it earlier?  It definitely seems wrong now.

rohanpm (changes requested):

exodus_gw/models/publish.py
negillett: The minimum that needs to happen here is setting `ln_item.content_type` to that of the associated matched item.

dichn: I made this change because the test case `test_commit_publish_unresolved_links` which would make `matches` an empty dict `{}`
```
(Pdb) l
 57  	            for row in db.query(match).filter(Item.web_uri.in_(ln_item_paths))
 58  	        }
 59  	
 60  	        import pdb; pdb.set_trace()
 61  	
 62  ->	        for ln_item in ln_items:
 63  	            if not matches.get(ln_item.link_to):
 64  	                raise HTTPException(
 65  	                    status_code=400,
 66  	                    detail=(
 67  	                        "Unable to resolve item object_key:"
(Pdb) matches
{}
```
So I thought once the matches is empty, I can just raise the exception.

rohanpm: That doesn't make sense because if matches is empty, then:

```
ln_item.object_key = matches.get(ln_item.link_to)
```

ln_item.object_key will be None, so then:

```
if not ln_item.object_key:
if not matches:
```

Either one of these will evaluate to True.  So why would it need to be changed?

dichn: Because of the matches updating
```
        matches = {
            row.match["web_uri"]: {
                "object_key": row.match["object_key"],
                "content_type": row.match["content_type"],
            }
            for row in db.query(match).filter(Item.web_uri.in_(ln_item_paths))
        }
```
The object_key fetching turns to `ln_item.object_key = matches.get(ln_item.link_to).get("object_key")`.
In the test case `test_commit_publish_unresolved_links`, `matches.get(ln_item.link_to)` became `None` which does not have the attribute `get`.

Then I thought I should not use previous procedure  1) object_key fetching 2) object_key checking. Because 1) would fail   when publish unresolved links.

rohanpm: If the situation is still unclear, it might make sense to try writing a test first.

By changing it to `if not matches` you've made it so that an error will only be raised here if *all* link items in the publish cannot be resolved.

What we need is to see an error if *any* of the link items in the publish cannot be resolved (even if some of them can be).

The fact you were able to change it to `if not matches` without any test failure, tells us that we are probably missing a test where some of the items are broken links and others are non-broken links, because if we had such a test it should have failed.  So you might consider adding a test like that.

rohanpm (approved):

dichn:
> your commit subject line is too long, wrapping into the body.

@negillett Yeah, the title may be a little bit longer, but I think it's fine ( I don't have a better, more brief title ). And it only affects the description in GitHub PR page I think.

negillett (approved):

PR #461, opened 2022-09-06
Add alembic_upgrade.py for DB migration [RHELDST-13374]

dichn:
Because a DB migration can take up to 15 minutes, we decide to decouple
it from the gunicorn initialization.

This commit adds a new script for calling alembic upgrade.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,8 +39,13 @@ setup(
         "Programming Language :: Python :: 3",
     ],
     install_requires=get_requirements(),
-    python_requires=">=3.8",
+    python_requires=">=3.9",
     project_urls={
         "Documentation": "https://release-engineering.github.io/exodus-gw",
+    },
+    entry_points={
+        "console_scripts": [
+            "exodus-gw-alembic-upgrade = exodus_gw.alembic_upgrade:entry_point"
+        ]
     },
 )
--- a/tests/test_alembic_upgrade.py
+++ b/tests/test_alembic_upgrade.py
@@ -1,0 +1,21 @@
+import mock
+import pytest
+from sqlalchemy.exc import OperationalError
+
+from exodus_gw.alembic_upgrade import entry_point
+from exodus_gw.models import Publish
+
+
+@mock.patch("exodus_gw.database.db_engine")
+def test_entry_point_alembic_upgrade(mock_db_engine, unmigrated_db):
+    # Sanity check: at first there's no tables
+    with pytest.raises(OperationalError):
+        unmigrated_db.query(Publish).count()
+
+    mock_db_engine.return_value = unmigrated_db.get_bind()
+
+    # Migrate should succeed
+    entry_point()
+
+    # Now we can query stuff
+    assert unmigrated_db.query(Publish).count() == 0
--- a/exodus_gw/alembic_upgrade.py
+++ b/exodus_gw/alembic_upgrade.py
@@ -1,0 +1,20 @@
+#
+# A wrapper for "alembic upgrade head" which will run upgrade operations to
+# the latest revision.
+#
+
+from .database import db_engine
+from .migrate import db_migrate
+from .settings import MigrationMode, load_settings
+
+
+def entry_point():
+    settings = load_settings()
+
+    settings.db_migration_mode = MigrationMode.upgrade
+    settings.db_migration_revision = "head"
+
+    engine = db_engine(settings)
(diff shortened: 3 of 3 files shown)


dichn:
Should I keep the logging module in script `alembic-upgrade`? I think the script will be called in a container, but the script will call the function `db_migrate` which contains many logging.

scripts/alembic-upgrade
```diff
@@ -0,0 +1,30 @@
...
+
+    engine = db_engine(settings)
+
+    db_migrate(engine, settings)
+
+    print("DB migration finished.")
+
+
+if __name__ == "__main__":
+    main(sys.argv[1:])
```
rohanpm, line 30: Have you seen this working?  I'd expect it to crash since you're passing arguments into `main()` but it doesn't accept any arguments.

rohanpm:
@dichn , I think it would be better for this not to be in `scripts` directory.  We want this to be installed properly into the container image; pip isn't installing anything from the scripts dir.

I believe this should be done as a console_scripts entry point, with the code living somewhere inside of `exodus_gw` directory.  Then it would be installed normally and runnable normally like any pip-installed command. 
 For an example in a current project using pyproject.toml you can look at https://url.corp.redhat.com/436ee8f .

scripts/alembic-upgrade
```diff
@@ -0,0 +1,30 @@
...
+
+from exodus_gw.database import db_engine
+from exodus_gw.migrate import db_migrate
+from exodus_gw.settings import MigrationMode, Settings
+
+LOG = logging.getLogger("exodus-gw")
+
+
+def main():
+    settings = Settings(
```
rohanpm, line 17: We should probably be using load_settings function, otherwise loggers don't get initialized from the config.  You can load_settings() first and then overwrite the two fields we want to force.

rohanpm (changes requested):

setup.py
```diff
@@ -39,8 +39,13 @@ def get_requirements():
...
     ],
     install_requires=get_requirements(),
-    python_requires=">=3.8",
+    python_requires=">=3.9",
     project_urls={
         "Documentation": "https://release-engineering.github.io/exodus-gw",
     },
+    entry_points={
+        "console_scripts": [
+            "exodus_gw_alembic_upgrade = exodus_gw.alembic_upgrade:entry_point"
```
rohanpm, line 48: Generally command-line tools prefer '-' over '_' to separate words (examples: ansible-playbook, ssh-keygen, gpg-agent).    This should be the same, so:

```suggestion
            "exodus-gw-alembic-upgrade = exodus_gw.alembic_upgrade:entry_point"
```

tests/test_alembic_upgrade.py
```diff
@@ -0,0 +1,9 @@
+import mock
+
+from exodus_gw.alembic_upgrade import entry_point
+
+
+@mock.patch("exodus_gw.database.db_engine")
+def test_entry_point_alembic_upograde(mock_db_engine, unmigrated_db):
+    mock_db_engine.return_value = unmigrated_db.get_bind()
+    entry_point()
```
rohanpm, line 9: Some kind of check that migrations actually happened would be good... as this test would pass even if the entry_point simply returned without doing anything.

dichn: Added a migration check similar to `test_db_migrate_typical`.

setup.py
dichn: console script updated to `exodus-gw-alembic-upgrade`

rohanpm (approved):

tests/test_alembic_upgrade.py
```diff
@@ -0,0 +1,21 @@
+import mock
+import pytest
+from sqlalchemy.exc import OperationalError
+
+from exodus_gw.alembic_upgrade import entry_point
+from exodus_gw.models import Publish
+
+
+@mock.patch("exodus_gw.database.db_engine")
+def test_entry_point_alembic_upograde(mock_db_engine, unmigrated_db):
```
negillett, line 10:
```suggestion
def test_entry_point_alembic_upgrade(mock_db_engine, unmigrated_db):
```

negillett (changes requested):

dichn:
Re-pushed for the typo fix.

rohanpm (approved):

negillett (approved):

PR #467, opened 2022-10-06
Revert "Parallelize batch writes [RHELDST-13451]"

crungehottman:
This reverts commit d2939b30c6fb14c00f2e89a46824ce987419a108.


negillett:
I'm not yet convinced this change is causing the observed memory issue.

crungehottman:
> I'm not yet convinced this change is causing the observed memory issue.

I'm not fully convinced either, but I'll leave this here for a while in case we decide we need to revert it.

rohanpm (changes requested):
I'd rather not revert it, instead find a way to only move forward.

I made a suggestion on RHELDST-13451 about reducing memory usage by setting some env vars.  If it doesn't work, I'd still rather we try some other things like reducing the thread count before reverting this.

negillett:
The memory problem on OCP doesn't seem related to threading at all anyway. This change doesn't alter the historical memory usage either.

negillett (changes requested):
I think we're good to close this.

crungehottman:
> I think we're good to close this.

Thanks for the reminder :)

PR #470, opened 2022-10-11
Improve use of batch write threads [RHELDST-13926]

negillett:
Previously, when writing items to DynamoDB, a single partition would be loaded and sumbitted to the thread pool executor, blocking until all items in the partition were completed.

This commit allows the number of partitions to be configured, meaning blocking only occurs once all item batches within n number of partitions are submitted to the executor.

Note that items are still only loaded per partition, so there should be no drastic memory usage increase. There is, however, correlation between number of partitions started and memory used due to the futures stored for each batch write but this should not normally be significant.

### DIFF
--- a/exodus_gw/worker/publish.py
+++ b/exodus_gw/worker/publish.py
@@ -1,9 +1,10 @@
 import asyncio
+import itertools
 import logging
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime
 from os.path import basename
-from typing import List
+from typing import Any, List
 
 import dramatiq
 from dramatiq.middleware import CurrentMessage
@@ -120,7 +121,7 @@ class Commit:
             return False
         return True
 
-    def batch_write_wrapper(self, executor, items, delete=False):
+    def batch_write_wrapper(self, executor, items, delete=False) -> List[Any]:
         futures = []
         batches = self.dynamodb.get_batches(items)
         for batch in batches:
@@ -130,8 +131,8 @@ class Commit:
                     self.dynamodb.write_batch, list(batch), delete=delete
                 )
             )
-        for future in as_completed(futures):
-            future.result()
+
+        return futures
 
     def write_publish_items(self) -> None:
         """Query for publish items, batching and yielding them to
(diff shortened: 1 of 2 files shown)


negillett:
@rohanpm @crungehottman @dichn @emanuelelevo

exodus_gw/worker/publish.py
```diff
@@ -150,31 +151,47 @@ def write_publish_items(self) -> None:
...
+                        ):
+                            final_items.append(item)
+                        else:
+                            items.append(item)
+
+                    # Save IDs of this chunk of items in case rollback is needed.
+                    self.rollback_item_ids.extend([item.id for item in items])
+                    futures.extend(self.batch_write_wrapper(executor, items))
+
+                for future in as_completed(futures):
```
rohanpm, line 183: I don't think there's been much of an improvement here.   The core issue I complained about in the old code is that there was a part of the loop where we unnecessarily waited for *all* outstanding batches to complete before starting work on any more.  That's still here, it's just been changed from "it blocks at every partition" to "it blocks at every N'th partition" (every 3rd by default).  But I think almost the same effect could have been achieved by just multiplying the partition size by 3.

Another way of saying it - from my original quote in the JIRA referring to the old code:

> If there's 1000 partitions it means there will be 1000 little pauses

Now with this change, if there's 1000 partitions there will be 333 little pauses (at the cost of 3x as many items loaded into memory at once)... so that could be considered an improvement.  But you could achieve the same thing already by just increasing the partition size by 3, which with the existing code would have resulted in 333 partitions and 333 pauses...

The complaint is about the following scenario:

- we've got N threads working on writing batches
- now imagine one batch write operation is slow for whatever reason (maybe it has to retry)
- then at this loop with as_completed, (N-1) threads can complete, and the loop has to wait for the remaining one thread
- it doesn't make sense, why shouldn't we continue onto the next items as quickly as possible instead of leaving most of the threads idle?

The old code had that problem, the new code still has it as well.  It is probably going to need a larger refactoring to fix that.

rohanpm (changes requested):

exodus_gw/worker/publish.py
negillett: > at the cost of 3x as many items loaded into memory at once

Well, only the items for a single partition is loaded at one time. So, yeah, it's an improvement but I'll take another step back and see what I can do.

PR #472, opened 2022-10-17
Improve use of batch write threads [RHELDST-13926]

negillett:
Previously, when writing items to DynamoDB, a single chunk of items was loaded and sumbitted to a thread pool executor, blocking until all items in the partition were completed.

This commit refactors the worker, queuing all batches from each partition and using multiple threads, rather than an executor, to write the batches as pulled from the queue. Blocking now may only occur when threads are signaled to close.


negillett:
@release-engineering/exodus 
I haven't tested this on dev just yet so I'm unsure of final settings but opening for more general review.

exodus_gw/worker/publish.py
```diff
@@ -16,9 +17,79 @@
...
+    def queue_batches(self, items: List[Item]):
+        for batch in self.dynamodb.get_batches(items):
+            self.queue.put_nowait(list(batch))
+
+    def write_batches(self):
+        """Will either submit batch write or delete requests based on
+        the 'delete' attribute.
+        """
+        while True:
+            try:
```
rohanpm, line 83: I think somewhere around here, we'll also want a check like this:

```python
    # return early if other threads already failed
    if self.write_errors:
        break
```

Otherwise if one thread hits a fatal error, we will still have to wait potentially a long time for other threads to finish their loop even though we already know we're going to fail the publish.

exodus_gw/worker/publish.py
```diff
@@ -16,9 +17,79 @@
...
+                batch = self.queue.get(True, self.settings.write_queue_timeout)
+                if batch is self.queue_sentinel:
+                    break
+                self.dynamodb.write_batch(batch, delete=self.delete)
+            except (RuntimeError, ValueError) as err:
+                self.write_errors.append(err)
```
rohanpm, line 89: The exception should probably be logged as well, because in the end we can only re-raise a single exception from write_errors.  So if there were more than one, and they're not logged, we've permanently lost the details of some errors.

exodus_gw/worker/publish.py
```diff
@@ -16,9 +17,79 @@
...
+        for thread in self.write_threads:
+            thread.join(timeout=60.0)
+
+        if self.queue.qsize() > 0:
+            # Don't warn for excess sentinels.
+            if not self.queue.get_nowait() is self.queue_sentinel:
+                LOG.warning("Commit may be incomplete, queue not empty")
+
+        if self.write_errors:
+            raise self.write_errors[-1]
```
rohanpm, line 72: I think it'd make more sense to raise the first one rather than the last.  Generally speaking, if multiple errors occurred then the first one tends to give the most valuable info.

exodus_gw/worker/publish.py
```diff
@@ -16,9 +17,79 @@
...
+            try:
+                batch = self.queue.get(True, self.settings.write_queue_timeout)
+                if batch is self.queue_sentinel:
+                    break
+                self.dynamodb.write_batch(batch, delete=self.delete)
+            except (RuntimeError, ValueError) as err:
+                self.write_errors.append(err)
+                break
+            finally:
+                self.queue.task_done()
```
rohanpm, line 92: I think this could be dropped, as we never make use of it.  This allows you to use self.queue.join() but we never do that (and it's redundant as we already join the threads).

rohanpm (changes requested):

negillett:
@release-engineering/exodus this is ready for another round of review

exodus_gw/settings.py
```diff
@@ -117,7 +117,12 @@ class Settings(BaseSettings):
     """Maximum write attempts to the DynamoDB table."""
     write_max_workers: int = 10
     """Maximum number of worker threads used in the DynamoDB batch writes."""
-
+    write_queue_size: int = 1000
+    """Maximum number of items the queue can hold at one time."""
+    write_queue_timeout: int = 60
```
rohanpm, line 122: Given the way this value is used within the code, I think the default should probably be quite a bit higher.  Say 10 minutes.

This number relates to how fast DDB writes happen and how fast postgres reads happen, which are both hard to predict.

When the writing side of the queue applies this timeout, they are waiting for the queue to become not full.  The reason the queue would transition from "full" to "not full" is that a thread managed to write a batch of items, so this is effectively a time limit on how long a single batch write can take.

60 seconds seems a bit too low for me.  I know it's going to be faster than that typically, but throw in the possibility of retries caused by temporary failures or by DynamoDB rate limiting and it seems feasible that we're genuinely going to need longer than that from time to time.

On the other side of things, for the reading side of the queue they are waiting for the queue to become not empty.  The reason the queue would transition from "empty" to "not empty" is that the main thread managed to read another partition of items from postgres and put them on the queue.  So this is *also* a time limit on how long that's allowed to take (if reading from postgres is slower than writing to DDB, which it probably isn't, but as I said - unpredictable).

In both cases 60 seconds seems a bit low to be a hard limit.  I think this timeout should be treated more as "if we didn't make progress in this time, we are *very* confident that something has gone wrong", so probably needs to be bumped up to a higher number.

rohanpm (changes requested):

rohanpm (approved):

PR #477, opened 2022-11-04
Make exodus-gw resilient against postgres connection drops [RHELDST-13928]

crungehottman:
Exodus-gw now uses the sqlalchemy pool_pre_ping option to detect and recover from a broken DB connection at the beginning of the request. To accommodate for a broken DB connection within the request, exodus-gw wraps select endpoints in a "retry" decorator.

### DIFF
--- a/exodus_gw/main.py
+++ b/exodus_gw/main.py
@@ -58,11 +58,13 @@ If you are a client looking to make use of exodus-gw, consult your organization'
 internal documentation for advice on which environment(s) you should be using.
 """
 
+import backoff
 import dramatiq
 from fastapi import FastAPI, Request
 from fastapi.exception_handlers import http_exception_handler
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse
+from sqlalchemy.exc import DBAPIError
 from sqlalchemy.orm import Session
 from starlette.concurrency import run_in_threadpool
 from starlette.exceptions import HTTPException as StarletteHTTPException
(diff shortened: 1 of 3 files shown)


exodus_gw/retry.py
```diff
@@ -0,0 +1,36 @@
...
+
+def retry():
+    """
+    The retry decoratior retries an operation up to three times.
+    """
+
+    def decorator(func):
+        @wraps(func)
+        def wrapsfunc(*args, **kwargs):
+            tries = 3
```
negillett, line 17: tries should probably be configurable

I'd suggest making it a parameter and providing a setting value in routers/publish.py

exodus_gw/database.py
```diff
@@ -17,4 +17,4 @@ def db_url(settings: Settings):
 
 
 def db_engine(settings: Settings):
-    return create_engine(db_url(settings))
+    return create_engine(db_url(settings), pool_pre_ping=True)
```
rohanpm, line 20: It might be better to put this part in a separate pull request.  This should be uncontroversial and low risk, but per my comments on the retry decorator I think the other part of this change needs a lot more work and testing.

rohanpm (changes requested):

negillett (approved):

exodus_gw/main.py
```diff
@@ -165,26 +167,36 @@ def new_db_session(engine):
...
-        broker.set_session(None)
-        await run_in_threadpool(request.state.db.close)
-        request.state.db = None
+        # Any dramatiq operations should also make use of this session.
+        broker = dramatiq.get_broker()
+        broker.set_session(request.state.db)
 
-    return response
+        try:
+            response = await call_next(request)
```
rohanpm, line 188: I was thinking of this approach before as well. Just one thing makes me nervous about it: I can't find any info on whether or not it's OK for a middleware to do call_next more than once for a single request.

It's never mentioned anywhere in the Starlette middleware docs as far as I can tell, and no examples do it.  At the same time there are some examples of middleware which *modify* the request, and it seems like those could easily be written in such a way that calling them more than once will break (it would have to be safe to do the modification more than once). This makes me worried that having one middleware call the next more than once could lead to subtle bugs.

However it does seem like the raise-db-and-resolve test you added should be covering this, so it must be generally working.  I guess we should go ahead with this approach, but keep it in mind that if we see any strange behaviors or new bugs related to request handling, this middleware calling the next more than once might be a cause.

rohanpm (approved):

PR #483, opened 2022-11-14
Support metadata on S3 uploads [RHELDST-14181]

negillett:
When migrating content in the future, we plan to store information on uploaded objects. This commit allows users to set metadata via headers for fields defined in settings and HEAD requests will now return metadata stored with the object.

### DIFF
--- a/exodus_gw/aws/util.py
+++ b/exodus_gw/aws/util.py
@@ -1,13 +1,45 @@
 import io
 import logging
 import re
-from typing import AnyStr
+from typing import AnyStr, Dict
 from xml.etree.ElementTree import Element, ElementTree, SubElement
 
 from defusedxml.ElementTree import fromstring
-from fastapi import HTTPException, Response
+from fastapi import HTTPException, Request, Response
+
+from ..settings import Settings
 
 LOG = logging.getLogger("exodus-gw")
+
+
+def extract_request_metadata(request: Request, settings: Settings):
+    # Any headers prefixed with "x-amz-meta-" will be picked out as
+    # metadata for s3 upload.
+    #
+    # https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html
+    metadata = {}
+    for k, v in request.headers.items():
+        if k.startswith("x-amz-meta-"):
+            metadata[k.replace("x-amz-meta-", "", 1)] = v
+
+    validate_metadata(metadata, settings)
+
+    return metadata
+
+
+def validate_metadata(metadata: Dict[str, str], settings: Settings):
+    valid_meta_fields = settings.upload_meta_fields
+    for k, v in metadata.items():
+        if k not in valid_meta_fields.keys():
+            raise HTTPException(400, detail="Invalid metadata field, '%s'" % k)
+
+        pattern = re.compile(valid_meta_fields[k])
+
+        if not re.match(pattern, v):
+            raise HTTPException(
+                400,
+                detail="Invalid value for metadata field '%s', '%s'" % (k, v),
+            )
 
 
 def validate_object_key(key: str):
--- a/exodus_gw/routers/upload.py
+++ b/exodus_gw/routers/upload.py
@@ -13,7 +13,7 @@ Differences from the AWS S3 API include:
 
 - Most optional arguments are not supported.
 
-- All `x-amz-*` headers are omitted from responses.
+- All `x-amz-*` headers, other than `x-amz-meta-*`, are omitted from responses.
 
 - The usual AWS authentication mechanism is unused; request signatures are ignored.
   Authentication is expected to be performed by other means.
@@ -63,7 +63,7 @@ bucket.upload_file('/tmp/hello.txt',
 """
 import logging
 import textwrap
-from typing import Optional
+from typing import Dict, Optional
 
 from botocore.exceptions import ClientError
 from fastapi import APIRouter, HTTPException, Path, Query, Request, Response
@@ -74,10 +74,11 @@ from ..aws.util import (
     RequestReader,
     content_md5,
     extract_mpu_parts,
+    extract_request_metadata,
     validate_object_key,
     xml_response,
 )
-from ..settings import Environment
+from ..settings import Environment, Settings
 
 LOG = logging.getLogger("s3")
 
@@ -119,6 +120,7 @@ async def multipart_upload(
             Must not be passed together with ``uploadId``."""
         ),
     ),
+    settings: Settings = deps.settings,
 ):
     """Create or complete a multi-part upload.
 
@@ -136,7 +138,8 @@ async def multipart_upload(
 
     if uploads == "" and uploadId is None:
         # Means a new upload is requested
-        return await create_multipart_upload(s3, env, key)
+        metadata = extract_request_metadata(request, settings)
+        return await create_multipart_upload(s3, env, key, metadata)
 
     if uploads is None and uploadId:
         # Given an existing upload to complete
@@ -167,6 +170,7 @@ async def upload(
     partNumber: Optional[int] = Query(
         None, description="Part number, where multi-part upload is used."
     ),
+    settings: Settings = deps.settings,
 ):
     """Write to an object, either as a standalone operation or within a multi-part upload.
 
@@ -186,7 +190,8 @@ async def upload(
 
     if uploadId is None and partNumber is None:
         # Single-part upload
-        return await object_put(s3, env, key, request)
+        metadata = extract_request_metadata(request, settings)
+        return await object_put(s3, env, key, request, metadata)
 
     # If either is set, both must be set.
     assert uploadId and partNumber
@@ -196,7 +201,11 @@ async def upload(
 
 
 async def object_put(
-    s3: S3ClientWrapper, env: Environment, key: str, request: Request
+    s3: S3ClientWrapper,
+    env: Environment,
+    key: str,
+    request: Request,
+    metadata: Dict[str, str],
 ):
     # Single-part upload handler: entire object is written via one PUT.
     reader = RequestReader.get_reader(request)
@@ -209,6 +218,7 @@ async def object_put(
         Body=reader,
         ContentMD5=content_md5(request),
         ContentLength=int(request.headers["Content-Length"]),
+        Metadata=metadata,
     )
 
     return Response(headers={"ETag": response["ETag"]})
@@ -246,11 +256,14 @@ async def complete_multipart_upload(
 
 
 async def create_multipart_upload(
-    s3: S3ClientWrapper, env: Environment, key: str
+    s3: S3ClientWrapper,
+    env: Environment,
+    key: str,
+    metadata: Dict[str, str],
 ):
     validate_object_key(key)
 
-    response = await s3.create_multipart_upload(Bucket=env.bucket, Key=key)  # type: ignore
+    response = await s3.create_multipart_upload(Bucket=env.bucket, Key=key, Metadata=metadata)  # type: ignore
 
     return xml_response(
         "CreateMultipartUploadOutput",
@@ -362,4 +375,7 @@ async def head(
 
         return Response(status_code=code)
 
-    return Response(headers={"ETag": response["ETag"]})
+    headers = {"ETag": response["ETag"]}
+    for k, v in response["Metadata"].items():
+        headers["x-amz-meta-%s" % k] = v
+    return Response(headers=headers)
--- a/exodus_gw/settings.py
+++ b/exodus_gw/settings.py
@@ -49,6 +49,11 @@ class Settings(BaseSettings):
     and authorization).
     """
 
+    upload_meta_fields: Dict[str, str] = {}
+    """Permitted metadata field names for s3 uploads and their regex
+    for validation. E.g., "exodus-migration-md5": "^[0-9a-f]{32}$"
+    """
+
     log_config: Dict[str, Any] = {
         "version": 1,
         "incremental": True,
--- a/tests/routers/upload/test_head.py
+++ b/tests/routers/upload/test_head.py
@@ -1,3 +1,5 @@
+import json
+
 from botocore.exceptions import ClientError
 from fastapi.testclient import TestClient
 
@@ -6,10 +8,28 @@ from exodus_gw.main import app
 TEST_KEY = "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
 
 
-async def test_head(mock_aws_client, auth_header):
+async def test_head(mock_aws_client, auth_header, monkeypatch):
     """Head request is delegated correctly to S3."""
 
-    mock_aws_client.head_object.return_value = {"ETag": "a1b2c3"}
+    monkeypatch.setenv(
+        "EXODUS_GW_UPLOAD_META_FIELDS",
+        json.dumps(
+            {
+                # md5sum of content being migrated
+                "exodus-migration-md5": "^[0-9a-f]{32}$",
+                # original source of migrated content
+                "exodus-migration-src": "^.{1,2000}$",
+            }
+        ),
+    )
+
+    mock_aws_client.head_object.return_value = {
+        "ETag": "a1b2c3",
+        "Metadata": {
+            "exodus-migration-md5": "94e19d5d30b26306167e9e7bae6b28fd",
+            "exodus-migration-src": "original/source",
+        },
+    }
 
     with TestClient(app) as client:
         r = client.head(
@@ -18,7 +38,12 @@ async def test_head(mock_aws_client, auth_header):
         )
 
     assert r.ok
-    assert r.headers["etag"] == "a1b2c3"
+    assert r.headers == {
+        "etag": "a1b2c3",
+        "x-amz-meta-exodus-migration-md5": "94e19d5d30b26306167e9e7bae6b28fd",
+        "x-amz-meta-exodus-migration-src": "original/source",
+        "content-length": "0",
+    }
 
 
 async def test_head_nonexistent_key(mock_aws_client, auth_header):
--- a/tests/routers/upload/test_upload.py
+++ b/tests/routers/upload/test_upload.py
@@ -1,3 +1,4 @@
+import mock
 import pytest
 from fastapi.testclient import TestClient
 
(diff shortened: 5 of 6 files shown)


exodus_gw/routers/upload.py
```diff
@@ -13,7 +13,7 @@
 
 - Most optional arguments are not supported.
 
-- All `x-amz-*` headers are omitted from responses.
+- All `x-amz-*` headers, other than metadata, are omitted from responses.
```
negillett, line 16: Not sure this is true...or where we'd see x-amz-* headers in the context of the AWS Python SDK.

exodus_gw/aws/util.py
```diff
@@ -17,6 +19,20 @@ def validate_object_key(key: str):
         raise HTTPException(400, detail="Invalid object key: '%s'" % key)
 
 
+def validate_metadata(metadata: Dict[str, str]):
+    valid_meta_fields = load_settings().upload_meta_fields
```
rohanpm, line 23: It doesn't seem like this function should have to load settings.  By the time this function is called, settings have already been loaded and ideally we don't reload them unnecessarily for every request.

Endpoints can obtain a loaded settings object by declaring a parameter like this:

```
settings: Settings = deps.settings
```

exodus_gw/routers/upload.py
```diff
@@ -97,6 +107,12 @@ async def multipart_upload(
     env: Environment = deps.env,
     s3: S3ClientWrapper = deps.s3_client,
     key: str = Path(..., description="S3 object key"),
+    exodus_migration_md5: Optional[str] = Header(
```
rohanpm, line 110: I don't think it should be done this way because it means you can't support new fields by just adding them into settings.  It seems to fail the point in the AC

> The specific set of metadata fields which can be set, and their valid values, are configurable by the exodus-gw service owners without requiring code changes

Can we please make the full set of supported headers based only on the settings, so that we can change it by changing the settings and not having to come back and revise this code?  Look at *all* headers on the request and extract all headers starting with "x-amz-meta-", then validate them.

exodus_gw/routers/upload.py
```diff
@@ -362,4 +403,9 @@ async def head(
 
         return Response(status_code=code)
 
-    return Response(headers={"ETag": response["ETag"]})
+    headers = {"ETag": response["ETag"]}
+
+    if "Metadata" in response:
```
rohanpm, line 408: It's not the right way to return it since it's not S3-compatible.

The AC said:

> if the object has any x-amz-meta-* headers returned from S3, those are returned to the client.

It needs to put the metadata into x-amz-meta-* headers in the response, because that's what S3 clients like boto are going to be looking for.  As well as the test you did in RHELDST-14181 comments using http command, you could try a script using boto library as a client of exodus-gw - it would probably show up that whatever you returned here did not make it into the "Metadata" returned by boto, because it's not looking for a header called "Metadata".

Logic here should be something along the lines of:

```
for key, value in response["Metadata"]:
  headers[f"x-amz-meta-{key}"] = value
```

exodus_gw/settings.py
```diff
@@ -49,6 +49,14 @@ class Settings(BaseSettings):
     and authorization).
     """
 
+    upload_meta_fields: Dict[str, str] = {
```
rohanpm, line 52: Could we please make this an empty default?  I think this is a case where it makes more sense for these particular settings to live only in our specific deployment of the service (i.e. maintain them only in ansible).  exodus-gw is going to live "forever" while exodus migration is a one-time thing.  So far, exodus-gw has a general API which does not know anything about the migration process, and it'd be good to keep it that way.

I'm also worried that, if we don't force ourselves to use ansible to adjust these settings, then we won't figure out the correct way to do that and it'll be more complicated than expected if we ever have to adjust them.

tests/routers/upload/test_head.py
```diff
@@ -18,7 +28,8 @@ async def test_head(mock_aws_client, auth_header):
         )
 
     assert r.ok
-    assert r.headers["etag"] == "a1b2c3"
+    assert r.headers["Etag"] == "a1b2c3"
+    assert r.headers["Metadata"] == json.dumps(metadata)
```
rohanpm, line 32: This should instead be seeing headers like x-amz-meta-exodus-migration-md5.

exodus_gw/routers/upload.py
rohanpm: The boto S3 client doesn't seem to give you direct access to headers.  x-amz-meta-* headers have presumably been parsed into the 'Metadata' field on response objects.

rohanpm (changes requested):

exodus_gw/aws/util.py
negillett: Ah, yes, thanks.

exodus_gw/routers/upload.py
negillett: Ah, okay. I wasn't sure they'd be included in the request object at this point. Thanks.

exodus_gw/routers/upload.py
```diff
@@ -362,4 +375,11 @@ async def head(
 
         return Response(status_code=code)
 
-    return Response(headers={"ETag": response["ETag"]})
+    headers = {"ETag": response["ETag"]}
+
+    for k, v in response["Metadata"].items():
+        if not k.startswith("x-amz-meta"):
```
rohanpm, line 381: This check seems suspicious, as surely either (1) all of the metadata keys returned in the response Metadata will include the leading x-amz-meta portion taken from the header (i.e. boto doesn't strip it), or (2) none of them do.  

When I posted a code snippet in the earlier code review, I assumed the situation was (2) since that seems more intuitive to me, but I don't know.

If the situation is (1) then it doesn't seem like rewriting the keys is needed.  (If you've seen that Metadata usually contains x-amz-meta prefix but you're worried about the possibility of it maybe not guaranteed to work that way, it seems more appropriate to warn and filter out unexpected fields, rather than rewriting them.)

If the situation is (2) then this check introduces a really weird behavior: if you store a metadata field which itself starts with x-amz-meta, meaning you would have to send a header like "x-amz-meta-x-amz-meta-foo", it seems to store OK.  But then when you later fetch that object, one level of "x-amz-meta" is stripped from the beginning of the header, so the field name doesn't survive a round-trip.  This would be an unexpected difference between exodus-gw behavior and S3 behavior.

Can you confirm which case is it?  Do the keys here include x-amz-meta prefix or not?

exodus_gw/aws/util.py
```diff
@@ -1,15 +1,42 @@
...
 from defusedxml.ElementTree import fromstring
-from fastapi import HTTPException, Response
+from fastapi import HTTPException, Request, Response
+
+from ..settings import Settings
 
 LOG = logging.getLogger("exodus-gw")
 
 
+def extract_request_metadata(request: Request, settings: Settings):
```
rohanpm, line 15: Minor: somewhere in this file it would be nice to have a link to the AWS docs mentioning S3 metadata.  To make it clear this is not something we've made up ourselves.

rohanpm (changes requested):

exodus_gw/routers/upload.py
negillett: Our metadata headers will have the 'x-amz-meta-' prefix because they have to in order to be picked out as metadata, per util.extract_request_metadata, and I've no reason yet to think boto would strip a substring from those keys. So, the situation should always be (1). I'll replace the key rewrite with a warning about the unexpected  metadata.

exodus_gw/settings.py
```diff
@@ -49,6 +49,15 @@ class Settings(BaseSettings):
     and authorization).
     """
 
+    upload_meta_fields: Dict[str, str] = {}
+    """Permitted metadata field names for s3 uploads and their regex for validation.
+    Must be prefixed with "x-amz-meta-".
```
rohanpm, line 54: Can we please drop the x-amz-meta- prefix in the doc and examples here?

The metadata *field names* don't contain "x-amz-meta-" prefix.  Only the HTTP headers used to serialize them contain this.  Seeing as clients usually won't have to deal with that detail, and it's redundant to require the same prefix on every key here anyway, it seems more sensible to leave it out.

tests/routers/upload/test_head.py
```diff
@@ -6,10 +8,28 @@
...
+async def test_head(mock_aws_client, auth_header, monkeypatch):
     """Head request is delegated correctly to S3."""
 
-    mock_aws_client.head_object.return_value = {"ETag": "a1b2c3"}
+    monkeypatch.setenv(
+        "EXODUS_GW_UPLOAD_META_FIELDS",
+        json.dumps(
+            {
+                # md5sum of content being migrated
+                "x-amz-meta-exodus-migration-md5": "^[0-9a-f]{32}$",
```
rohanpm, line 19: Per my other comment on Settings, I don't think the x-amz-meta- prefix should be included here.

tests/routers/upload/test_upload.py
```diff
@@ -1,27 +1,49 @@
...
-            headers=auth_header(roles=["test-blob-uploader"]),
+            headers={**auth_header(roles=["test-blob-uploader"]), **metadata},
         )
 
     # It should succeed
     assert r.ok
 
-    # It should return the ETag
+    # It should return the correct headers
     assert r.headers["ETag"] == "a1b2c3"
```
rohanpm, line 48: I'm surprised it's currently missing in the test, but can we please add at least one assertion that mock_aws_client.put_object was called and that the Metadata argument is correct?

I believe that would have shown up the write-side bug with x-amz-meta-, as it would have shown that the Metadata dict contained e.g. x-amz-meta-exodus-migration-md5 when it should really have contained exodus-migration-md5.

rohanpm (changes requested):

exodus_gw/aws/util.py
```diff
@@ -1,15 +1,46 @@
...
+    # Any headers prefixed with "x-amz-meta-" will be picked out as
+    # metadata for s3 upload.
+    #
+    # https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html
+    metadata = {}
+    for k, v in request.headers.items():
+        if k.startswith("x-amz-meta-"):
+            metadata[k] = v
```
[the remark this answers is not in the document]
negillett: Oh, I see. We don't need to keep the prefix because, once uploaded, s3 will prepend it for us. And, when returned by headObject, this prefix is stripped. Perfect. Got it.
This behavior isn't mentioned anywhere in the documentation that I've seen.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html?highlight=put_object#S3.Client.put_object
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html?highlight=put_object#S3.Client.head_object

negillett:
```
$ http --cert /tmp/rhsm-pulp/cert.pem --cert-key /tmp/rhsm-pulp/key.pem PUT https://exodus-gw-c3i-ngillett-test.apps.ocp-c1.prod.psi.redhat.com/upload/test/5282fe9b67bbe58f639fd1b2446b280a136052f7fad60281dcd7dec5b02bc71e x-amz-meta-exodus-migration-md5:7957cc6e33d7b8277c7e5e324b95503b x-amz-meta-exodus-migration-src:original/source
HTTP/1.1 200 OK
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Connection: keep-alive
Expires: 0
Pragma: no-cache
Referrer-Policy: no-referrer
Strict-Transport-Security: max-age=31536000 ; includeSubDomains
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-XSS-Protection: 1 ; mode=block
content-length: 0
date: Thu, 17 Nov 2022 21:21:49 GMT
etag: "d41d8cd98f00b204e9800998ecf8427e"
server: uvicorn



$  http --cert /tmp/rhsm-pulp/cert.pem --cert-key /tmp/rhsm-pulp/key.pem HEAD https://exodus-gw-c3i-ngillett-test.apps.ocp-c1.prod.psi.redhat.com/upload/test/5282fe9b67bbe58f639fd1b2446b280a136052f7fad60281dcd7dec5b02bc71e
HTTP/1.1 200 OK
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Connection: keep-alive
Expires: 0
Pragma: no-cache
Referrer-Policy: no-referrer
Strict-Transport-Security: max-age=31536000 ; includeSubDomains
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-XSS-Protection: 1 ; mode=block
content-length: 0
date: Thu, 17 Nov 2022 21:21:53 GMT
etag: "d41d8cd98f00b204e9800998ecf8427e"
exodus-migration-md5: 7957cc6e33d7b8277c7e5e324b95503b
exodus-migration-src: original/source
server: uvicorn
```

exodus_gw/aws/util.py
```diff
@@ -1,15 +1,47 @@
...
 
+def extract_request_metadata(request: Request, settings: Settings):
+    # Any headers prefixed with "x-amz-meta-" will be picked out as
+    # metadata for s3 upload.
+    #
+    # https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html
+    metadata = {}
+    for k, v in request.headers.items():
+        if k.startswith("x-amz-meta-"):
+            metadata[k.replace("x-amz-meta-", "")] = v
```
rohanpm, line 23: This is slightly incorrect since it will replace *all* instances of x-amz-meta-, not just the leading prefix.  Though it would be very weird to include an x-amz-meta- inside the metadata field name, it would be legal to do so and we shouldn't silently strip it out if so.

It can be fixed to strip just the leading prefix like this:

```suggestion
            metadata[k.replace("x-amz-meta-", "", 1)] = v
```

exodus_gw/routers/upload.py
```diff
@@ -362,4 +375,4 @@ async def head(
 
         return Response(status_code=code)
 
-    return Response(headers={"ETag": response["ETag"]})
+    return Response(headers={"ETag": response["ETag"], **response["Metadata"]})
```
rohanpm, line 378: This doesn't seem right:

- we're returning HTTP headers here, so metadata headers should have the `x-amz-meta-` prefix
- but `response["Metadata"]` isn't going to have that

So I guess if there is a metadata field "foo" this is going to wrongly return a header "foo" rather than "x-amz-meta-foo".

I think you need to explicitly add the x-amz-meta- prefix onto each element in response["Metadata"].

tests/routers/upload/test_head.py
```diff
@@ -18,7 +38,12 @@ async def test_head(mock_aws_client, auth_header):
         )
 
     assert r.ok
-    assert r.headers["etag"] == "a1b2c3"
+    assert r.headers == {
+        "etag": "a1b2c3",
+        "exodus-migration-md5": "94e19d5d30b26306167e9e7bae6b28fd",
```
rohanpm, line 43: The expected keys here aren't right, they should have x-amz-meta- prefix.  These are HTTP headers, so the header name is equal to the metadata field name with "x-amz-meta-" prepended.

tests/routers/upload/test_upload.py
rohanpm: I wasn't asking to check the return value, I was asking to check whether the correct metadata ever makes it into S3.

In object_put  you added:

```
+        Metadata=metadata,
```

My point is, I believe this is entirely uncovered by the test.  If you deleted that line or if you passed incorrect metadata, I'm pretty sure this test would still pass right now, meaning we are not actually testing whether we ever passed the metadata into S3.

It should be possible to verify that correct Metadata was passed by either looking at 
`mock_aws_client.put_object.mock_calls` or using one of the `mock_aws_client.put_object.assert_called_*` methods.

rohanpm (changes requested):

tests/routers/upload/test_head.py
negillett: s3 strips that prefix for us on HEAD requests apparently; seen in my test commented above.
> When you retrieve the object using the SOAP API, the prefix is removed, regardless of which API you used to upload the object.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata

exodus_gw/routers/upload.py
negillett: Hmm. Okay. So we wan't the "x-amz-meta-" prefix. Is there a reason? I thought they were just used internally by S3 and by us to determine what should be stored as metadata with an object.

tests/routers/upload/test_head.py
negillett: Okay. We want to add the prefix back ourselves. Fixed.

exodus_gw/routers/upload.py
rohanpm: Yes, we want the x-amz-meta- prefix on HTTP headers, because:

- that's how the S3 API works
- that's what all S3 clients are going to be looking for (e.g. boto will not see the metadata if the headers are missing that prefix)
- if there's no prefix then there's issues with clashes.  What if a metadata field had the same name as a standard header like "ETag", "Content-Type" or similar?

tests/routers/upload/test_head.py
rohanpm: I don't think this is right.  Actually the above doc says:

> When you retrieve the object using the REST API, this prefix is returned.

We are using the REST API, not the SOAP API.

Also your above test comment showed a query to exodus-gw, not to S3 or localstack.  Your test only showed that exodus-gw behavior is buggy.

scripts/launch-devenv
```diff
@@ -34,7 +34,7 @@ start_services(){
 launch_log_terminal(){
     if [ $LAUNCH_LOGS == true ]; then
         if command -v gnome-terminal; then
-            gnome-terminal -- journalctl --user "--unit=exodus-gw-*" -f
+            gnome-terminal -- journalctl --user "--unit=exodus-gw-*" -f > gw-dev-env.txt
```
rohanpm, line 37: Hmm, I personally don't use this script and it won't affect my review either way, but I'm wondering if it was intentional to include this?

If you did intend to add something like this, instead of `>` it's possible that something like `| tee -a gw-dev-env.txt` makes more sense so that the logs can still be seen on the terminal and so that previously generated logs don't get removed every time you launch.

rohanpm (changes requested):
Looks like a file named 'tee' has been incorrectly added to the commit, please remove that.  Didn't find any other issues.

rohanpm (approved):

PR #491, opened 2022-12-12
Add noload option to cleanup queries [RHELDST-14737]

negillett:
Previously, when loading Tasks and Publishes for scheduled cleanup operations, any children were automatically loaded as well, using more memory than necessary and causing crashes by exceeding the limits set in OpenShift.

This commit adds the noload option to cleanup queries with keys="*" so that no children are ever loaded.

Because children aren't loaded, publish items are no longer deleted automatically. So a delete query was added for items when cleaning up old publishes.

### DIFF
--- a/exodus_gw/worker/scheduled.py
+++ b/exodus_gw/worker/scheduled.py
@@ -2,10 +2,10 @@ import logging
 from datetime import datetime, timedelta
 
 import dramatiq
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session, noload
 
 from exodus_gw.database import db_engine
-from exodus_gw.models import Publish, Task
+from exodus_gw.models import Item, Publish, Task
 from exodus_gw.schemas import PublishStates, TaskStates
 from exodus_gw.settings import Settings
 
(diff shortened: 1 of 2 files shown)


negillett:
Safety scanner is blocking for pre-versions of packages... Would using pre-versions not in itself be a safety risk?

rohanpm (dismissed):

rohanpm:
> Safety scanner is blocking for pre-versions of packages... Would using pre-versions not in itself be a safety risk?

Furthermore, sqlalchemy 1 => 2 seems like a big enough deal that there is a migration guide written for it, https://docs.sqlalchemy.org/en/20/changelog/migration_20.html, so I'm guessing we can't do it by the usual automated updates.

I think the right thing to do in a case like this is something like https://github.com/release-engineering/exodus-gw/pull/492 , acknowledge this particular vulnerability and give ourselves a few months to handle the upgrade, filed separately.

negillett:
Rebased

negillett:
Carrying over @rohanpm's +1 from before the rebase

PR #494, opened 2022-12-15
Add GET publish endpoint [RHELDST-14412]

negillett:
This commit implements a new endpoint to retrieve a publish from the service database. No items are returned with the publish as there is currently no use case for them and implementation would be non-trival.

### DIFF
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -76,7 +76,7 @@ from typing import Dict, List, Union
 from uuid import UUID, uuid4
 
 from fastapi import APIRouter, Body, HTTPException, Query
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session, noload
 
 from .. import auth, deps, models, schemas, worker
 from ..settings import Environment, Settings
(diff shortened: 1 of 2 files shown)


exodus_gw/routers/publish.py
```diff
@@ -287,3 +287,61 @@ def commit_publish(
...
+                        "links": {
+                            "self": "/prod/publish/497f6eca-6276-4993-bfeb-53cbbbba6f08",
+                            "commit": "/prod/publish/497f6eca-6276-4993-bfeb-53cbbbba6f08/commit",
+                        },
+                        "items": [],
+                    }
+                }
+            },
+        },
+        422: {
```
rohanpm, line 313: Seems like this should be 404, not 422.

tests/routers/test_publish.py
```diff
@@ -779,3 +779,73 @@ def test_commit_no_publish(auth_header):
...
+    """
+
+    with TestClient(app) as client:
+        r = client.get(
+            "/test/publish/%s" % fake_publish.id,
+            headers=auth_header(roles=["test-publisher"]),
+        )
+
+    # It should have failed
+    assert r.status_code == 422
```
rohanpm, line 847: Should be 404 and not 422.

rohanpm (changes requested):

exodus_gw/routers/publish.py
negillett: @rohanpm, the Jira specified 422 but 404 makes sense too.

rohanpm: I see, I think the JIRA was wrong, will comment there.

rohanpm (approved):

exodus_gw/routers/publish.py
```diff
@@ -287,3 +287,62 @@ def commit_publish(
...
+    This is achieved using the noload() query option.
+    """
+
+    db_publish = (
+        db.query(models.Publish)
+        .options(noload("items"))
+        .filter(
+            # Since sub-environments share some resources, filter for env as well.
+            models.Publish.id == publish_id,
+            models.Publish.env == env.name,
```
rohanpm, line 338: The other endpoints aren't using 'env' during lookup.  Noticing that made me realize that the other endpoints are buggy, and they should also all be using 'env' like this one is.  Filed a new issue RHELDST-15336.

negillett: 👍🏻 Thanks

PR #501, opened 2023-01-25
Make commit_publish idempotent [RHELDST-15029]

negillett:
When commit_publish is called multiple times or retried, it will now return the task to which the publish was previously assigned rather than raising an exception (because the publish is not currently "PENDING").

### DIFF
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -261,6 +261,15 @@ def commit_publish(
         )
 
     if db_publish.state != "PENDING":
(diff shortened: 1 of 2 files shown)


negillett:
Yet another Bandit issue. It complained about constraints like package[subpackage] so I removed those but it's still complaining.

negillett:
Another magic pip-compile commit rebase.

crungehottman (approved):

emanuelelevo (approved):

PR #504, opened 2023-01-30
Make update_publish_items idempotent by no duplicating web_uri [RHELDST-15028]

emanuelelevo:
When update_publish_items is called, if the item's web_uri already exists on that publish, it will now update the item in db rather then adding a new item.

### DIFF
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -76,6 +76,7 @@ from typing import Dict, List, Union
 from uuid import UUID, uuid4
 
 from fastapi import APIRouter, Body, HTTPException, Query
+from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import Session, noload
 
 from .. import auth, deps, models, schemas, worker
@@ -193,8 +194,24 @@ def update_publish_items(
             % (db_publish.id, db_publish.state),
         )
 
-    for item in items:
-        db.add(models.Item(**item.dict(), publish_id=db_publish.id))
+    # Convert the list into dict and update each dict with a publish_id.
+    items_data = [
+        {**item.dict(), "publish_id": db_publish.id} for item in items
+    ]
+
+    LOG.debug("Adding %s items into '%s'", len(items_data), db_publish.id)
+
+    statement = insert(models.Item).values(items_data)
+
+    # Update all target table columns, except for the primary_key column.
+    update_dict = {c.name: c for c in statement.excluded if not c.primary_key}
+
+    update_statement = statement.on_conflict_do_update(
+        index_elements=["publish_id", "web_uri"],
+        set_=update_dict,
+    )
+
+    db.execute(update_statement)
 
     return {}
 
--- a/exodus_gw/migrations/versions/6461bad8ed91_.py
+++ b/exodus_gw/migrations/versions/6461bad8ed91_.py
@@ -1,0 +1,28 @@
+"""add unique constraint on (publish_id, web_uri)
+
+Revision ID: 6461bad8ed91
+Revises: 5bd0b38df850
+Create Date: 2023-02-02 14:02:05.437573
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "6461bad8ed91"
+down_revision = "5bd0b38df850"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("items") as batch_op:
+        batch_op.create_unique_constraint(
+            "items_publish_id_web_uri_key", ["publish_id", "web_uri"]
+        )
+
+
+def downgrade():
+    with op.batch_alter_table("items") as batch_op:
+        batch_op.drop_constraint(
+            "items_publish_id_web_uri_key", type_="unique"
+        )
--- a/tests/routers/test_publish.py
+++ b/tests/routers/test_publish.py
@@ -187,13 +187,14 @@ def test_update_publish_items_path_normalization(db, auth_header):
     # Publish object should now have matching items
     db.refresh(publish)
 
+    items = sorted(publish.items, key=lambda item: item.link_to)
     item_dicts = [
         {
             "web_uri": item.web_uri,
             "object_key": item.object_key,
             "link_to": item.link_to,
         }
-        for item in publish.items
+        for item in items
     ]
 
     # Should have stored normalized web_uri and link_to paths
@@ -276,6 +277,74 @@ def test_update_publish_items_no_uri(db, auth_header):
     # It should have failed with 400
     assert r.status_code == 400
     assert r.json() == {"detail": ["No URI: %s" % expected_item]}
+
+
+def test_update_publish_items_existing_uri(db, auth_header):
+    """PUTting an item which item's web_uri already exists creates expected objects in DB."""
+
+    publish_id = "11224567-e89b-12d3-a456-426614174000"
+
+    publish = Publish(
+        id=uuid.UUID("{%s}" % publish_id),
+        env="test",
+        state="PENDING",
+        items=[
+            Item(
+                web_uri="/uri1",
+                object_key="1" * 64,
+                publish_id=publish_id,
+            ),
+            Item(
+                web_uri="/uri2",
+                object_key="2" * 64,
+                publish_id=publish_id,
+            ),
+        ],
+    )
+
+    with TestClient(app) as client:
+        # Ensure a publish object exists
+        db.add(publish)
+        db.commit()
+
+        # Try to add an item which item's web_uri already exists
+        r = client.put(
+            "/test/publish/%s" % publish_id,
+            json=[
+                {
+                    "web_uri": "/uri1",
+                    "object_key": "3" * 64,
+                },
+            ],
+            headers=auth_header(roles=["test-publisher"]),
+        )
+
+    # It should have succeeded
+    assert r.status_code == 200
+
+    # Publish object should now have matching items
+    db.refresh(publish)
+
+    items = sorted(publish.items, key=lambda item: item.web_uri)
+    item_dicts = [
+        {
+            "web_uri": item.web_uri,
+            "object_key": item.object_key,
+        }
+        for item in items
+    ]
+
+    # Should have stored exactly what we asked for
+    assert item_dicts == [
+        {
+            "web_uri": "/uri1",
+            "object_key": "3" * 64,
+        },
+        {
+            "web_uri": "/uri2",
+            "object_key": "2" * 64,
+        },
+    ]
 
 
 def test_update_publish_items_invalid_item(db, auth_header):
@@ -703,7 +772,7 @@ def test_commit_publish_linked_items(mock_commit, fake_publish, db):
 
     # Whole items
     item1 = Item(
-        web_uri="/some/path",
+        web_uri="/the/path",
         object_key="1" * 64,
         publish_id=fake_publish.id,
         link_to=None,  # It should be able to handle None/NULL link_to values...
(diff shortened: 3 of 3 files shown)


exodus_gw/routers/publish.py
```diff
@@ -194,7 +194,22 @@ def update_publish_items(
         )
 
     for item in items:
-        db.add(models.Item(**item.dict(), publish_id=db_publish.id))
+        # Check if the item's web_uri already exists in any of the current items on the publish.
+        web_uri = (
```
negillett, line 198: Can we change this to something like "existing_publish_item"?
It's misleading currently, as "web_uri" is just an attribute of the returned Item.

rohanpm (changes requested):

exodus_gw/routers/publish.py
```diff
@@ -194,7 +194,22 @@ def update_publish_items(
...
+            db.query(models.Item)
+            .filter(
+                models.Item.publish_id == db_publish.id,
```
rohanpm, line 201: I'm still nervous about the performance impact here.  I know you said you didn't see any impact from adding an index.  I believe that, just because our table would be pretty small in general.  So maybe the suggestion to add indexes was wrong.

Still, don't we update items in batches of 10,000 at once?  That means we need to do 10,000 queries here.  I'm not sure how long it will take, but it seems like it could be too slow.  I tried locally after populating my DB with 1 million items, and one of these requests takes about 50ms.  10,000 of them would take ~500 seconds, or about 8 minutes, which is too slow.  I don't know if our cloud postgres will be slower or faster than that, it's concerning though.

I'm OK if you want to go ahead with this as it is now and return to it later if problems show up (hopefully would be shown up by test_exodus_large_publish).  It wouldn't surprise me at all if we have to come back to it and refactor it so that it does a single bulk SELECT here, rather than one SELECT per item.

exodus_gw/routers/publish.py
```diff
@@ -193,8 +193,24 @@ def update_publish_items(
             % (db_publish.id, db_publish.state),
         )
 
+    # Get existing items web_uri on the publish.
+    existing_publish_items = {
```
rohanpm, line 197: And now I'm concerned about memory usage.  Remembering the goal of supporting a 500,000 item publish - that's how many items this will load into memory at once.  I don't think there is any other code in exodus-gw which loads all those items on a publish at once, so this could easily be introducing new memory issues.

I guess we can watch for failures in the "large publish" test after submitting this though.

rohanpm: Actually, let me add one additional option which might have been overlooked and is probably the best practice in general for doing this kind of operation.

The table should have a uniqueness constraint on (publish_id, web_uri).  Then, when we "db.add", we should arrange for the INSERT statement to have the "ON CONFLICT DO UPDATE" modifier.  The DB will then automatically create a new row when needed, or update the existing row when needed.

I haven't done it with sqlalchemy before, but according to https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#orm-upsert-statements it is possible to do that when using sqlalchemy ORM, and also possible to do it with a bulk insert.  So it should be possible to do all of these inserts without having to read anything out of the DB other than db_publish.

rohanpm (changes requested):

exodus_gw/routers/publish.py
emanuelelevo: This is interesting and I had no idea the wrapper could do such things. I'll look into it.

emanuelelevo:
Unit test changes:
- `tests/routers/test_publish.py::test_update_publish_items_path_normalization`: assertion now works for sorted publish items.
- `tests/routers/test_publish.py::test_commit_publish_linked_items`: it was previously adding two items with the same web_uri for the same publish, since `web_uri="/some/path"` is already present under `fake_publish()`, returning error:
```
E       sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: items.publish_id, items.web_uri
E       [SQL: INSERT INTO items (id, web_uri, object_key, content_type, link_to, publish_id) VALUES (?, ?, ?, ?, ?, ?)]
E       [parameters: ('933fcfef-0119-4153-9f2d-608dab5a6d32', '/some/path', '1111111111111111111111111111111111111111111111111111111111111111', 'some type', None, '123e4567-e89b-12d3-a456-426614174000')]
E       (Background on this error at: https://sqlalche.me/e/14/gkpj)

../../../.pyenv/versions/3.9.12/envs/exodus-gw-3.9.12/lib/python3.9/site-packages/sqlalchemy/engine/default.py:736: IntegrityError
```

negillett (approved):

exodus_gw/routers/publish.py
```diff
@@ -76,6 +76,7 @@
 from uuid import UUID, uuid4
 
 from fastapi import APIRouter, Body, HTTPException, Query
+from sqlalchemy.dialects.postgresql import insert
```
rohanpm, line 79: I'm a little surprised that this passes tests, because the tests in this repo aren't using postgres but instead sqlite.

I guess that it works because in this case, sqlite and postgres seem to compile to exactly the same SQL for an "on conflict do update".

We should probably be ensuring that we use the appropriate dialect for the DB we're connecting to though.
Could we please arrange that so it uses postgresql dialect when connected to postgres and sqlite dialect when connected to sqlite?

I don't find any guidance about this in sqlalchemy docs but I can think of some ways to do it, like here:

1) Use db.connection().dialect.name to try to load "any" dialect

```
import sqlalchemy.dialects

def dialect_for_db(db: Session):
    return getattr(sqlalchemy.dialects, db.connection().dialect.name)
```

2) Hardcode support for just postgres and sqlite based on the URL

```
def dialect_for_db(db: Session):
    url = db.connection().engine.url
    if url.startswith('sqlite://'):
        import sqlalchemy.dialects.sqlite
        return sqlalchemy.dialects.sqlite
    import sqlalchemy.dialects.postgresql
    return sqlalchemy.dialects.postgresql
```

Either way, a helper function like that could be tested by a small test providing a mock db Session.

exodus_gw/routers/publish.py
```diff
@@ -194,7 +195,21 @@ def update_publish_items(
...
+        update_dict = {
+            c.name: c for c in statement.excluded if not c.primary_key
+        }
+
+        update_statement = statement.on_conflict_do_update(
+            index_elements=["publish_id", "web_uri"],
+            set_=update_dict,
+        )
+
+        db.execute(update_statement)
```
rohanpm, line 212: This seems like it might introduce a significant performance regression, because as I understand it:

Prior to this change: `db.add` does not actually interact with remote DB, it just puts the object on a "to be inserted at next flush" list.  Then just before commit occurs later, all of these objects are flushed to the DB, which should allow for multiple inserts to happen in a single statement.

With this change: db.execute really does execute an SQL statement on the DB.  So we are changing the inserts from potentially a single SQL statement handling 10,000 rows, to 10,000 insert statements handling 1 row each.

Is it possible to make it so there's only a single db.execute?  I feel like that should be relatively straightforward, because:

- in insert(models.Item).values(...), the input to "values" can be a list of dicts rather than just a single dict.  So it could contain all 10,000 rows right there.
- the update_dict is not referring to specific values but just referencing back to the columns being updated.  So I think it could be used for a single query updating multiple rows without a problem.

Maybe it works with the following minor change:

- change the insert statement to:

```
insert(...).values([
    {**item.dict(), publish_id=db_publish.id}
    for item in items
])
```

- drop the `for item in items` loop

rohanpm (changes requested):

exodus_gw/routers/publish.py
emanuelelevo: Yes, I've noticed that and I was using it on some previous commits https://github.com/release-engineering/exodus-gw/compare/6f4485e923974b58feb61485a6460fa6a3d93560..76bea1f773ec87f2f0da67e29c293d42275f61f1 . I've brought it back.

exodus_gw/routers/publish.py
emanuelelevo: Implementing this is giving me all sort of errors like:
```
[2023-02-06 10:25:52 +0100] [338071] [INFO] Scheduled actor cleanup uses message f630bd10-14b0-5f04-bda1-5642b9933dae
ImportError while loading conftest '/home/elevorin/src/github/exodus-gw/tests/conftest.py'.
tests/conftest.py:13: in <module>
    from exodus_gw import database, main, models, settings  # noqa
exodus_gw/main.py:76: in <module>
    from .routers import cdn, deploy, publish, service, upload
exodus_gw/routers/publish.py:138: in <module>
    def dialect_for_db(db: Session):
.tox/py39/lib/python3.9/site-packages/fastapi/routing.py:638: in decorator
    self.add_api_route(
.tox/py39/lib/python3.9/site-packages/fastapi/routing.py:577: in add_api_route
    route = route_class(
.tox/py39/lib/python3.9/site-packages/fastapi/routing.py:446: in __init__
    self.dependant = get_dependant(path=self.path_format, call=self.endpoint)
.tox/py39/lib/python3.9/site-packages/fastapi/dependencies/utils.py:309: in get_dependant
    param_field = get_param_field(
.tox/py39/lib/python3.9/site-packages/fastapi/dependencies/utils.py:403: in get_param_field
    field = create_response_field(
.tox/py39/lib/python3.9/site-packages/fastapi/utils.py:90: in create_response_field
    raise fastapi.exceptions.FastAPIError(
E   fastapi.exceptions.FastAPIError: Invalid args for response field! Hint: check that <class 'sqlalchemy.orm.session.Session'> is a valid Pydantic field type. If you are using a return type annotation that is not a valid Pydantic field (e.g. Union[Response, dict, None]) you can disable generating the response model from the type annotation with the path operation decorator parameter response_model=None. Read more: https://fastapi.tiangolo.com/tutorial/response-model/
ERROR: InvocationError for command /home/elevorin/src/github/exodus-gw/.tox/py39/bin/pytest (exited with code 4)
```

So, doing it right does seem a ticket by its own. 
Moreover, other parts of the code are using `from sqlalchemy.dialects.postgresql import something` and not just this change.

emanuelelevo: Please file a ticket for it.

exodus_gw/routers/publish.py
```diff
@@ -194,7 +195,21 @@ def update_publish_items(
...
+        statement = insert(models.Item).values(
+            **item.dict(), publish_id=db_publish.id
+        )
+
+        # Update all target table columns, except for the primary_key column.
+        update_dict = {
+            c.name: c for c in statement.excluded if not c.primary_key
+        }
+
+        update_statement = statement.on_conflict_do_update(
```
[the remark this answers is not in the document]
emanuelelevo: Yes, it's an upsert. I won't rename anything since the entire function is called `update_publish_items`.

negillett (approved):

exodus_gw/routers/publish.py
rohanpm: Well, it's hard to say for sure without seeing the code but it very much looks like that error was caused by `def dialect_for_db` having some `@router.post` or other router decorator applied to it, which doesn't make sense since it's not an endpoint for the web app.

You are right that `from sqlalchemy.dialects.postgresql` appears elsewhere in the code, but if you look closely they all already have accounted for the sqlite case as well.  Those other imports are for JSONB and UUID types.  If you look in `exodus_gw/models/sqlite_compat.py` you'll see that code was written to make that work with sqlite.

I'm not going to insist that this has to be handled in this pull request, I'm approving it anyway, but given that the above two points seem to be simple misunderstandings maybe you're willing to have another try getting it to work?

rohanpm (approved):

exodus_gw/routers/publish.py
emanuelelevo: Don't take me wrong, I didn't even really try on the first place. I do understand that the above should be implemented but It can be done on a separate ticket. We have been grooming/scoring tickets and this ticket has already took longer then expected.

rohanpm: I did approve it already so that you'd be free to submit it if you don't want to get into this issue right now.  In that case I would probably have a quick try of implementing the suggested change myself after you submit this.

PR #505, opened 2023-01-31
Add 'env' to publish queries [RHELDST-15336]

negillett:
Previously, when querying publishes from the service db, we only filtered by publish IDs. This commit adds env to the query filter to ensure we're only manipulating publishes belonging to the service environment used.

### DIFF
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -177,7 +177,10 @@ def update_publish_items(
     db_publish = (
         db.query(models.Publish)
         .with_for_update()
-        .filter(models.Publish.id == publish_id)
+        .filter(
+            models.Publish.id == publish_id,
+            models.Publish.env == env.name,
+        )
         .first()
     )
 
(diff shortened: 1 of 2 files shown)


exodus_gw/routers/publish.py
```diff
@@ -176,8 +176,11 @@ def update_publish_items(
 
     db_publish = (
         db.query(models.Publish)
-        .with_for_update()
```
rohanpm, line 179: Why are the "with_for_update" being removed?

I don't think they should be.  The point of that is that the records are locked until the current transaction ends.  This is how we ensure it's safe in the case of concurrent requests.

For example: we load this publish, then we later check `if db_publish.state != "PENDING":`.  Given that some time has passed after we load the object from the DB and before we do that check, how do we know the state is up-to-date? (e.g. there could have been another request concurrently which has started to commit the publish.)

The answer is that `.with_for_update()` grabs a lock, so the query here and the similar query in commit_publish won't both be able to get the same object at the same time - one of them will be forced to wait.  If we remove it then concurrent requests seem like they'll lead to corruption.

rohanpm (changes requested):

exodus_gw/routers/publish.py
negillett: Ah, this was a mistake. Thanks, @rohanpm

rohanpm (approved):

emanuelelevo (approved):

crungehottman (approved):

PR #514, opened 2023-02-28
sqlalchemy upgrade to 2.0 [RHELDST-15252]

dichn:
This commit aims to resolve the backwards-incompatible issues which are resulted from upgrading sqlalchemy to 2.0.

- Use sqlalchemy.types.Uuid and use as_uuid=False everywhere

- Replace declarative_base() with DeclarativeBase

- Replace Declarative use of Column with mapped_column()


rohanpm:
This change has to include an sqlalchemy version bump in requirements.in, requirements.txt.  Otherwise our CI and deployments are still trying to use sqlalchemy 1.x.

dichn:
re-pushed for sqlalchemy version bump in 
* requirements.in
* requirements.txt
* test-requirements.txt

exodus_gw/models/dramatiq.py
```diff
@@ -12,16 +16,16 @@ class DramatiqMessage(Base):
...
+        Uuid(as_uuid=False),
         primary_key=True,
     )
 
     # ID of consumer who has assigned this message to itself.
     # Null means message is not yet assigned.
     # Not a foreign key since consumers can disappear while leaving
     # their messages behind.
-    consumer_id = Column(String, nullable=True)
+    consumer_id: Mapped[str] = mapped_column(String, nullable=True)
```
rohanpm, line 28: For the columns where nullable=True, it would be possible to have a value of None, so instead of `Mapped[str]` should it be `Mapped[Optional[str]]` ?

dichn:
For the [union-attr] error that mypy raised, I made the following change to handle it according to this doc[1].
But it made the code less concise.
```
diff --git a/exodus_gw/worker/autoindex.py b/exodus_gw/worker/autoindex.py
index 68a1cbb..e6e3d6b 100644
--- a/exodus_gw/worker/autoindex.py
+++ b/exodus_gw/worker/autoindex.py
@@ -85,7 +85,9 @@ class AutoindexEnricher:
         self.env_name = env_name
         self.env = get_environment(env_name, settings)
         self.settings = settings
-        self.db = inspect(publish).session
+        inspection = inspect(publish)
+        assert inspection is not None
+        self.db = inspection.session
         self.item_query = self.db.query(Item).filter(
             Item.publish_id == publish.id
         )
diff --git a/exodus_gw/worker/deploy.py b/exodus_gw/worker/deploy.py
index 7f65b1a..3c84a3c 100644
--- a/exodus_gw/worker/deploy.py
+++ b/exodus_gw/worker/deploy.py
@@ -19,6 +19,8 @@ def complete_deploy_config_task(task_id: str):
     db = Session(bind=db_engine(settings))
     task = db.query(models.Task).filter(models.Task.id == task_id).first()
 
+    assert task is not None
+
     if task.state != "IN_PROGRESS":
         LOG.warning("Task %s in unexpected state, '%s'", task.id, task.state)
         return
@@ -41,6 +43,8 @@ def deploy_config(config: Dict[str, Any], env: str, from_date: str):
         .first()
     )
 
+    assert task is not None
+
     if task.state not in ("NOT_STARTED", "IN_PROGRESS"):
         LOG.warning("Task %s in unexpected state, '%s'", task.id, task.state)
         return

```

[1] https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-attribute-exists-in-each-union-item-union-attr

dichn:
Re-pushed for using `Optional[]` to imply all nullable parameters

exodus_gw/models/dramatiq.py
dichn: Yes. Thanks for pointing it out.

rohanpm:
> assert inspection is not None

I don't have a problem with adding assertions like that to help mypy, though I don't think you need to include "is not None".  Just `assert inspection` should work.

exodus_gw/worker/publish.py
```diff
@@ -192,10 +192,10 @@ def has_items(self) -> bool:
         LOG.debug("No items to write for publish %s", self.publish.id)
         return False
 
-    def _query_task(self, actor_msg_id: str) -> Task:
+    def _query_task(self, actor_msg_id: str):
         return self.db.query(Task).filter(Task.id == actor_msg_id).first()
 
-    def _query_publish(self, publish_id: str) -> Publish:
+    def _query_publish(self, publish_id: str):
```
dichn, line 198: This change is for resolving the mypy complains
```
exodus_gw/worker/publish.py:205: error: Incompatible return value type (got "Optional[Publish]", expected "Publish")  [return-value]
            return publish
                   ^~~~~~~
```

negillett (dismissed):

exodus_gw/dramatiq/consumer.py
```diff
@@ -90,8 +90,13 @@ def __reset_lost_messages(self, db):
             db.query(DramatiqMessage)
             .with_for_update(of=DramatiqMessage)
             .outerjoin(DramatiqMessage.consumer)
-            .filter(DramatiqMessage.consumer_id != None)
-            .filter(DramatiqConsumer.id == None)
+            .filter(
+                DramatiqMessage.consumer_id
+                != None  # disable pylint: disable=C0121
```
rohanpm, line 95: Given that sqlalchemy expects you to do this everywhere, I think it'd be better to just disable this pylint check for the whole project rather than having to put this on each occurrence.

In theory, just remove the `singleton-comparison` line in this repo's `.pylintrc`.

exodus_gw/models/publish.py
```diff
@@ -38,7 +31,9 @@ def resolve_links(self):
         ln_items = (
             db.query(Item)
             .filter(Item.publish_id == self.id)
-            .filter(func.coalesce(Item.link_to, "") != "")
+            .filter(
+                func.coalesce(Item.link_to, "") != ""  # pylint: disable=E1102
```
rohanpm, line 35: Any idea why this is needed?  What was the error message?

rohanpm (approved):

exodus_gw/models/publish.py
dichn: @rohanpm 
It's this error `exodus_gw/models/publish.py:34:20: E1102: func.coalesce is not callable (not-callable)`

rohanpm (approved):

dichn:
re-pushed for removing the singleton-comparison line in  .pylintrc

exodus_gw/models/publish.py
rohanpm: Hmm, I don't know why pylint is raising that complaint, but if tests are passing I'm willing to assume there is no actual bug in our code here.

rohanpm (approved):
The migration to sqlalchemy 2.x in theory allows removing the sqlite_uuid hack in sqlite_compat.py (because sqlalchemy.types.Uuid was introduced).  It would be nice to do that, but not strictly required in scope of this change.

PR #526, opened 2023-03-22
Include AWS credentials in upload API documentation

crungehottman:


exodus_gw/routers/upload.py
```diff
@@ -49,6 +49,10 @@
 # Prepare S3 resource pointing at exodus-gw
 s3 = boto3.resource('s3',
                     endpoint_url='https://exodus-gw.example.com/upload',
+                    # Provide AWS credentials to the boto client
```
rohanpm, line 52: I think these should be actual string values which can be just copy-pasted, not placeholders.  The values are not used for authentication here, we are simply forced to provide some values by boto (to prevent it looking in config files and complaining if they're missing).  So something like:

```
# In a typical setup using PKI auth, these values are not used during
# auth to exodus-gw, but boto client always insists on having *some* keys.
# Dummy values can be provided to prevent boto looking for credentials
# in config files.
aws_access_key_id="dummy",
aws_secret_access_key="dummy",
```

Explained a bit more on the gitlab comment for the related MR.

rohanpm (changes requested):

exodus_gw/routers/upload.py
crungehottman: Thanks!

negillett (dismissed):

emanuelelevo (dismissed):

rohanpm (dismissed):

rohanpm (commented):
Two scripts in `examples` directory configure a boto client, so it would be nice for those to be updated as well for consistency with the docs here.  (If it's not then those scripts probably have the same hidden requirement right now that you need to have `AWS_PROFILE` set and you need to have some credentials in config files.)

rohanpm (approved):

PR #528, opened 2023-03-27
Extend container's connection keep-alive to 15 seconds

negillett:
It was found that connections were being dropped between platform-sidecar and exodus-gw. The default keep-alive is 2 seconds and there's evidence of requests exceeding this limit. This commit extends the max time limit for requests to 15 seconds.


rohanpm:
I'll merge this now so we can see ASAP if it makes things more stable.

PR #538, opened 2023-04-21
Log progress info during DynamoDB writes

rohanpm:
Previously during DynamoDB writes we only generated log messages such as "Items successfully written" per batch, which doesn't help much understanding the progress.

Let's add some new logging which makes it easier to track the progress and performance. An example of the new log messages generated during commit (maximum of one log message per 5 seconds):

    Writing phase 1 items: 775 (of 10000) [ 8% ] [6.9 p/sec]

Inspired by performance issues observed testing against localstack. These logs should make it easy to directly compare the performance (items per second) between localstack and each AWS environment.

The old log message still exists but was downgraded to DEBUG.

### DIFF
--- a/exodus_gw/worker/progress.py
+++ b/exodus_gw/worker/progress.py
@@ -1,0 +1,83 @@
+import logging
+from threading import Lock
+from time import monotonic
+
+LOG = logging.getLogger("exodus-gw")
+
+
+class ProgressLogger:
+    """A helper to generate progress logs during long-running processes."""
+
+    def __init__(self, message: str, items_total: int, interval: float = 5.0):
+        """Construct a progress logger.
+
+        Arguments:
+            message
+                A prefix for each generated log message.
+                Should indicate the operation taking place, e.g. "Writing items"
+
+            items_total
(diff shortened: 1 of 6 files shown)


emanuelelevo (approved):

exodus_gw/worker/progress.py
```diff
@@ -0,0 +1,83 @@
...
+        self.message = message
+        self.lock = Lock()
+        self.items_total = items_total
+        self.items_processed = 0
+        self.start_time = monotonic()
+        self.last_write = 0.0
+        self.interval = interval
+
+    def adjust_total(self, increment: int):
+        """Add or subtract from the configured items_total.
```
negillett, line 36: Looks like this only increments

negillett (approved):

exodus_gw/worker/progress.py
crungehottman: If the `increment` parameter is a negative integer it will effectively decrement `items_total`

crungehottman (approved):

PR #563, opened 2023-07-25
Log backoffs on EndpointConnectionErrors [RHELDST-18585]

crungehottman:
The backoff library does not log retry attempts by default [1].

"By default, backoff and retry attempts are logged to the ‘backoff’
logger. By default, this logger is configured with a NullHandler, so
there will be nothing output unless you configure a handler.

[...]

It is also supported to specify a Logger (or LoggerAdapter) object
directly."

Now, the _batch_write backoff decorator has been configured to log
backoff events for any EndpointConnectionErrors encountered during
batch writes. This way, engineers can more easily identify when
retries due to EndpointConnectionErrors are attempted during batch
writes.

The backoff attempts are logged at DEBUG using the exodus-gw logger.

[1] https://pypi.org/project/backoff/

### DIFF
--- a/tests/aws/test_dynamodb.py
+++ b/tests/aws/test_dynamodb.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from datetime import datetime, timezone
 
 import mock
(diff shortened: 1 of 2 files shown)


negillett:
@crungehottman, is the log level for this configurable? This probably isn't something we want on the INFO level.

crungehottman:
> @crungehottman, is the log level for this configurable? This probably isn't something we want on the INFO level.

Took a bit of hunting down but yes, the log level is configurable. Now the backoff events are logged at DEBUG. Thanks!

tests/aws/test_dynamodb.py
```diff
@@ -198,13 +199,25 @@ def test_write_batch_excs(mock_boto3_client, fake_publish, delete, caplog):
...
     with mock.patch("time.sleep"):
         with pytest.raises(EndpointConnectionError):
             ddb = dynamodb.DynamoDB("test", Settings(), NOW_UTC)
             ddb.write_batch(fake_publish.items)
 
-    assert mock_boto3_client.batch_write_item.call_count == 20
+    p = re.compile(
+        r"Backing off _batch_write\(\.\.\.\) for [0-9]+[.]?[0-9]+s \(botocore\.exceptions\.EndpointConnectionError: Could not connect to the endpoint URL: \\\"fake-url\\\"\)"
+    )
+    num_logs = len(p.findall(caplog.text))
```
negillett, line 217: @crungehottman `num_logs` (and thereby `p`) don't look used

crungehottman: Thanks! The `p` var is used in the following line, but I removed `num_logs` here: https://github.com/release-engineering/exodus-gw/pull/565

PR #567, opened 2023-08-03
Do not log unauthenticated logins [RHELDST-19460]

crungehottman:
The OpenShift readiness and/or liveness probes request the '/healthcheck' endpoint every every few seconds.

Requests from unauthenticated users need not be logged. Thus, to reduce the number of logs produced by exodus-gw, exodus-gw should only log requests from authenticated users. This will eliminate noisy "login" logs caused by the readiness/liveness probes.

### DIFF
--- a/exodus_gw/auth.py
+++ b/exodus_gw/auth.py
@@ -137,10 +137,11 @@ async def log_login(
     roles: Set[str] = Depends(caller_roles),
     caller_name: str = Depends(caller_name),
 ):
-    LOG.info(
-        "Login: path=%s, user=%s, roles=%s",
-        request.url.path,
-        caller_name,
-        roles,
-        extra={"event": "login", "success": True},
-    )
+    if caller_name != "<anonymous user>":
+        LOG.info(
+            "Login: path=%s, user=%s, roles=%s",
+            request.url.path,
+            caller_name,
+            roles,
+            extra={"event": "login", "success": True},
+        )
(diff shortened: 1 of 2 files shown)


negillett:
Health checks don't require auth and have no impact on the service so does it make sense to log those login events at all? Other routes that don't need auth, i.e., task and whoami, may still be interesting to see (when/if we can read IPs) and with this change we wouldn't capture those.

rohanpm (approved):

rohanpm:
> Other routes that don't need auth, i.e., task and whoami, may still be interesting to see

All requests should still be covered by regular access logs though, so removing the 'login' event logs in those cases shouldn't remove traces of those APIs being hit.

crungehottman:
> Health checks don't require auth and have no impact on the service so does it make sense to log those login events at all?

This bit of logging is less intended for usefulness and more intended for consistency in our EOI logging. We have a distinct log event for both authentication (event: login) and authorization (event: auth). We don't have a "login" endpoint, so we needed to agree upon a "login event" definition for the sake of EOI logging. An exodus-gw "login event" can be defined as a request to exodus-gw in which a certificate is accepted by platform-sidecar. A user can include a certificate with a request for any endpoint (including the health check endpoints) and platform-sidecar will verify (or reject) the certificate regardless of whether auth is required to use that endpoint. So, any request in which a certificate is provided is considered a "login event". I shouldn't have categorized an anonymous user's "login" as a "login event" given that a certificate wasn't provided with that request, but at the time I also thought "these might be interesting" so I initially left them in anyway.

PR #570, opened 2023-08-15
Update test-requirements via pip-compile

negillett:
There is a conflict reported by bandit in the current chore: pip-compile
PR, https://github.com/release-engineering/exodus-gw/pull/558. This commit manually updates requirements, adding the
conflicting package, "packaging".


negillett:
@crungehottman @rohanpm @dichn

crungehottman (dismissed):

exodus_gw/routers/publish.py
```diff
@@ -137,7 +139,7 @@ def publish(
 def update_publish_items(
     items: List[schemas.ItemBase] = Body(
         ...,
-        example=[
+        examples=[
```
rohanpm, line 142: This change doesn't seem right. I've assumed the example => examples change is because the library changed from using a single example value, to supporting a list of example values, and that's why you wrapped all the other values in a list.

If you don't also wrap this value in a list, then isn't this changing the meaning of it?  Instead of one example request body, being a list, I think you've changed the meaning into 4 example request bodies, each being a dict.  

The API is designed to take a list, so each example needs to be a list (and it's acceptable to have just one example).

rohanpm (changes requested):

crungehottman (approved):

rohanpm (approved):

dichn (approved):

PR #574, opened 2023-08-23
Retries respect deadlines [RHELDST-18193]

negillett:
Previously, DynamoDB write retries were oblivious to deadlines, which meant that, typically, the retry attempts would have to be exausted before a task could be marked as expired. With exponential backoff on retries, hours of retrying could elapse before an expired task was realized.

This commit sees that DynamoDB clients are aware of and do not exceed deadlines set by callers. This happens by calculating the remaining seconds before the deadline and using that value as the retries' max_time, which is resepcted by backoff's wait generators as well.

### DIFF
--- a/exodus_gw/aws/dynamodb.py
+++ b/exodus_gw/aws/dynamodb.py
@@ -1,5 +1,6 @@
 import json
 import logging
+from datetime import datetime
 from itertools import islice
 from threading import Lock
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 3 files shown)


rohanpm (approved):
While I think this commit makes sense, I don't think it addresses the problem from RHELDST-18193.

There's at least two levels of retry relevant for commit.  There's the one you've updated here, and I think the update makes sense since this retry ought to respect the deadline.

But if you look at the log posted on the issue, it's a different level of retry which was relevant for that incident.  We got "Failed to process message commit" from the dramatiq worker.  This means whatever error happened inside of the task must have been raised and propagated all the way out of the 'commit' actor.  So during those 2 hours delay there, it wasn't this `backoff` here in control of the retry, it was the dramatiq retrying of failed messages/actors.

I'm hitting Approve on this because I do believe it makes sense for *all* levels of retry to take the deadline into account, but the issue would need some other changes as well.

PR #577, opened 2023-08-31
Add logging handler to write to healthcheck file [RHELDST-19614]

negillett:
To track the healthiness of workers in OCP, we'll probe for a specific
file, checking it's latest update time and restarting should a certain
time elapse since that update.

This commit implements the necessary mechanism for writing that file.
By using a custom logging handler, we implement an emit method to log
updates the healthcheck file. The location of the healthcheck file can
be configured via a new setting, "worker_health_filepath".

### DIFF
--- a/exodus_gw/logging.py
+++ b/exodus_gw/logging.py
@@ -9,12 +9,41 @@ def loggers_init(settings):
 
     root_logger = logging.getLogger()
     if not root_logger.hasHandlers():
-        handler = logging.StreamHandler()
-        root_logger.addHandler(handler)
+        root_logger.addHandler(logging.StreamHandler())
+
+    if not any([isinstance(h, GWHandler) for h in root_logger.handlers]):
+        root_logger.addHandler(GWHandler(settings))
 
     for handler in root_logger.handlers:
         datefmt = settings.log_config.get("datefmt")
         handler.setFormatter(JsonFormatter(datefmt))
+
+
+class GWHandler(logging.Handler):  # type: ignore
+    """GWHandler's emit implementation just logs additional information to a
+    file to indicate healthiness of worker(s).
+    """
+
+    def __init__(self, settings):
+        super().__init__()
+        self.settings = settings
+
+    def emit(self, _):
+        """Writes the current datetime to a file specified in settings.
+        Discards the received record.
+        """
+        self.acquire()
+        try:
+            filepath = self.settings.worker_health_filepath
+            with open(filepath, "w") as healthy:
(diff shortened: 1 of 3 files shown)


exodus_gw/logging.py
```diff
@@ -9,14 +9,63 @@ def loggers_init(settings):
 
     root_logger = logging.getLogger()
     if not root_logger.hasHandlers():
-        handler = logging.StreamHandler()
+        handler = GWHandler(settings)
```
rohanpm, line 12: I don't think this should be under the `hasHandlers` condition.

I think the point of this block was to ensure: if there are no handlers at all, then ensure there's at least a basic stream handler set up.  But in practice I don't think we can necessarily predict whether this block is entered, given that:

- It's possible for the `dictConfig` above to have installed some handlers.
- It's possible for third-party code to have installed some handlers before we get here (and there's multiple relevant cases: running under uvicorn, gunicorn, dramatiq or pytest).

I don't think it would be right to not install the special handler in those cases.

I also don't think it was necessary to have GWHandler *extend* a StreamHandler.  It's possible that StreamHandler isn't even the desired type of handler, e.g. uvicorn/gunicorn/dramatiq all accept command-line arguments to log to a file which I believe causes them to install something other than a StreamHandler.

Loggers are designed to support any number of handlers, so we could have left GWHandler dedicated to just updating the healthcheck file while logging to a stream happens via StreamHandler as normal.  Wouldn't that be cleaner?

The final complicating factor is that I think this function can sometimes be called more than once in the process, at least in the "development server with hot reload enabled" case.

Taking that all together I think the logic here ought to be something more like this:

```
# if no handlers at all...
if not root_logger.hasHandlers():
    # then add at least a basic stream handler so logs can be seen
    root_logger.addHandler(logging.StreamHandler())

# if no GWHandler...
if not any([isinstance(h, GWHandler) for h in root_logger.handlers]):
    # then ensure there's one of those too
    root_logger.addHandler(GWHandler(settings))
```

rohanpm (changes requested):

exodus_gw/logging.py
negillett: Hmm. Yeah, I didn't consider missing the StreamHandler block a realistic probability. And maybe it's simpler to keep the handlers separate.

I've simplified the GWHandler, extending logging.Handler instead, and added it to the logger as a separate handler.

rohanpm (approved):

rohanpm:
May as well get this deployed.

PR #583, opened 2023-09-12
Add request/correlation IDs to log messages [RHELDST-18313]

negillett:
Previously, it was unclear which messages were logged for which request. Now, using the new 'request_id' field, one can match log messages that are associated with a particular request.

### DIFF
--- a/exodus_gw/main.py
+++ b/exodus_gw/main.py
@@ -58,8 +58,11 @@ If you are a client looking to make use of exodus-gw, consult your organization'
 internal documentation for advice on which environment(s) you should be using.
 """
 
+from uuid import uuid4
+
 import backoff
 import dramatiq
+from asgi_correlation_id import CorrelationIdMiddleware, correlation_id
 from fastapi import Depends, FastAPI, Request
 from fastapi.exception_handlers import http_exception_handler
 from fastapi.exceptions import RequestValidationError
@@ -97,6 +100,20 @@ app.include_router(upload.router)
 app.include_router(publish.router)
 app.include_router(deploy.router)
 app.include_router(cdn.router)
+
+app.add_middleware(CorrelationIdMiddleware, generator=lambda: uuid4().hex[:8])
+
+
+@app.exception_handler(Exception)
+async def unhandled_exception_handler(request, exc):
+    return await http_exception_handler(
+        request,
+        StarletteHTTPException(
+            500,
+            "Internal server error",
+            headers={"X-Request-ID": correlation_id.get() or ""},
+        ),
+    )
 
 
 @app.exception_handler(RequestValidationError)
--- a/tests/routers/test_common.py
+++ b/tests/routers/test_common.py
@@ -85,19 +85,20 @@ def test_login_log(endpoint, user, roles, caplog, auth_header):
     with TestClient(app) as client:
         if roles:
             if endpoint == "/foo/publish":
-                client.post(
+                r = client.post(
                     endpoint, headers=auth_header(roles=["test-publisher"])
                 )
             else:
-                client.get(
+                r = client.get(
                     endpoint, headers=auth_header(roles=["test-publisher"])
                 )
         else:
-            client.get(endpoint)
+            r = client.get(endpoint)
         expected_log = {
             "level": "INFO",
             "logger": "exodus-gw",
             "time": "2023-07-28 13:24:03.596",
+            "request_id": r.headers["X-Request-ID"],
             "message": f"Login: path={endpoint}, user={user}, roles={roles}",
             "event": "login",
             "success": True,
--- a/tests/routers/upload/test_manage_mpu.py
+++ b/tests/routers/upload/test_manage_mpu.py
@@ -4,7 +4,7 @@ import mock
 from fastapi.testclient import TestClient
 
 from exodus_gw.aws.util import xml_response
-from exodus_gw.deps import get_environment, get_s3_client, get_settings
+from exodus_gw.deps import get_environment, get_s3_client
 from exodus_gw.main import app
 from exodus_gw.routers.upload import multipart_upload
 from exodus_gw.settings import load_settings
@@ -140,6 +140,7 @@ async def test_bad_mpu_call(auth_header):
         b"<Endpoint>/upload/test/b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c</Endpoint>"
         b"</Error>"
     )
+    assert r.headers["X-Request-ID"]
 
 
 async def test_abort_mpu(mock_aws_client, auth_header):
--- a/exodus_gw/logging.py
+++ b/exodus_gw/logging.py
@@ -3,6 +3,8 @@ import json
 import logging
 import logging.config
 
+from asgi_correlation_id import CorrelationIdFilter
+
 
 def loggers_init(settings):
     logging.config.dictConfig(settings.log_config)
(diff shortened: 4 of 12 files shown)


tests/routers/test_common.py
```diff
@@ -85,19 +85,20 @@ def test_login_log(endpoint, user, roles, caplog, auth_header):
...
                     endpoint, headers=auth_header(roles=["test-publisher"])
                 )
         else:
-            client.get(endpoint)
+            r = client.get(endpoint)
         expected_log = {
             "level": "INFO",
             "logger": "exodus-gw",
             "time": "2023-07-28 13:24:03.596",
+            "request_id": r.headers["X-Request-ID"][:8],
```
rohanpm, line 101: I'm curious about the :8 , does the X-Request-ID end up with some kind of prefix which doesn't show up in the request_id in the log?

exodus_gw/main.py
```diff
@@ -98,6 +99,8 @@
 app.include_router(deploy.router)
 app.include_router(cdn.router)
 
+app.add_middleware(CorrelationIdMiddleware)
+
```
rohanpm, line 104: What's here looks good, but I would ask can we please also add an exception_handler here to process the 500 Internal Server Error case.

I'm looking at https://github.com/snok/asgi-correlation-id#exception-handling and it's mentioned that the X-Request-ID by default will not be added on 500 responses.  That seems like an important case & I'm a bit surprised that's the default.  Cases where our server is crashing might be where the request ID is most valuable.

There is a FastAPI example there and I *think* just copy-pasting that exactly as it is would work.  It should play nicely with the other `app.exception_handler` already defined here.

rohanpm (approved):

tests/routers/test_common.py
negillett: The "X-Request-ID" header has a full 32 char ID but only the first 8 are logged as "request_id", per the filter settings. I felt the full ID would clutter the logs unnecessarily--8 chars is enough to match messages, I think.

exodus_gw/main.py
negillett: Actually, from a new test, it looks like HTTP errors of code 500 include the correlation ID just fine so long as it originates from the app. The only cases in which X-Request-ID is null is when the app isn't really in use, such as logs coming from DB.

exodus_gw/main.py
```diff
@@ -116,7 +123,11 @@ async def custom_http_exception_handler(request, exc):
 
     if path.startswith("/upload"):
         return xml_response(
-            "Error", Code=exc.status_code, Message=exc.detail, Endpoint=path
+            "Error",
+            Code=exc.status_code,
+            Message=exc.detail,
+            Endpoint=path,
+            Headers=request.headers,
```
rohanpm, line 130: Same point here that I don't think it's right to copy the request headers into the response.

exodus_gw/main.py
rohanpm: I had the same thought about the handlers, but I did investigate it a bit before making the suggestion and found there is special handling which should make handling of `Exception` work as desired.

It ultimately boils down to this code: https://github.com/encode/starlette/blob/93074e82ab15e860456d4a7074c0494eed78ea0c/starlette/applications.py#L84

You can see that an exception handler for `Exception` or for 500 status is picked out separately from all the rest, and then ends up installed at "the very outermost middleware" (as a comment in that file calls it).

So if we add an `Exception` handler exactly as the example, that should play nicely with all the other more specific handlers.  It does not override them, instead it's used as a last resort.

This is why I really do expect that copy-pasting the asgi-correlation-id README example ought to work and is the right thing to do rather than updating the exception handlers we've already added.  Those aren't designed to catch all types of exceptions anyway.

Not 100% sure but I'm guessing that "DB is offline" is probably a case where none of those existing exception handlers get invoked and X-Request-ID would currently be missing.

tests/routers/test_common.py
rohanpm: I see.  I agree 32 chars is too much.

I'm wondering if the right way is to instead override `generator` when creating the CorrelationIdMiddleware.  Docs say it defaults to `generator=lambda: uuid.uuid4().hex`, I guess changing it to `generator=uuid.uuid4().hex[:8]` would make the value as 8 chars all the way through both the logging & request headers.

That is probably slightly nicer - I'm thinking, if someone does in fact report a failure and their failed response has a 32-char request ID, if anyone tries to copy-paste that directly into Splunk they're going to find nothing.  Are they really going to remember that they need to chop only 8 chars?  Will they remember whether that's at the beginning or the end?  Will they do it correctly or will they accidentally get the wrong number of chars and fail to find any matching messages?

rohanpm (changes requested):

exodus_gw/main.py
negillett: Ah, okay. I totally missed the method name (for FastAPI) as well which hints at its uniqueness.
Thanks!

tests/routers/test_common.py
negillett: This actually works really well. I wish they advertised this solution rather than suggesting truncating via filter, creating this inconsistency between the log and header. Or move truncation down to the middleware level, adding the uuid_length option to that class. But I digress. Thanks for the suggestion.

exodus_gw/main.py
```diff
@@ -98,13 +99,19 @@
...
 @app.exception_handler(RequestValidationError)
 async def validation_exception_handler(request, exc):
     # When validating requests, multiple violations are possible.
     # Let's return them all to expedite troubleshooting.
     msgs = [e["msg"] for e in exc.errors()]
-    return JSONResponse(status_code=400, content={"detail": msgs})
+    return JSONResponse(
+        status_code=400,
+        content={"detail": msgs},
+        headers=request.headers,
```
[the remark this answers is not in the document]
negillett: Yeah, you're right. Request headers probably aren't good here and, through more testing, the header is already included in JSONReponse. This is not the case for our custom xml_response, however.

exodus_gw/main.py
negillett: I've removed the request headers but included the X-Request-ID header, as it's not automatically applied like the others (presumably because it's our own function).

negillett:
@rohanpm @crungehottman @dichn, ready for another look

crungehottman (dismissed):

tests/routers/upload/test_manage_mpu.py
```diff
@@ -132,14 +132,12 @@ async def test_bad_mpu_call(auth_header):
...
-    assert r.content == (
-        b"<?xml version='1.0' encoding='UTF-8'?>\n"
-        b"<Error>"
+    assert (
         b"<Code>400</Code>"
         b"<Message>Invalid uploadId='my-upload', uploads=''</Message>"
         b"<Endpoint>/upload/test/b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c</Endpoint>"
-        b"</Error>"
-    )
+        b"<Headers>{'X-Request-ID':"
```
rohanpm, line 139: This doesn't seem correct, on two levels:

- rather than adding the X-Request-ID as an HTTP response header it's added it into a new `<Headers>` section of the response body.
- rather than representing the X-Request-ID as a key/value in the way you'd expect for either XML or JSON, this seems to be just the default python stringification of a dict.

Given that these responses are meant to be used on the APIs which aim to be S3-compatible, we have to be very careful about inserting new parts into the response which are not in real S3, because existing clients like boto might not expect those and might start crashing or doing something strange.

I think what you want to do is, instead of:

```python
return xml_response(..., Headers={'X-Request-ID': ...})
```

Update the response object after creating it:

```python
response = xml_response(...)
response.headers['X-Request-ID'] = ...
return response
```

... because the xml_response function is designed to put all of its kwargs into the XML document.  It doesn't have any way to adjust the headers.

I would suggest also double-checking if the custom_http_exception_handler really needed *any* changes at all, because what *ought* to be happening is that the installed CorrelationIdMiddleware already gets a chance to rewrite the value returned from that function anyway.  So it's unclear why the handler itself would also need to do anything.

rohanpm (changes requested):

tests/routers/upload/test_manage_mpu.py
negillett: Looks like the middleware is adding the header to the xml_response as well.
Also, while the new unhandled_exception_handler is covered by a test (test_db_rollback_on_raise_db), removing it entirely doesn't cause any failures. So I guess this new handler is really just an insurance policy.

rohanpm (approved):

PR #585, opened 2023-09-18
Ensure openapi files are copied to docs [RHELDST-19887]

negillett:
Previously we relied on Sphinx to overwrite build files with those listed in html_extra_files, specifically api.html. However, for some reason, that is no longer happening. This commit adds a copy command to the gen-openapi script to ensure that the appropriate files are present in the final build.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -65,7 +65,7 @@ master_doc = "index"
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -104,14 +104,14 @@
...
-html_static_path = ["_static"]
+# html_static_path = ["_static"]
 
 # This dir contains the generated openapi spec and the static redoc-based
 # api.html for viewing. Copy it alongside the other docs.
 # Note: to enable sphinx linking to api.html, we also have an api.rst
 # and it's expected to be overwritten here. We depend on the implementation
 # detail that html_extra_path copying happens after .rst rendering.
-html_extra_path = ["openapi"]
+# html_extra_path = ["openapi"]
```
rohanpm, line 114: I don't think anyone is going to come back and undo this change later, so can you please clean up the comment above as well?

It doesn't make sense now.  The code is like:

- an explanation of why we're about to set certain config
- and then... no config is set

There is no hint to the reader that what the comment says is actually NOT how it works now and I can see someone reading that comment and coming away with a wrong understanding of how things work.

docs/conf.py
```diff
@@ -104,14 +104,14 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ["_static"]
+# html_static_path = ["_static"]
 
 # This dir contains the generated openapi spec and the static redoc-based
 # api.html for viewing. Copy it alongside the other docs.
 # Note: to enable sphinx linking to api.html, we also have an api.rst
 # and it's expected to be overwritten here. We depend on the implementation
```
rohanpm, line 112: I'm not sure but this comment (about depending on the implementation detail that html_extra_path copying happens after .rst rendering) might be pointing at the reason it broke.  I don't think it was guaranteed to work that way, so maybe it no longer does.

rohanpm (changes requested):

rohanpm (approved):

PR #587, opened 2023-09-19
Move correlation_id middleware and add validator [RHELDST-18313]

negillett:
This commit moves the addition of CorrelationIdMiddleware to the end of main.py so that it's invoked first and the ID is added across the entire application. Additionally a custom correlation ID validator was added to prevent validation errors when the middleware encounters its own IDs, e.g., retries.

### DIFF
--- a/exodus_gw/main.py
+++ b/exodus_gw/main.py
@@ -58,6 +58,7 @@ If you are a client looking to make use of exodus-gw, consult your organization'
 internal documentation for advice on which environment(s) you should be using.
 """
 
+import re
 from uuid import uuid4
 
 import backoff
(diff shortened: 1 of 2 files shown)


exodus_gw/main.py
```diff
@@ -219,3 +218,14 @@ async def db_session_wrap():
         return response
 
     return db_session_wrap()
+
+
+def request_id_validator(short_uuid: str):
+    return re.match(r"[0-9a-f]{8}", short_uuid)
```
rohanpm, line 224: You want this expression to be anchored, because otherwise it matches even if there's garbage on the end of the string.

e.g.

```
>>> import re

# bad
>>> re.match(r"[0-9a-f]{8}", "a1b2c3d4e5 some other stuff")
<re.Match object; span=(0, 8), match='a1b2c3d4'>

# good (should not match)
>>> re.match(r"^[0-9a-f]{8}$", "a1b2c3d4e5 some other stuff")
```

(Actually 'match' is already anchored at the beginning, so only the $ to anchor at the end is needed, but to me it always feels cleaner and easier to understand if both are included.)

rohanpm (changes requested):

rohanpm (approved):

PR #588, opened 2023-09-21
Retry upload on UnseekableStreamError [RHELDST-20355]

negillett:
We're still seeing intermittent UnseekableStreamErrors on uploads when boto attempts to retry because we stream the body directly from the client without buffering. Rather than loading the content to memory, let's just retry the put operation in those scenarios.

### DIFF
--- a/tests/routers/upload/test_upload.py
+++ b/tests/routers/upload/test_upload.py
@@ -1,7 +1,6 @@
-from uuid import UUID
-
 import mock
 import pytest
+from botocore.exceptions import UnseekableStreamError
 from fastapi.testclient import TestClient
 
 from exodus_gw.main import app
@@ -12,7 +11,9 @@ TEST_KEY = "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
 async def test_full_upload(
     mock_aws_client, mock_request_reader, auth_header, monkeypatch
 ):
-    """Uploading a complete object is delegated correctly to S3."""
+    """Uploading a complete object is delegated correctly to S3 and can tolerate
+    intermittent UnseekableStreamErrors.
+    """
 
     monkeypatch.setenv(
         "EXODUS_GW_UPLOAD_META_FIELDS",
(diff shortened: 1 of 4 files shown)


crungehottman:
I think we'd also need to retry multipart uploads

tests/routers/upload/test_upload.py
```diff
@@ -26,7 +27,11 @@ async def test_full_upload(
     }
 
     mock_request_reader.return_value = b"some bytes"
-    mock_aws_client.put_object.return_value = {"ETag": "a1b2c3"}
+    mock_aws_client.put_object.side_effect = [
```
rohanpm, line 30: I believe this side effect is impossible:

When the request body is unseekable, there's no way that retrying can somehow make it become seekable.  The request body is already "used up" and no amount of server-side retries could help.

rohanpm (changes requested):
I don't think this change makes sense, the issue can't be solved by retrying in the server, unless we also make some larger refactors on how requests are read.

To put it another way, these functions are not retry-safe because they have the side-effect of changing the current position of the incoming request's stream.

So, the request's body is accessed as a file-like object, and when we do s3.put_object bytes will be read from that object.  If we fail part way through, the bytes already read are not "put back".  e.g. if we have already read 5MB, on the next retry if we start reading again it's going to start from 5MB into the object, so we'd end up calling s3.put_object with some of the bytes missing.

How can that be solved?  Well, you can explicitly ask the stream to go back to the beginning by `.seek(0)`.  But that's the problem: s3.put_object *already* tries to do that internally, but it cannot because the stream isn't seekable, hence the UnseekableStreamError.

Adding another layer of retry on top isn't going to help us here, the retries should also all fail with UnseekableStreamError (best case), or even corrupt the upload if the code is not smart enough to realize it's been passed a stream which is no longer pointing at the beginning of the content.

rohanpm:
> I don't think this change makes sense, the issue can't be solved by retrying in the server, unless we also make some larger refactors on how requests are read.
> 

I was able to confirm for sure that this doesn't work.  I figured out a kinda-realistic way to trigger the UnseekableStreamError on demand locally, in https://gist.github.com/rohanpm/5d7bc4cefb772d8cfb34597b336e028c .

When I use that patch from the gist, *without* this change, I get UnseekableStreamError.

When I add this change on top, I get backoff saying it'll retry due to the UnseekableStreamError but it eventually fails regardless with "Stream consumed":

```
Sep 22 15:10:07 photinia.usersys.redhat.com sh[45951]: {"level": "DEBUG", "logger": "s3", "time": "2023-09-22 05:10:07.547", "request_id": "c0773dbf", "message": "Backing off object_put(...) for 0.8s (botocore.exceptions.UnseekableStreamError: Need to rewind the stream <exodus_gw.aws.util.RequestReader object at 0x7f5930c7fe80>, but stream is not seekable.)", "event": null, "success": null}
[...]
Sep 22 15:10:10 photinia.usersys.redhat.com sh[45951]: botocore.exceptions.HTTPClientError: An HTTP Client raised an unhandled exception: Stream consumed
```

"Stream consumed" being another way of what I'm saying above - the stream has already been read from, the bytes can't be "put back" afterward, so a retry doesn't work.

So yeah, unfortunately I'm quite confident that this change doesn't do anything helpful.

PR #591, opened 2023-09-26
Update S3ClientWrapper to prevent retries [RHELDST-20355]

negillett:
It was discovered that boto client configuration supports two retry attempt settings. The one used previously is not inclusive of the initial attempt so `"max_attempts": 1` means one retry will be attempted. This commit changes this setting to "total_max_attempts" because it is 1) preferred by the library and 2) inclusive of the initial attempt.

This directly addresses the intermittent UnseekableStreamErrors seen recently. These errors are only raised when boto attempts to retry S3 uploads, which we were unknowlingly commanding it to do.

Additionally in this commit, a key-value pair is added to the client's request to prevent redirects if/when botocore's newest redirector class is used.

### DIFF
--- a/exodus_gw/aws/client.py
+++ b/exodus_gw/aws/client.py
@@ -26,7 +26,7 @@ class S3ClientWrapper:
             # We don't allow any retries - it's not possible since we're streaming
             # request bodies directly to S3, we don't buffer it anywhere, so we
             # can't send it more than once.
-            config=Config(retries={"max_attempts": 1}),
+            config=Config(retries={"total_max_attempts": 1}),
         )
 
     async def __aenter__(self):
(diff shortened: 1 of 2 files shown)


exodus_gw/aws/client.py
```diff
@@ -46,7 +46,7 @@ async def __aexit__(self, exc_type, exc, tb):
 
     @staticmethod
     def no_redirects(**kwargs):
```
rohanpm, line 48: This might all even be dead code now, maybe it only wasn't dead before due to the issue with max_attempts=1 meaning "retry once", which is not what I thought when writing that code.

After this fix, it looks like we've got config to disable retries altogether, but then also this no_redirects code to disable retries in another specific case, which seems like it should be redundant.

If there is any cleanup to be done, I'd be in favor of pursuing that separately though, as it definitely makes sense to fix the max_attempts/total_max_attempts issue regardless.

rohanpm (approved):

exodus_gw/aws/client.py
negillett: Yeah, maybe. I'd need another look at the botocore code to be sure this case doesn't bypass the typical retry path. Overall in favor of potentially cleaning it up in the future though.

PR #596, opened 2023-10-02
Introduce concept of phase1 commit [RHELDST-20490]

rohanpm:
Internally, items have always been categorized into either phase1 or phase2 during commit, with phase2 being committed last. This is needed to ensure, for example, that a repomd.xml is not committed until after all other files referenced by it have been committed.

This change now documents that concept and exposes it via the API. By default, clients get the same behavior as always, but they can now also explicitly request a phase1 commit.

A phase1 commit writes phase1 items to the DB, same as always, but then stops without proceeding to write phase2 items. It also leaves the publish open for later modifications and a later phase2 commit.

## Why?

This will be used to solve the following Pub/rhsm-pulp/exodus-gw integration problem:

Imagine that you need to publish multiple Pulp yum repos. You want this to be atomic, so you want them all to use the same exodus-gw publish.

You ask the repos to publish, and they succeed, but then something goes wrong during the later exodus-gw commit which is performed outside of Pulp.

So, you restart the whole process and republish the repos again using a new exodus-gw publish, and commit that successfully.

Problem: as far as Pulp is concerned, the publish tasks from the first attempt were completely successful, as it does not "see" the later failure to commit. Therefore, Pulp incorrectly thinks that the RPMs processed by those tasks are fully published to the CDN, and so skips publishing them during later tasks. This leads to missing content on the CDN.

Solution: exodus-rsync, as invoked by Pulp during publish, will request a phase1 commit during publish of each repo. This ensures the processed RPMs (or other non-entrypoint files) are fully published on the CDN by the time the Pulp publish task succeeds, matching Pulp's expectations. The publish of repomd.xml is still held back until a later phase2 commit, retaining the atomic semantics across multiple repos.

### DIFF
--- a/exodus_gw/worker/publish.py
+++ b/exodus_gw/worker/publish.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from os.path import basename
 from queue import Empty, Full, Queue
 from threading import Thread
-from typing import Any, List
+from typing import Any, List, Optional
 
 import dramatiq
 from dramatiq.middleware import CurrentMessage
@@ -14,7 +14,7 @@ from sqlalchemy.orm import Session, lazyload
 
 from exodus_gw.aws.dynamodb import DynamoDB
 from exodus_gw.database import db_engine
-from exodus_gw.models import CommitTask, Item, Publish
+from exodus_gw.models import CommitModes, CommitTask, Item, Publish
 from exodus_gw.schemas import PublishStates, TaskStates
 from exodus_gw.settings import Settings, get_environment
 
@@ -145,7 +145,11 @@ class _BatchWriter:
                 break
 
 
-class Commit:
+class CommitBase:
+    # Acceptable publish states for this type of commit.
+    # Subclasses need to override.
+    PUBLISH_STATES: list[PublishStates] = []
+
     def __init__(
         self,
         publish_id: str,
@@ -156,7 +160,7 @@ class Commit:
     ):
         self.env = env
         self.from_date = from_date
-        self.rollback_item_ids: List[str] = []
+        self.written_item_ids: List[str] = []
         self.settings = settings
         self.db = Session(bind=db_engine(self.settings))
         self.task = self._query_task(actor_msg_id)
@@ -195,16 +199,14 @@ class Commit:
                 task.deadline,
                 extra={"event": "publish", "success": False},
             )
-            # Fail expired task and associated publish
-            self.task.state = TaskStates.failed
-            self.publish.state = PublishStates.failed
+            self.on_failed()
             self.db.commit()
             return False
         return True
 
     @property
     def publish_ready(self) -> bool:
-        if self.publish.state == PublishStates.committing:
+        if self.publish.state in self.PUBLISH_STATES:
             return True
         LOG.warning(
             "Publish %s in unexpected state, '%s'",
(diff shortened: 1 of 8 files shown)


exodus_gw/worker/publish.py
```diff
@@ -342,68 +433,61 @@ def write_publish_items(self) -> None:
...
+    settings: Settings = Settings(),
 ) -> None:
     actor_msg_id = CurrentMessage.get_current_message().message_id
-    commit_obj = Commit(publish_id, env, from_date, actor_msg_id, settings)
+
+    commit_mode = commit_mode or CommitModes.phase2.value
+    commit_class: type[CommitBase] = {
+        CommitModes.phase1.value: CommitPhase1,
+        CommitModes.phase2.value: CommitPhase2,
+    }[commit_mode]
```
negillett, line 473: Do we need to create both phase classes? Seems like only the one corresponding to `commit_mode` would ever be used. Maybe this is just an alternative to a conditional? E.g., `commit_class = CommitPhase1 if "1" in commit_mode else CommitPhase2`.

negillett (approved):

exodus_gw/worker/publish.py
```diff
@@ -323,12 +339,87 @@ def write_publish_items(self) -> None:
...
+
+    def pre_write(self):
+        # Any steps prior to DynamoDB write.
+        # Base implementation does nothing.
+        pass
+
+
+class CommitPhase1(CommitBase):
+    # phase1 commit is allowed to proceed in either of these states.
+    PUBLISH_STATES = [PublishStates.committing, PublishStates.pending]
```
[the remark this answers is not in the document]
rohanpm: At the time of *enqueuing* a phase1 commit it's guaranteed that the publish state will be PENDING, but at the time of starting to work on it, the state may have changed to COMMITTING.

I have this sequence of events in mind:

1. client submits phase1 commit, publish was and remains PENDING
2. client submits phase2 commit, publish was PENDING but is now updated to COMMITTING
3. worker starts to work on phase1 commit.

The question is, what should happen at step (3)?  Currently my answer is that the phase1 commit should be allowed to proceed normally even if a phase2 commit has already been enqueued.  It seems like that'll work just fine, so it didn't seem right to unnecessarily introduce a crash or failure in the above scenario.  Therefore it should be allowed to proceed even in COMMITTING state.

exodus_gw/worker/publish.py
rohanpm: I'm not quite sure what you mean by "create both phase classes".  The constructor is only invoked on whichever one we've decided to use (e.g. the code in these lines has `CommitPhase1` and not `CommitPhase1(...)`).

You're right about the overall point of this code though.  This is basically another way of writing:

```
if commit_mode == CommitModes.phase1.value:
  commit_class = CommitPhase1
elif commit_mode == CommitModes.phase2.value:
  commit_class = CommitPhase2
else:
  # raise some kind of error about the unexpected value
```

The problem I have with the if/else example above is that the final "else" path will need to have a test written for it or be marked as 'pragma no coverage', and the problem I have with the `commit_class = CommitPhase1 if "1" in commit_mode else CommitPhase2` example is that it makes it possible for `commit_mode` to contain random garbage without that being detected, since it treats all unknown values as valid.

The way I wrote it will detect/crash immediately if the code is passed an unexpected value, while at the same time maintain 100% statement coverage without requiring a test to explicitly cover the unexpected value case.

rohanpm:
I'll get the exodus-rsync side of this up for review as well, so they can be reviewed together before submitting this if desired.

exodus_gw/worker/publish.py
negillett: Ah, I see now. They're not constructed at that point. Yeah, that makes sense. Thanks.

rohanpm:
The different parts of this are up for review in case anyone wants to look at them all together.

- https://github.com/release-engineering/exodus-gw/pull/596
- https://github.com/release-engineering/exodus-rsync/pull/364
- https://github.com/release-engineering/pubtools-exodus/pull/28

exodus_gw/worker/publish.py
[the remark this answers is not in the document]
crungehottman: Okay, that makes sense. Thanks!

crungehottman (approved):

crungehottman:
Another question related to the other two commits you submitted:

> exodus-rsync, as invoked by Pulp during publish, will request a phase1 commit during publish of each repo. This ensures the processed RPMs (or other non-entrypoint files) are fully published on the CDN by the time the Pulp publish task succeeds, matching Pulp's expectations. The publish of repomd.xml is still held back until a later phase2 commit, retaining the atomic semantics across multiple repos

Where/by what means do the phase 2 commits occur?

rohanpm:
> Another question related to the other two parts you submitted for review:
> 
> > exodus-rsync, as invoked by Pulp during publish, will request a phase1 commit during publish of each repo. This ensures the processed RPMs (or other non-entrypoint files) are fully published on the CDN by the time the Pulp publish task succeeds, matching Pulp's expectations. The publish of repomd.xml is still held back until a later phase2 commit, retaining the atomic semantics across multiple repos
> 
> Where/by what means do the phase 2 commits occur?

This part hasn't changed, so it's at the same place it always was, which is: https://github.com/release-engineering/pubtools-exodus/blob/032079efd17029957e408688b35971462df021bd/pubtools/exodus/_hooks/pulp.py#L80

Note, the code has not been updated to explicitly add "phase2" option since that's the default anyway, and specifying it just makes the code incompatible with current versions of exodus-gw.

rohanpm:
> > 
> > Where/by what means do the phase 2 commits occur?
> 
> This part hasn't changed, so it's at the same place it always was

...and to expand on that a little: since pubtools-exodus is both the place which asks rhsm-pulp to do a phase1 commit, and also the place which manages the final phase2 commit, this is meant to mitigate the risk that some code might ask for a phase1 commit and forget to do the later phase2 commit.  As the responsibility for triggering both of them lies in the same place.

PR #598, opened 2023-10-04
Fix autoindex step unable to retry S3 requests [RHELDST-20662]

rohanpm:
There is an S3ClientWrapper utility class which provides some default config to our S3 clients. One of the things it does is disable retries, for the reasons explained in RHELDST-20355.

However: those reasons only apply in the /upload API case where we are taking an HTTP request from a client and forwarding it to S3. It does NOT apply in the worker case, where the worker is acting as a regular S3 client with no special requirements or restrictions. Therefore it was inappropriate to use the config which disables retries here.

Change it to just construct an S3 client in the normal way, ignoring the wrapper. The one piece of logic we still need from the wrapper (respect the endpoint URL environment variable) is reproduced here.

This problem has been present for a long time, but prior to RHELDST-20355 there was an off-by-one error where the client allowed a single retry even though that was not intended. With that fixed, there were truly no retries allowed here. According to logs, some errors have always been happening here but the RHELDST-20355 fix increased the failure rate from 10% to 20%.

### DIFF
--- a/exodus_gw/worker/autoindex.py
+++ b/exodus_gw/worker/autoindex.py
@@ -1,15 +1,16 @@
 import gzip
 import hashlib
 import logging
+import os
 import tempfile
 from time import monotonic
 from typing import AsyncGenerator, BinaryIO, Generator, Optional
 
+import aioboto3
 from repo_autoindex import ContentError, Fetcher, autoindex
 from sqlalchemy import inspect
 from sqlalchemy.orm import Session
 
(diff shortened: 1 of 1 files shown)


rohanpm:
This is covered by existing tests, in that it is verified that the autoindex step does the expected S3 uploads.  I didn't think it was warranted to write new tests to specifically verify that it retries on errors, since all we're doing here is asking the code to use a default-configured S3 client.

crungehottman (approved):

PR #606, opened 2023-10-13
Include content & origin signatures in redirect request [RHELDST-18071]

negillett:
Rather than depending on exodus-cdn to generate its own signatures to
complete the redirection, generate the signatures here and forward them
with the redirect request. This should improve runtime and allows us to
remove code in exodus-cdn (lambda) that's present here.

### DIFF
--- a/exodus-gw.ini
+++ b/exodus-gw.ini
@@ -13,7 +13,7 @@ bucket = my-bucket2
 table = my-table2
 config_table = my-config2
 
-cdn_url = https://test2.cloudfront.net
+cdn_url = https://test2.cloudfront.net/_/cookie
 cdn_key_id = XXXXXXXXXXXXXX
 
 [env.test3]
@@ -22,7 +22,7 @@ bucket = my-bucket3
 table = my-table3
 config_table = my-config3
 
-cdn_url = https://test3.cloudfront.net
+cdn_url = https://test3.cloudfront.net/_/cookie
 cdn_key_id = XXXXXXXXXXXXXX
 
 [loglevels]
--- a/exodus_gw/routers/cdn.py
+++ b/exodus_gw/routers/cdn.py
@@ -4,9 +4,9 @@ import base64
 import json
 import logging
 import os
-import urllib.parse
 from collections import OrderedDict
 from datetime import datetime, timedelta, timezone
+from urllib.parse import urlparse
 
 from botocore.utils import datetime2timestamp
 from cryptography.hazmat.backends import default_backend
@@ -52,7 +52,30 @@ def cf_b64(data: bytes):
     )
 
 
-def sign_url(url: str, timeout: int, env: Environment):
+def cf_cookie(url: str, env: Environment, expires: datetime, username: str):
+    policy = build_policy(url, expires)
+    signature = rsa_signer(env.cdn_private_key, policy)
+    policy_encoded = cf_b64(policy).decode("utf-8")
+    signature_encoded = cf_b64(signature).decode("utf-8")
+
+    LOG.info(
+        "Generated cookie for: user=%s, key=%s, resource=%s, expires=%s, policy=%s",
+        username,
+        env.cdn_key_id,
+        url,
+        expires,
+        policy_encoded,
+        extra={"event": "cdn", "success": True},
+    )
+
+    return {
+        "CloudFront-Key-Pair-Id": env.cdn_key_id,
+        "CloudFront-Policy": policy_encoded,
+        "CloudFront-Signature": signature_encoded,
+    }
+
+
+def sign_url(url: str, timeout: int, env: Environment, username: str):
     if not env.cdn_url:
         LOG.error(
             "Missing cdn_url in exodus-gw environment settings",
@@ -81,7 +104,7 @@ def sign_url(url: str, timeout: int, env: Environment):
         )
 
     dest_url = os.path.join(env.cdn_url, url)
-    expiration = datetime.now(timezone.utc) + timedelta(seconds=timeout)
+    expires = datetime.now(timezone.utc) + timedelta(seconds=timeout)
 
     LOG.info(
         "redirecting %s to %s. . .",
@@ -90,11 +113,26 @@ def sign_url(url: str, timeout: int, env: Environment):
         extra={"event": "cdn", "success": True},
     )
 
-    policy = build_policy(dest_url, expiration)
+    policy = build_policy(dest_url, expires)
     signature = rsa_signer(env.cdn_private_key, policy)
+
+    cookies = []
+    for resource in ("/content/*", "/origin/*"):
+        parsed_url = urlparse(env.cdn_url)
+        base_url = f"{parsed_url.scheme}://{parsed_url.netloc}"
+        policy_url = f"{base_url}{resource}"
+        cookie = cf_cookie(policy_url, env, expires, username)
+        append = (
+            f"; Secure; HttpOnly; SameSite=lax; Domain={base_url}; "
+            f"Path={resource}; Max-Age={timeout}"
+        )
+        cookies.extend([f"{k}={v}{append}" for k, v in cookie.items()])
+    cookies_bytes = bytes(json.dumps(cookies), "utf-8")
+
     params = [
-        "Expires=%s" % int(datetime2timestamp(expiration)),
+        "Expires=%s" % int(datetime2timestamp(expires)),
         "Signature=%s" % cf_b64(signature).decode("utf8"),
+        "Set-Cookies=%s" % cf_b64(cookies_bytes).decode("utf-8"),
         "Key-Pair-Id=%s" % env.cdn_key_id,
     ]
     separator = "&" if "?" in url else "?"
@@ -140,6 +178,7 @@ def cdn_redirect(
     url: str = Url,
     settings: Settings = deps.settings,
     env: Environment = deps.env,
+    call_context: auth.CallContext = deps.call_context,
 ):
     """Redirects to a requested URL on the CDN.
 
@@ -151,7 +190,12 @@ def cdn_redirect(
     The URL used in the redirect will become invalid after a server-defined
     timeout, typically less than one hour.
     """
-    signed_url = sign_url(url, settings.cdn_signature_timeout, env)
+    username = (
+        call_context.client.serviceAccountId
+        or call_context.user.internalUsername
+        or "<unknown user>"
+    )
+    signed_url = sign_url(url, settings.cdn_signature_timeout, env, username)
     return Response(
         content=None, headers={"location": signed_url}, status_code=302
     )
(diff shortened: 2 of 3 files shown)


exodus_gw/routers/cdn.py
```diff
@@ -94,6 +94,7 @@ def sign_url(url: str, timeout: int, env: Environment):
     signature = rsa_signer(env.cdn_private_key, policy)
     params = [
         "Expires=%s" % int(datetime2timestamp(expiration)),
+        "Policy=%s" % cf_b64(policy).decode("utf8"),
```
rohanpm, line 97: This doesn't seem like it's going to be useful.  The policy and signature here only enable access to the /_/cookie endpoint, so it can't be reused in a cookie for accessing any other paths.

You can think of it this way: we're shifting responsibility for generating the policy/signature which lives in the cookie from `/_/cookie` endpoint to here.  Therefore the policies/signatures generated here will have to include at least whatever `/_/cookie` was previously generating.

I'm pretty sure there will need to be three policies in total:
1. one for `/content/*`
2. one for `/origin/*`
3. one for `dest_url` (which starts with `/_/cookie/`)

The first two policies need to be somehow wrapped up for the request for `/_/cookie/`, e.g. encoded into query parameters.  But that should NOT be using the standard `Policy`, `Signature` since the point is to get the policy into our own exodus-cdn code, and not processed by CloudFront.

rohanpm (changes requested):

exodus_gw/routers/cdn.py
negillett: Ah, I see. Thanks for clarifying!

negillett:
@rohanpm @crungehottman updated to forward origin and content signatures

crungehottman (approved):

exodus_gw/routers/cdn.py
```diff
@@ -90,11 +90,24 @@ def sign_url(url: str, timeout: int, env: Environment):
         extra={"event": "cdn", "success": True},
     )
 
-    policy = build_policy(dest_url, expiration)
-    signature = rsa_signer(env.cdn_private_key, policy)
+    dest_policy = build_policy(dest_url, expiration)
+    dest_signature = rsa_signer(env.cdn_private_key, dest_policy)
+
+    content_url = os.path.join(env.cdn_url, "/content/*")
```
rohanpm, line 96: I don't think these os.path.join make sense because they're just going to return whatever the second argument was, since it's an absolute path. e.g.

```
>>> os.path.join('https://abc.cloudfront.net/foo/bar', '/content/*')
'/content/*'
```

I believe this is a rare case where you might actually want to use urljoin:

```
# this is exactly what we want, the "/" brings us up to the root of the URL's domain and no further
>>> from urllib.parse import urljoin
>>> urljoin('https://abc.cloudfront.net/foo/bar', '/content/*')
'https://abc.cloudfront.net/content/*'
```

exodus_gw/routers/cdn.py
```diff
@@ -90,11 +90,24 @@ def sign_url(url: str, timeout: int, env: Environment):
...
+
+    origin_url = os.path.join(env.cdn_url, "/origin/*")
+    origin_policy = build_policy(origin_url, expiration)
+    origin_signature = rsa_signer(env.cdn_private_key, origin_policy)
+
     params = [
         "Expires=%s" % int(datetime2timestamp(expiration)),
-        "Signature=%s" % cf_b64(signature).decode("utf8"),
+        "Signature=%s" % cf_b64(dest_signature).decode("utf8"),
+        "Content-Signature=%s" % cf_b64(content_signature).decode("utf-8"),
```
rohanpm, line 107: While this approach should work, I would like to suggest that we wrap up all the info in a single parameter.

The reason is that, right now it looks like you'll have to update exodus-cdn to look explicitly at Content-* and Origin-* parameters, which means the detail of "we generate cookies for /content/ and /origin/" is spread across two places and they'll both have to change if we want different logic at some point.

I think it would be a good idea to instead design this as a list and have exodus-lambda apply whatever's in the list, that way if we ever want to change the setup we only have to change what we put in this list and everything will work out.

Concretely I mean something like this:

```python
# policies to be inserted into cookies.
cookie_policies = [
  # note: path should NOT have the trailing "*" here because this value
  # needs to go into "Path" attribute on the cookie which does not use wildcards.
  {"path": "/content/", "policy": content_policy, "signature": content_signature},
  {"path": "/origin/", "policy": origin_policy, "signature": origin_signature},
]

params = [
    ...,
    "CookiePolicies=%s" % cf_b64(json.dumps(cookie_policies)).decode("utf-8"),
]
```

And then exodus-lambda does the inverse (base64 decode, json load) to get the list back, and the hardcoded two calls to signer.cookies_for_policy in exodus-lambda are replaced with iterating over this list.

rohanpm (changes requested):

exodus_gw/routers/cdn.py
negillett: Oh, interesting. Thanks!

negillett:
@rohanpm, @crungehottman, I decided to take this a step further and orient all the data as expected by set-cookie in exodus-lambda. Now each list item can be inserted directly into the set-cookie dict once decoded.

negillett:
Note that this brings the querystring length just over 2800 characters. We'll need to increase the accepted length limit in exodus-lambda to accommodate (3000?).

exodus_gw/routers/cdn.py
```diff
@@ -92,9 +92,31 @@ def sign_url(url: str, timeout: int, env: Environment):
...
+        cookies.extend(
+            [
+                "CloudFront-Key-Pair-Id=%s%s"
+                % (env.cdn_key_id, cookie_headers),
+                "CloudFront-Policy=%s%s" % (cookie_policy, cookie_headers),
+                "CloudFront-Signature=%s%s"
+                % (cookie_signature, cookie_headers),
+            ]
+        )
+    cookies_bytes = bytes(str(cookies), "utf-8")
```
rohanpm, line 114: You used default python stringification here rather than e.g. a json dump.  How do you propose to parse it on the other end?

Did you check the value that this generates?  It seems wrong in a couple of ways.  Here's what the value produced by the test looks like (some whitespace added):

```
['CloudFront-Key-Pair-Id=XXXXXXXXXXXXXX; Secure; HttpOnly; SameSite=lax; Domain=http://localhost:8049/_/cookie; Path=/content/; Max-Age=1800',
'CloudFront-Policy=b\'{"Statement":[{"Resource":"http://localhost:8049/content/*","Condition":{"DateLessThan":{"AWS:EpochTime":1644971400}}}]}\'; Secure; HttpOnly; SameSite=lax;
Domain=http://localhost:8049/_/cookie; Path=/content/; Max-Age=1800',
'CloudFront-Signature=b"G\\xa9>\\x04k\\xcd...
```

Problems:

- it's not of any particular format, not even round-trippable python which could be passed to "eval".
- the CloudFront-Signature part seems like it contains a stringification of the raw bytes rather than having them encoded in the required way.

tests/routers/test_cdn.py
```diff
@@ -17,11 +17,50 @@ def test_cdn_redirect_(monkeypatch, dummy_private_key):
...
-        "Signature=QXdMBQNyDLYeIsJzV7bKHnqYQSErcz9OYdJTuIYKVHCDaDiqP"
-        "OUjqkSXX4fm7A-Fi2roZSlWhyd4emrlC8hvNdPLZb3-7LHMVqau1QK9qFlh"
-        "Zz~aP1i4~Zud-kTot4JO4ewE8LdCkQL1pda-on~wVTXhiAtB7EaX8aR3dnB"
-        "ZmYo_&Key-Pair-Id=XXXXXXXXXXXXXX"
+        "http://localhost:8049/_/cookie/some/url?"
+        "Expires=1644971400&"
+        "Signature=QXdMBQNyDLYeIsJzV7bKHnqYQSErcz9OYdJTuIYKVHCDaDiqPOUjqkSXX4f"
+        "m7A-Fi2roZSlWhyd4emrlC8hvNdPLZb3-7LHMVqau1QK9qFlhZz~aP1i4~Zud-kTot4JO"
+        "4ewE8LdCkQL1pda-on~wVTXhiAtB7EaX8aR3dnBZmYo_&"
+        "Cookie-Policies=WydDbG91ZEZyb250LUtleS1QYWlyLUlkPVhYWFhYWFhYWFhYWFhYO"
```
rohanpm, line 25: In order to find that the encoded value of Cookie-Policies doesn't make sense, I copy-pasted this into a text file, cleaned it up and ran it through `base64 -d` to check it.

It's not reasonable to expect reviewers to do that, so can you please add a test which somehow covers the newly added code *before* the output value gets encoded?  For example, put that new code in a standalone function and add a new test which calls it.

rohanpm (changes requested):

rohanpm:
> Note that this brings the querystring length just over 2800 characters. We'll need to increase the accepted length limit in exodus-lambda to accommodate (2000->3000?).

I suggest making that change immediately so that all exodus-cdn environments will already be capable of accepting the longer query string before we submit this into exodus-gw.

exodus_gw/routers/cdn.py
negillett: Ah, overlooked the policy when dropping json and forgot to bring over the b64 decodes. Thanks.

exodus-gw.ini
```diff
@@ -4,7 +4,7 @@ bucket = my-bucket
 table = my-table
 config_table = my-config
 
-cdn_url = http://localhost:8049/_/cookie
+cdn_url = http://localhost:8049
```
rohanpm, line 7: Why was this changed?

exodus_gw/routers/cdn.py
```diff
@@ -90,11 +115,23 @@ def sign_url(url: str, timeout: int, env: Environment):
...
+    policy = build_policy(dest_url, expires)
     signature = rsa_signer(env.cdn_private_key, policy)
+
+    cookies = []
+    for path in ("/content/*", "/origin/*"):
+        cookie_url = urljoin(env.cdn_url, path)
+        cookie = cf_cookie(
+            cookie_url, env.cdn_private_key, env.cdn_key_id, expires, username
+        )
+        append = f"; Secure; HttpOnly; SameSite=lax; Domain={env.cdn_url}; Path={path}; Max-Age={timeout}"
```
rohanpm, line 127: Maybe this is why the cdn_url setting was changed, so that `Domain={env.cdn_url}` would work?

I don't think that makes sense though, because we still need the redirect endpoint to redirect to `/_/cookie` specifically and not `/`, and it still preferably remains configurable as it is today.

I think you just need to extract the correct part out of `env.cdn_url` and not expect it to already be the needed value.

rohanpm (changes requested):

exodus-gw.ini
negillett: I mistook it for an error/something set for testing. From a quick check of real configs, it looks like this is the correct/expected path and it's the other envs that are incorrect.

rohanpm (approved):

PR #611, opened 2023-10-19
Upgrade runtime.txt to python 3.11 [RHELDST-20628]

dichn:


crungehottman (approved):

rohanpm:
In case anyone else doesn't remember what this is for.  The point is that dependabot can determine the version of python used by looking at this file.

rohanpm (approved):

PR #612, opened 2023-10-20
Signatures in redirect requests: minor fixes [RHELDST-18071]

negillett:
This commit includes some small fixes to the recent addition of
CloudFront signatures in redirect requests.

- Path cookie header: wildcard removed
- Domain cookie header: URL scheme removed
- Set-Cookies querystring param renamed to avoid confusion
- Adds separate cookie ttl setting


exodus_gw/routers/cdn.py
```diff
@@ -132,7 +131,7 @@ def sign_url(url: str, timeout: int, env: Environment, username: str):
     params = [
         "Expires=%s" % int(datetime2timestamp(expires)),
         "Signature=%s" % cf_b64(signature).decode("utf8"),
-        "Set-Cookies=%s" % cf_b64(cookies_bytes).decode("utf-8"),
+        "Set-Cookie=%s" % cf_b64(cookies_bytes).decode("utf-8"),
```
rohanpm, line 134: I'm not that keen on this particular change, due to the fact that Set-Cookie is the name of the standard HTTP response header for setting cookies - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie .

It'll work technically fine as nothing prevents using a standard header name as a query string parameter, but I can imagine people getting confused by this.  e.g. if you look at some exodus-gw/exodus-cdn request logs, Set-Cookie is going to appear in two different contexts, once as this parameter and once in response headers, and someone will probably jump to the conclusion that they are "the same thing" and not understand how things are actually working.

rohanpm (approved):

exodus_gw/routers/cdn.py
negillett: It should probably be something else entirely then. I think Set-Cookies is equally confusing, though technically different. CloudFront-Cookies or something.

rohanpm (approved):

rohanpm (changes requested):
Found one more issue:

This change seems like it effectively reduces the timeouts on the cookies a lot.

There were previously two relevant timeouts:

1. Timeout applied by exodus-gw on the signed URL pointing at CDN

This should be a short timeout because the URL is expected to be fetched immediately, and because URLs can be leaked easily

2. Timeout applied by exodus-lambda /_/cookie endpoint when setting cookies

This should be a much longer timeout, because cookies are much less susceptible to leaking and because the idea is to let the users access cloudfront in their browser for a long-term debugging session (e.g. throughout the working day) without having to jump back to exodus-gw for a new cookie every few minutes

I think that when you moved all the logic for generating the cookies from exodus-lambda into exodus-gw, you've made it so that both of these are now applying timeout (1).  So the cookies expire much faster than they should.

Can you check that please?  Have a look at what the cookie TTLs are set to on exodus-cdn today.  We want the overall behavior to remain the same.

negillett:
> Found one more issue:
> ...
> Can you check that please? Have a look at what the cookie TTLs are set to on exodus-cdn today. We want the overall behavior to remain the same.

@rohanpm this actually increases the cookie expiration. In exodus-cdn it's set to 720. Here it does share the endpoint expiration, which is 1800. We can add a separate setting for each set though if it seems necessary.

rohanpm:
> > Found one more issue:
> > ...
> > Can you check that please? Have a look at what the cookie TTLs are set to on exodus-cdn today. We want the overall behavior to remain the same.
> 
> @rohanpm this actually increases the cookie expiration. In exodus-cdn it's set to 720. Here it does share the endpoint expiration, which is 1800. We can add a separate setting for each set though if it seems necessary.

I don't think that's right.  It can be shown by a simple test:

```
$ curl -Lv -b '' https://exodus-gw.corp.redhat.com/live/cdn/content/cookie-ttl-test 2>&1 | grep set-cookie | head -n1
< set-cookie: CloudFront-Key-Pair-Id=K3F04O0UVZ3KAN; Secure; HttpOnly; SameSite=lax; Domain=dgivg1yai1738.cloudfront.net; Path=/content/; Max-Age=43200
```

The current Max-Age is 43200, i.e. 12 hours.

Note that the `cookie_ttl` setting to exodus-lambda uses minutes and not seconds.

exodus_gw/routers/cdn.py
```diff
@@ -113,26 +118,25 @@ def sign_url(url: str, timeout: int, env: Environment, username: str):
...
         )
         cookies.extend([f"{k}={v}{append}" for k, v in cookie.items()])
     cookies_bytes = bytes(json.dumps(cookies), "utf-8")
 
     params = [
-        "Expires=%s" % int(datetime2timestamp(expires)),
+        "Expires=%s" % int(datetime2timestamp(signature_expires)),
         "Signature=%s" % cf_b64(signature).decode("utf8"),
-        "Set-Cookies=%s" % cf_b64(cookies_bytes).decode("utf-8"),
+        "CloudFront-Cookies=%s" % cf_b64(cookies_bytes).decode("utf-8"),
```
rohanpm, line 139: I've not noticed this on earlier reviews, but I don't think this is going to work because CloudFront-Cookies here is our own query string parameter, which means it needs to be included *inside* of the URL we're signing.  This logic instead adds it on afterward.

I'm reading https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-canned-policy.html#private-content-canned-policy-creating-signature and under "Values that you specify in the policy statement for a signed URL that uses a canned policy" it says that the base URL used to generate the signature must include our own query string parameters, so that would include CloudFront-Cookies.  

That means earlier when we do `build_policy(dest_url, signature_expires)`, we must ensure that CloudFront-Cookies was already incorporated into the `dest_url` by that point - if it's not then it's not covered by our signature.

rohanpm (changes requested):

exodus_gw/routers/cdn.py
negillett: Seems odd but okay. I wish they'd included more detail/reasoning. Like, are we doing this because parameters other than Key-Pair-Id, Signature, Expires, and Policy will be stripped by AWS? At any rate, it should only add a step to unpacking the cookies in Lambda.

negillett: @rohanpm, actually, if this is true then this effort is dead. We would need the private key in exodus-lambda to decrypt the signature and extract the cookies. So, we can either try it like it is or revert to generating cookies in exodus-lambda.

negillett: The only other way I see this working, if it doesn't already, is /if/ we can attach Policy along with Expires, Key-Pair-Id, and Signature. I've drafted this here: https://github.com/release-engineering/exodus-gw/compare/master...negillett:exodus-gw:18071-a

rohanpm: It's not odd, it'd be odd if it *didn't* work that way because that'd leave a major security gap.

Imagine for example that you have some CF-hosted app behind https://myapp.cloudfront.net and there is a page like https://myapp.cloudfront.net/ViewOrder.aspx?order_id=abc123 , and it's using signed URLs.

If the way it worked was that you signed the URL `https://myapp.cloudfront.net/ViewOrder.aspx`, and then added order_id onto the result, and CF only verified the  `https://myapp.cloudfront.net/ViewOrder.aspx` part of the URL while passing all query string parameters through, then what stops the user from taking the URL like `https://myapp.cloudfront.net/ViewOrder.aspx?Expires=1698268353&Signature=...&Key-Pair-Id=...&order_id=abc123` and changing the value of order_id in order to view the details of orders belonging to other users?

The way it works is instead that you need to sign `https://myapp.cloudfront.net/ViewOrder.aspx?order_id=abc123`, and that makes the result tamper-resistant, as if order_id or other elements of the query string are changed, the signature will no longer validate.

> if this is true then this effort is dead. We would need the private key in exodus-lambda to decrypt the signature and extract the cookies

Why?  This doesn't sound right at all.  Signing doesn't encrypt the URL.

For example, when we sign a URL like https://dgivg1yai1738.cloudfront.net/_/cookie/foo/bar, the lambda doesn't have to decrypt the /_/cookie/foo/bar part.  The query string should be no different.

> I'm not yet convinced the cookies need to be in policy url since they're not meant to go through to said url, they just need to make it through to the Lambda function

I don't know what you mean by "go through to said URL".  In my mind, if a request makes it into our lambda, it has "gone through".

Here's the sequence of events as I understand it.  I'm going to use the same example URL as at https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-canned-policy.html .

- client wants to access: `https://d111111abcdef8.cloudfront.net/images/horizon.jpg?size=large&license=yes`
- client generates "canned policy" enabling access to this URL, following exactly the structure documented at https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-canned-policy.html#private-content-canned-policy-creating-policy-statement
- client uses *private key* to generate signature for that policy and adds the 3 resulting parameters related to signing: `https://d111111abcdef8.cloudfront.net/images/horizon.jpg?size=large&license=yes&Expires=[...]&Signature=[...]&Key-Pair-Id=[...]`
- client passes URL to CF
- in CF, if signing is enabled, then CF:
   - strips the three signing parameters from the URL. This recovers the original URL of `https://d111111abcdef8.cloudfront.net/images/horizon.jpg?size=large&license=yes`.
   - internally generates a "canned policy", using this URL and the value of Expires, exactly in the structure shown in the docs.  Since the client and CF generated the policy in exactly the same way, they'll generate exactly the same policy (this is why it doesn't have to be passed in the URL).
   - Looks up the *public key* via Key-Pair-Id.
   - Uses the public key to verify that the Signature is a valid signature for the canned policy just generated.
- Assuming that all succeeds, and if lambda@edge is used, then...
- CF assembles an "event" in the structure documented at https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html and it'll include:
    - `uri`: `/images/horizon.jpg`
    - `querystring`: `size=large&license=yes`
- CF invokes the lambda

The CloudFront-Cookies query string parameter is no different from e.g. the `size` parameter in above example.  Don't get confused by the fact that it happens to contain some CloudFront signatures, CF itself is not going to care about that at all during any of the steps mentioned above, it's not even going to look at the value.

> So far I haven't been able to find anything that explicitly says this won't work

You can easily confirm it doesn't work like this...

Generate a signed URL:

```
$ curl -sv -b '' https://exodus-gw.corp.redhat.com/live/cdn/qs-test 2>&1 | grep location:
< location: https://dgivg1yai1738.cloudfront.net/_/cookie/qs-test?Expires=1698270062&Signature=gVLgD[...]&Key-Pair-Id=[...]
```

Verify that the signed URL can be requested:

```
$ curl -sI 'https://dgivg1yai1738.cloudfront.net/_/cookie/qs-test?Expires=1698270062&Signature=gVLgD[...]&Key-Pair-Id=[...]' | head -n1
HTTP/2 302 
```

Try adding something onto the query string and see that the request is rejected:

```
$ curl -sI 'https://dgivg1yai1738.cloudfront.net/_/cookie/qs-test?Expires=1698270062&Signature=gVLgD[...]&Key-Pair-Id=[...]&SomeKey=SomeVal' | head -n1
HTTP/2 403 
```

negillett: Thanks, @rohanpm. My problem stemmed from only adding the cookies to the policy url, not actually changing `dest_url`. I lead myself to believe that the cookies would be locked in the signature. I'll just have to make sure the querystring doesn't end up in the redirect URL from origin-request.

negillett:
Rebased

exodus_gw/routers/cdn.py
```diff
@@ -113,29 +118,30 @@ def sign_url(url: str, timeout: int, env: Environment, username: str):
...
 
     params = [
-        "Expires=%s" % int(datetime2timestamp(expires)),
+        "Expires=%s" % int(datetime2timestamp(signature_expires)),
         "Signature=%s" % cf_b64(signature).decode("utf8"),
-        "Set-Cookies=%s" % cf_b64(cookies_bytes).decode("utf-8"),
         "Key-Pair-Id=%s" % env.cdn_key_id,
     ]
-    separator = "&" if "?" in url else "?"
+    separator = "&" if "?" in dest_url else "?"
```
rohanpm, line 144: Minor - you know for sure now that dest_url always contains "?" by this point, so this could be just "&" every time now.

rohanpm (dismissed):

exodus_gw/routers/cdn.py
negillett: I considered that but didn't want it to be forgotten should the url change. I suppose that's not a great reason for leaving excess code.

negillett: Made the change. Also replaced remaining % str interpolation with f-string.

crungehottman (approved):

PR #615, opened 2023-10-27
CDN Redirect: policy resources must escape '?' [RHELDST-18071]

negillett:
According to https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html#private-content-custom-policy-statement

### DIFF
--- a/exodus_gw/routers/cdn.py
+++ b/exodus_gw/routers/cdn.py
@@ -30,7 +30,8 @@ router = APIRouter(tags=[openapi_tag["name"]])
 def build_policy(url: str, expiration: datetime):
     datelessthan = int(datetime2timestamp(expiration))
     condition = {"DateLessThan": {"AWS:EpochTime": datelessthan}}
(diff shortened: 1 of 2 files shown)


crungehottman (approved):

rohanpm:
I don't think this change was correct, as https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html#private-content-custom-policy-statement is talking about using custom policies (where you include Policy in the URL) and we're using canned policies here.  

The doc https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-canned-policy.html for canned policies doesn't say the same about escaping "?".

It makes sense that they would be different, because custom policies support wildcards and "?" is one of the characters used for that (a single-character wildcard) so it'd have to be escaped there.  But canned policies are always for one specific URL, so there is no need to escape wildcards.

PR #629, opened 2023-11-28
Fix link resolution not isolated to current publish [RHELDST-21893]

rohanpm:
Items using link_to should only be resolved to a proper object within the current publish. If resolution is not limited in this way, it's even possible for links set up on "pre" to leak over into "live" or vice-versa.

Although link resolution was always intended to work this way since initially scoped, the fact that it was buggy ever since implemented means that unfortunately other systems might have come to rely on it, and Pub/Pulp in the "hook disabled" case is known to rely on it. For that reason, a temporary escape hatch is added to allow continuing with the original buggy behavior if needed.

There is no specific test in this repo targeting this bug, but there is a system test in pipeline-test repo covering it.

### DIFF
--- a/exodus_gw/models/publish.py
+++ b/exodus_gw/models/publish.py
@@ -1,3 +1,4 @@
+import os
 import uuid
 from datetime import datetime
 from typing import Optional, Union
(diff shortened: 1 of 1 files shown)


rohanpm:
I've submitted this without waiting for review because the impact of this bug is more severe than originally thought, and it needs to be resolved urgently.

PR #640, opened 2023-12-05
Add PULP_MANIFEST.asc to default entry point patterns

rohanpm:
This file is a detached signature of PULP_MANIFEST. It needs to be treated as an entrypoint / as phase2 content, similarly to the existing repomd.xml.asc.

### DIFF
--- a/exodus_gw/settings.py
+++ b/exodus_gw/settings.py
@@ -160,6 +160,7 @@ class Settings(BaseSettings):
         "repomd.xml",
         "repomd.xml.asc",
         "PULP_MANIFEST",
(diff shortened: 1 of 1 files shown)


rohanpm:
There is some info about this in epic CLOUDDST-15902.

I don't know when this is intended to be used, probably not for a while, but it seems best to add it into settings now so it won't be forgotten.  It's harmless to have this in settings even if unused for the time being.

Also, some of the JIRAs mention PULP_MANIFEST.sig instead, but that's just because the JIRAs are poorly maintained.  This is the correct filename.

crungehottman (approved):

PR #659, opened 2024-01-08
Don't require content_type on linked source items [RHELDST-21894]

negillett:


tests/routers/test_publish.py
```diff
@@ -328,14 +328,14 @@ def test_update_publish_items_path_normalization(db, auth_header):
...
     ]
 
-    # Should have stored normalized web_uri and link_to paths
+    # Should have stored normalized web_uri and link_to paths (update now resolves links as well).
     assert item_dicts == [
-        {"web_uri": "/some/path", "object_key": "1" * 64, "link_to": ""},
         {
             "web_uri": "/link/to/some/path",
-            "object_key": "",
-            "link_to": "/some/path",
```
negillett, line 337: Not sure how this went undetected until now.
The recent inclusion of link resolution in update_publish_items should have caused this to fail.

rohanpm: I believe the absence of content_type means it would have been falling into this branch which tolerates unresolvable links during publish update:

```
                if partial:
                    # Unresolvable links are permitted currently.
                    continue
```

If this test had tried to commit the publish at the end, then I suppose it would have failed.

rohanpm (approved):

PR #695, opened 2024-04-17
Record uploader info in S3 metadata during upload [RHELDST-234411]

rajulkumar:
When an object is uploaded to S3, uploading entity(uploader) info i.e user or service account is recorded as "gw-uploader" in the object metadata. This provides an additional source for auditing uploaded content apart from logs.

### DIFF
--- a/exodus_gw/routers/upload.py
+++ b/exodus_gw/routers/upload.py
@@ -71,7 +71,15 @@ bucket.upload_file('/tmp/hello.txt',
 import logging
 import textwrap
 
-from fastapi import APIRouter, HTTPException, Path, Query, Request, Response
+from fastapi import (
+    APIRouter,
+    HTTPException,
+    Path,
+    Query,
+    Request,
+    Response,
+    Depends,
+)
 
 from .. import auth, deps
 from ..aws.client import S3ClientWrapper
@@ -126,6 +134,7 @@ async def multipart_upload(
         ),
     ),
     settings: Settings = deps.settings,
+    caller_name: str = Depends(auth.caller_name),
 ):
     """Create or complete a multi-part upload.
 
@@ -144,6 +153,8 @@ async def multipart_upload(
     if uploads == "" and uploadId is None:
         # Means a new upload is requested
         metadata = extract_request_metadata(request, settings)
+        # add uploader info in the metadata to track the object modifying entity
+        metadata["gw-uploader"] = caller_name
         return await create_multipart_upload(s3, env, key, metadata)
 
     if uploads is None and uploadId:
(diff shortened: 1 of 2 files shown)


negillett (dismissed):

negillett (approved):

exodus_gw/routers/upload.py
```diff
@@ -229,6 +234,17 @@ async def object_put(
     return Response(headers={"ETag": response["ETag"]})
 
 
+async def extract_uploader_info(request: Request):
+    # get uploader/request initiator info from the request header
+    context = await call_context(request)
+    if context.user.internalUsername:
```
rohanpm, line 240: I believe code like this has now been duplicated 3-4 times (search for "client.serviceAccountId").  Maybe you can consider adding a `@property` onto the `CallContext` class instead?

That would also help you with the coverage issue, as this is currently failing CI due to not all the paths here being tested.  If the code was de-duplicated it's more likely at least one existing test already covers all the paths.

exodus_gw/routers/upload.py
```diff
@@ -229,6 +234,17 @@ async def object_put(
     return Response(headers={"ETag": response["ETag"]})
 
 
+async def extract_uploader_info(request: Request):
+    # get uploader/request initiator info from the request header
+    context = await call_context(request)
```
rohanpm, line 239: Can you please not call `call_context` directly in this way but instead declare it as a dependency on the endpoint, as is done in other places?  This uses the dependency injection framework.

```
call_context: auth.CallContext = deps.call_context,
```

If you're wondering what's the point of that.  The point is that it keeps the code decoupled from whatever's needed to construct a CallContext.

For example, if you imagine the CallContext object needs to start looking at something other than `request` (e.g. maybe it needs to start querying the DB).  If you have used dependency injection, then absolutely nothing here needs to change.  On the other hand if you wrote the code the way you have it now, then that change would ripple throughout.  The call here to `call_context`, the signature and the usage of `extract_uploader_info`, and the signature of `multipart_upload` would all need to change to accept and pass a DB session object throughout.

By the way, there is already a `caller_name` dependency which might be good enough to directly put into `gw-uploader` and simplify all this code.

```
caller_name: str = Depends(caller_name)
```

rajulkumar: Thanks for the explanation. It helps to understand using DI here.
I have used the caller_name as suggested. I was a bit skeptical to use it previously as it added 'user' or 'serviceaccount' text before the actual caller name(so added the new method extract_uploader_info), but I believe this info will be helpful to identify caller types. Hence, used it as-is.

rohanpm (approved):

crungehottman (approved):

PR #697, opened 2024-04-18
dev env: refresh sidecar version

rohanpm:
Use a newer version of sidecar, the same as used in production. Set "use-default-authorities: false", which became necessary if using a development CA as done here.


rohanpm:
@rajulkumar this is supposed to fix the sidecar issue you were having.

crungehottman (approved):

negillett (approved):

PR #699, opened 2024-04-19
Enforce /origin/files invariant [RHELDST-23443]

crungehottman:
In order to prevent conflicts, ensure that all clients follow the correct usage of the /origin/files/sha256 structure.

When clients publish underneath the /origin/files/sha256/ path, it is enforced that:

(1) The path always matches this format:
/origin/files/sha256/(first two letters of sha256sum)/(full sha256sum)/(basename)

(2) Every object_key is either absent, or is equal to the full sha256sum present in the path.

When either rule is broken, the request is denied with a 400 response.

### DIFF
--- a/exodus_gw/routers/publish.py
+++ b/exodus_gw/routers/publish.py
@@ -278,6 +278,76 @@ def update_publish_items(
 
     db_publish.resolve_links(ln_items=resolvable)
 
+    # Enforce correct usage of the /origin/files directory layout.
+    #
+    # Paths published under /origin/files must always match the format:
+    # /origin/files/sha256/(first two characters of sha256sum)/(full sha256sum)/(basename)
+    #
+    # Additionally, every object_key must either be "absent" or equal to the full sha256sum
+    # present in the web_uri.
+    origin_base_regex = re.compile("^(/content)?/origin/files/sha256/.*$")
+    origin_pattern = (
+        "^(/content)?/origin/files/sha256/[0-f]{2}/[0-f]{64}/[^/]{1,300}$"
+    )
+    origin_regex = re.compile(origin_pattern)
+    origin_items = [i for i in items if re.match(origin_base_regex, i.web_uri)]
+    for i in origin_items:
+        # All content under /origin/files/sha256 must match the regex
+        if not re.match(origin_regex, i.web_uri):
+            LOG.error(
+                "Origin path %s does not match regex %s",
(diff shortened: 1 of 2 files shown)


exodus_gw/routers/publish.py
```diff
@@ -278,6 +278,80 @@ def update_publish_items(
...
+    # present in the web_uri.
+    origin_regex = (
+        "^(/content)?/origin/files/sha256/[0-f]{2}/[0-f]{64}/[^/]{1,300}$"
+    )
+    origin_pattern = re.compile(origin_regex)
+    origin_items = [
+        i
+        for i in items
+        if re.match(
+            re.compile("^(/content)?/origin/files/sha256/.*$"), i.web_uri
```
rohanpm, line 296: This looks a bit weird: regular expressions don't usually need to be compiled.  They can be compiled to make the code more efficient if the pattern will be used many times.  However, this is going to compile the regex at every iteration of the loop over items, which could be 10,000 times.

If this is going to be compiled it should probably be at the same level as origin_regex, e.g. you could have `origin_base_regex`, `origin_regex`.

exodus_gw/routers/publish.py
```diff
@@ -278,6 +278,80 @@ def update_publish_items(
...
+                origin_regex,
+                extra={
+                    "publish_id": publish_id,
+                    "event": "publish",
+                    "success": False,
+                },
+            )
+            raise HTTPException(
+                400,
+                detail="Invalid path format for '%s' (unmatched regex)"
```
rohanpm, line 314: I think it would be nice if this message gave the user a hint that this is specifically about origin/files so they have something to search for.  Just "invalid path format ... unmatched regex" seems very difficult to find more information about, and in fact the user might assume it's an exodus-gw bug.  If it points towards origin/files, at least the caller can search for how other code deals with /origin/files and might realize that they are doing something which doesn't follow established conventions.

In fact in each of these branches I feel like the message you've got in the logs is a better user-facing message than the one you've put into the exceptions right now.

rohanpm (changes requested):

exodus_gw/routers/publish.py
rohanpm: ... or moving both of them up to global scope so they are compiled once at import time.

rohanpm (approved):

negillett (approved):

PR #714, opened 2024-05-22
Fix some confusing logs around cache flush

rohanpm:
exodus-gw would log "fastpurge is not enabled" even in cases where fastpurge *was* enabled, but the publish object had no URLs relevant for flush. Many publishes created by exodus-migrator fall into this category.

Move the conditionals around a bit so that the log messages are more factual in all cases.


negillett:
@crungehottman, should we drop or implement this?

PR #730, opened 2024-08-06
Add config endpoint [RHELDST-25037]

rbikar:
(1)
Added new endpoint `/{env}/config and a new router
for GET method. This new endpoint retrieves all available
CDN configuration and returns it in JSON response to client.

(2)
In order to consolidate endpoints that manipulate CDN configuration
the previously used endpoint /deploy-config is moved to /config endpoint.

The old endpoint is defined as deprecated and calls the `config_post()`
directly providing the same fuctionality as before.

The new endpoint provides the same functionality.


negillett (approved):

crungehottman (dismissed):

exodus_gw/routers/deploy.py
```diff
@@ -152,44 +77,17 @@ def deploy_config(
...
-        state="NOT_STARTED",
-    )
-    db.add(task)
+    This endpoint is deprecated, use /{env}/config instead.
+    """
+    path_params = {"env": env.name}
+    url = request.url_for("config_post", **path_params)
 
-    return task
+    return RedirectResponse(url=url)
```
rohanpm, line 93: I don't particularly like the idea of doing a redirect for this deprecation.
Is it possible to just call the new method instead, e.g. `return config_post(...)` ?

The thing about 3xx redirects is that it's not guaranteed that all clients will follow them.  HTTP clients usually offer the choice to follow redirects or not, and some clients also implement behaviors like "only follow redirects for GET" or "if we get a redirect response to a POST, change the method to GET".

So: changing the URL of an API accessed via POST, and giving a redirect response, is not necessarily going to be backwards-compatible for all clients.  It'd be safer if the request was just processed normally, at both the old and new URL.

rbikar: done

rbikar:
Now the old endpoint only calls the function used for new endpoint with any redirection in place.
ptal.

negillett (approved):

crungehottman (approved):

rohanpm (approved):

PR #772, opened 2025-01-20
Get config endpoint returns valid empty config [RHELDST-28276]

crungehottman:
Previously, exodus-gw would fail with a 500 error when attempting to use the "get config" endpoint (i.e., GET /{env}/config) against an environment with an empty config table.

Now, when attempting to use the "get config" endpoint against an environment with an empty config table, exodus-gw returns a default object representing a valid empty config.

```
$ curl -s --cert ~/certs/$USER.crt --key ~/certs/$USER.key \
     https://localhost:8010/test/config | jq
{
  "listing": {
    "": {
      "var": "releasever",
      "values": [
        ""
      ]
    }
  },
  "origin_alias": [
    {
      "src": "",
      "dest": "",
      "exclude_paths": []
    }
  ],
  "releasever_alias": [
    {
      "src": "",
      "dest": "",
      "exclude_paths": []
    }
  ],
  "rhui_alias": [
    {
      "src": "",
      "dest": "",
      "exclude_paths": []
    }
  ]
}
```

### DIFF
--- a/exodus_gw/aws/dynamodb.py
+++ b/exodus_gw/aws/dynamodb.py
@@ -169,10 +169,16 @@ class DynamoDB:
 
     def query_definitions(self) -> dict[str, Any]:
         """Query the definitions in the config_table. If definitions are found, return them. Otherwise,
-        return an empty dictionary."""
+        return a valid empty configuration."""
 
-        # Return an empty dict if a query result is not found
-        out: dict[str, Any] = {}
+        # If a query result is not found, return a reasonable default object representing a valid
+        # empty configuration.
+        out: dict[str, Any] = {
+            "listing": {},
+            "origin_alias": [],
+            "releasever_alias": [],
+            "rhui_alias": [],
+        }
 
         LOG.info(
             "Loading exodus-config as at %s.",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -67,6 +67,32 @@ def fake_dynamodb_query(
     return side_effect
 
 
+@pytest.fixture(params=["binary-config", "text-config"])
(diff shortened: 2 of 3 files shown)


exodus_gw/aws/dynamodb.py
```diff
@@ -169,10 +169,16 @@ def aliases_for_config_update(self) -> list[tuple[str, str, list[str]]]:
...
-        return an empty dictionary."""
-
-        # Return an empty dict if a query result is not found
-        out: dict[str, Any] = {}
+        return a valid empty configuration."""
+
+        # If a query result is not found, return a reasonable default object representing a valid
+        # empty configuration.
+        out: dict[str, Any] = {
+            "listing": {"": {"values": [""], "var": "releasever"}},
```
rohanpm, line 177: I'm a bit confused about the inclusion of empty strings here.
Do those *have* to be included to satisfy a schema?  If so we should probably fix the schema.

I rather expected empty config to mean empty objects and lists, like `"listing": {}, "origin_alias": [], `...

Especially for aliases, I would be concerned that empty strings in src/dest are not going to be interpreted as "empty config".  For example, checking if a path `startswith` src may be a reasonable way to check if a path is affected by an alias, but every single path starts with an empty string.

I would also worry that, if this is the default response for a clean environment, then any workflow along these lines:

- install exodus-gw
- GET current config
- add elements into the config and POST it back

... would permanently keep these elements in the config.

crungehottman: Good points, thanks!

rohanpm (approved):

PR #780, opened 2025-02-07
Dev env updates for Alpine-based db image [RHELDST-29655]

crungehottman:
The postgres image used in the dev env (quay.io/exodus/postgres:12) is an Alpine-based image. The image tag has been updated to clarify that the exodus-gw-db image used in the dev env is Alpine-based (quay.io/exodus/postgres:12-alpine).

Using the Alpine-based image without any changes to the dev env configuration produces the following error:

`FATAL:  private key file "/var/lib/postgresql/db-service-key.pem" must be owned by the database user or root`

Previously, we used a Debian-based postgres image wherein the standard postgres UID was 999. In Alpine-based images, the standard postgres UID is 70. The dev env configuration has been updated to reflect the UID discrepancy.

### DIFF
--- a/scripts/systemd/exodus-gw-cert.service
+++ b/scripts/systemd/exodus-gw-cert.service
@@ -26,9 +26,11 @@ ExecStartPost=cp \
   %E/exodus-gw-dev/db-service-key.pem
 
 # private key file must be owned by the database user or root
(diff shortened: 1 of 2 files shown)


lebouillon:
LGTM

PR #787, opened 2025-03-07
Lock pydantic version to fix recurring conflicts

negillett:
pydantic conflicts are repeatedly breaking CI checks, so lock version to latest


negillett:
@release-engineering/exodus ptal

PR #794, opened 2025-04-14
Fixes

negillett:
A couple commits here for maintenance.

Don't squash.


negillett:
note: please don't squash

PR #796, opened 2025-04-16
multipart uploads: abort if object already exists [RHELDST-30644]

negillett:
To help prevent duplicate upload operations, this commit adds a check
for an s3 object's existence before attempting an upload. If the object
does exist, it aborts the current multipart upload and returns a
response with data from the existing object.

### DIFF
--- a/tests/routers/upload/test_upload.py
+++ b/tests/routers/upload/test_upload.py
@@ -7,6 +7,19 @@ from exodus_gw.main import app
 
 TEST_KEY = "b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c"
 
+# Uploads will first check object's existence with a get.
+# Use this error to simulate non-existence, allow test uploads.
+NoSuchKey = ClientError(
+    {
+        "Error": {
+            "Code": "NoSuchKey",
+            "Message": "The specified key does not exist.",
+        },
+        "ResponseMetadata": {"HTTPStatusCode": 404},
+    },
+    "GetObject",
+)
+
 
 async def test_full_upload(
     mock_aws_client, mock_request_reader, auth_header, monkeypatch
@@ -60,6 +73,7 @@ async def test_part_upload(mock_aws_client, mock_request_reader, auth_header):
 
     mock_request_reader.return_value = b"best bytes"
     mock_aws_client.upload_part.return_value = {"ETag": "a1b2c3"}
+    mock_aws_client.head_object.side_effect = NoSuchKey
 
     with TestClient(app) as client:
         r = client.put(
@@ -76,6 +90,39 @@ async def test_part_upload(mock_aws_client, mock_request_reader, auth_header):
     assert r.headers["x-request-id"]
 
     # It should have an empty body
+    assert r.content == b""
+
+
+async def test_part_upload_duplicate(mock_aws_client, auth_header, caplog):
+    """Calling multipart_put when the object has already been uploaded
+    aborts the mpu and returns a valid response."""
+
+    caplog.set_level(10, "s3")
+
+    mock_aws_client.head_object.return_value = {"ETag": "a1b2c3"}
+
+    with TestClient(app) as client:
+        r = client.put(
+            "/upload/test/%s?uploadId=my-upload&partNumber=3" % TEST_KEY,
+            headers=auth_header(roles=["test-blob-uploader"]),
+        )
+
+    # It should check HEAD, log message, and issue abort
+    mock_aws_client.head_object.assert_called()
+    assert (
+        f"canceling multipart upload my-upload, object {TEST_KEY} already exists"
+        in caplog.text
+    )
+    mock_aws_client.abort_multipart_upload.assert_called()
+    # It should not upload
+    mock_aws_client.upload_part.assert_not_called()
+    # It should not try to complete upload
+    mock_aws_client.complete_multipart_upload.assert_not_called()
+
+    assert r.status_code == 200
+    assert r.headers["etag"] == "a1b2c3"
+    assert r.headers["content-length"] == "0"
+    assert r.headers["x-request-id"]
     assert r.content == b""
 
 
--- a/exodus_gw/routers/upload.py
+++ b/exodus_gw/routers/upload.py
@@ -71,6 +71,7 @@ bucket.upload_file('/tmp/hello.txt',
 import logging
 import textwrap
 
+from botocore.exceptions import ClientError
 from fastapi import (
     APIRouter,
     Depends,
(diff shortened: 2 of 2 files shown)


negillett:
@crungehottman, could you take a look when you have a moment?

tests/routers/upload/test_upload.py
```diff
@@ -155,3 +172,35 @@ async def test_put_error(mock_aws_client, mock_request_reader, auth_header):
         "<RequestId>aabbccdd</RequestId>"
         "</Error>"
     )
+
+
+async def test_upload_duplicate(mock_aws_client, auth_header, caplog):
```
crungehottman, line 177: Could you attempt to create a test which specifically tests the acceptance criteria (i.e., "exodus-gw reduces duplicate multipart uploads by checking an item's existence **between parts, canceling if item is complete**"). (e.g., if provided a 4-part multipart upload, `upload_part` is called twice before bailing, `abort_multipart_upload` is called, etc.)

negillett: Hmm I'm not confident that it's possible to do a black box test like that for this.
AFAIK, to test the whole process we would have to chunk a file, call `create_multipart_upload`, then `upload` for each part, and then `complete_multipart_upload`, all independently, (aws-sdk-go via exodus-rsync wraps this process up for us [1]) and I'm not clear on how that would be more valuable.

Also, by "canceling" we shouldn't mean "abort". Calling `abort_multipart_upload` would delete the item from s3, which we don't want. Rather, by "cancel" we are just skipping superfluous uploads (`multipart_put`/`upload_part`). And that's for each part. There's no good way right now to skip all remaining parts as well--that would seem to require rolling our own wrapper or creating a database table to check off parts.

One thing I did kind of change from the Jira issue's ask is the scope of the check. Instead of covering just multipart_upload, I placed it in upload to cover single-part as well, there seemed no reason not to.
Another thing is the method used for the check--GET rather than HEAD. I'm not sure if that's really beneficial. `object_put` and `multipart_put` are **capable** of returning a response body, which wouldn't be included with a HEAD response, but it looks like they don't in practice.

If I got any of this wrong, please lmk! Totally willing to make changes if justified + possible.

[1] https://github.com/aws/aws-sdk-go/blob/main/service/s3/s3manager/upload.go#L280

crungehottman: > I'm not clear on how that would be more valuable.

It isn't possible to test this use case in pipeline-test or another integration test because extensive mocking would be required, so it would be nice to have some additional assurances (e.g., a test in exodus-gw) that bailing out of a multipart upload won't have unintended consequences (e.g., an incomplete upload, an error reported to a client).

> Another thing is the method used for the check--GET rather than HEAD

I do think we should use a HEAD request to check if the object exists. A HEAD request is more cost effective, faster, and uses less memory.

> Calling abort_multipart_upload would delete the item from s3, which we don't want

Hmm, does calling `abort_multipart_upload` delete the item from the s3 bucket? I didn't gather that from the documentation. From what I read, `abort_multipart_upload` only seems to prevent new `UploadParts` from being added to a temp object (associated with the UploadId); Note that during a multipart upload, an object is not written to the s3 bucket until `CompleteMultipartUpload` is called. 

https://docs.aws.amazon.com/AmazonS3/latest/API/API_AbortMultipartUpload.html
https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html

If we don't abort a partially complete upload, it would be possible to run `CompleteMultipartUpload` on an incomplete upload, overwriting a complete object with an incomplete object.

negillett: > I do think we should use a HEAD request to check if the object exists. A HEAD request is more cost effective, faster, and uses less memory.

Yeah. As long as a body isn't expected anywhere, that's definitely best.

>From what I read, abort_multipart_upload only seems to prevent new UploadParts from being added to a temp object (associated with the UploadId)

You're right. It deletes the temp object storage, not the obj from s3 associated with the uploadID. 
Giving it some more thought, I now /suspect/ what's happening looks something like this;
```
Thread 1: upload p3 of 5a1df3c for ID 27593
Thread 2: upload p2 of 5a1df3c for ID 98724
# Each rsync worker called create mpu for the same content
Thread 1: upload p4 of 5a1df3c for ID 27593
Thread 2: upload p3 of 5a1df3c for ID 98724
Thread 1: Complete ID 27593 -- S3 obj 5a1df3c created
Thread 2: upload p4 of 5a1df3c for ID 98724
Thread 2: Complete ID 98724 -- S3 obj 5a1df3c is overwritten
```
In this case, yeah, we should definitely be aborting to clean up those parts.
It should look like this;
```
[...]
Thread 1: Complete ID 27593 -- S3 obj 5a1df3c created
Thread 2: upload p4 of 5a1df3c for ID 98724, positive HEAD
Thread 2: Abort issued for ID 98724
```

> it would be nice to have some additional assurances (e.g., a test in exodus-gw) that bailing out of a multipart upload won't have unintended consequences (e.g., an incomplete upload, an error reported to a client).

Yeah, the thing about that though is all our AWS functionality is mocked in exodus-gw tests, so we'd be testing the results of a mock set in that same test. Right?
However, we can at least assert that the mocked AWS gets an AbortMultipartUpload call when multipart_put attempts to upload part of a completed object.

negillett: I do believe QE could safely test for edge cases here, though. One would need to install this, setup a race condition for a mpu, maybe feeding the same piece of large content through exodus-rsync, to QE AWS.

crungehottman: > Yeah, the thing about that though is all our AWS functionality is mocked in exodus-gw tests, so we'd be testing the results of a mock set in that same test. Right? Those unintended consequences would only be apparent when using real AWS. 

Well, some unintended consequences can be predicted based on information provided to us in the AWS and boto3 documentation. For example, if AWS doesn't receive an `AbortMultipartUpload` request, we know it is possible for an incomplete upload to occur, so we'd want to check that the mocked AWS receives a request to `AbortMultipartUpload`; if AWS receives a request to `CompleteMultipartUpload` after aborting the multipart upload, a client-side error will be thrown (propagating down to an error for exodus-gw users), so we'd want to check that the mocked AWS doesn't receive a `CompleteMultipartUpload` call. Mocking certainly won't help us identify unknown, unintended consequences, but it still allows us to reliably check for the known, unintended consequences.

> However, we can at least assert that the mocked AWS gets an AbortMultipartUpload call when multipart_put attempts to upload part of a completed object.

Thank you!

> I do believe QE could safely test for edge cases here, though. One would need to install this, setup a race condition for a mpu, maybe feeding the same piece of large content through exodus-rsync, to QE AWS.

Yes, +1, good point. QE can check the exodus-gw logs for `canceling multipart upload %s, object %s already exists` to confirm that the competing upload was actually cancelled.

tests/routers/upload/test_upload.py
```diff
@@ -79,6 +93,37 @@ async def test_part_upload(mock_aws_client, mock_request_reader, auth_header):
...
+
+    # It should check HEAD, log message, and issue abort
+    mock_aws_client.head_object.assert_called()
+    assert (
+        f"canceling multipart upload my-upload, object {TEST_KEY} already exists"
+        in caplog.text
+    )
+    mock_aws_client.abort_multipart_upload.assert_called()
+    # It should not upload
+    mock_aws_client.upload_part.assert_not_called()
```
crungehottman, line 118: Everything looks great, but I have one more (small) thing to request :sweat_smile: 

Please add a check to ensure that complete_multipart_upload is not called:

`mock_aws_client.complete_multipart_upload.assert_not_called()`

negillett: Ah, yeah, good call

PR #802, opened 2025-04-29
multipart uploads: make duplicate handling more robust [RHELDST-30644]

negillett:
It's possible that multiple processes will attempt to abort a multipart upload. In this case only the first will succeed in doing so and others will receive a NoSuchUpload error from s3. This commit handles this case by catching such errors. Additionally, the check-abort process was applied to the POST (create/complete mpu) endpoint to prevent attempts at completing aborted multipart uploads.

The head check and abort functionality was taken from multipart_put, split, and moved into two new internal functions, _already_uploaded and _ensure_aborted.


exodus_gw/routers/upload.py
```diff
@@ -101,6 +101,40 @@
...
+):
+    try:
+        response = await head(env, s3, key)
+        LOG.debug(
+            "s3 object already exists: %s", key, extra={"event": "upload"}
+        )
+        return response
+    except (
+        ClientError
+    ):  # head only capable of raising S3.Client.exceptions.NoSuchKey
```
crungehottman, line 117: In an attempt to future-proof this a bit, could you please explicitly except the `S3.Client.exceptions.NoSuchKey` exception? (e.g., https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/exceptions/NoSuchKey.html#S3.Client.exceptions.NoSuchKey)

exodus_gw/routers/upload.py
```diff
@@ -101,6 +101,40 @@
...
+async def _ensure_aborted(
+    uploadId: str,
+    env: Environment = deps.env,
+    s3: S3ClientWrapper = deps.s3_client,
+    key: str = Path(..., description="S3 object key"),
+):
+    try:
+        await abort_multipart_upload(env, s3, key, uploadId)
+    except (
+        ClientError
```
crungehottman, line 131: In an attempt to future-proof this a bit, could you please explicitly except the `S3.Client.exceptions.NoSuchUpload` exception? (e.g., https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/exceptions/NoSuchUpload.html)

crungehottman (changes requested):

exodus_gw/routers/upload.py
negillett: Good idea 👍🏻

exodus_gw/routers/upload.py
```diff
@@ -160,6 +195,12 @@ async def multipart_upload(
 
     if uploads is None and uploadId:
         # Given an existing upload to complete
+        if await _already_uploaded(env, s3, key):
+            # Attempt to abort the duplicate mpu
+            LOG.debug(
+                "duplicate multipart upload detected, attempting to abort"
+            )
+            return await _ensure_aborted(uploadId, env, s3, key)
```
crungehottman, line 203: `_ensure_aborted` returns an empty response, should we instead attempt to mimic the `complete_multipart_upload` response, to maintain backwards compatibility and to hide the "abort multipart upload if already uploaded" implementation detail from clients/users?

tests/routers/upload/test_manage_mpu.py
```diff
@@ -122,6 +132,53 @@ async def fake_body():
...
+        )
+
+    # It should have checked for object existence
+    mock_aws_client.head_object.assert_called()
+    # It should at least attempt to abort the mpu
+    mock_aws_client.abort_multipart_upload.assert_called()
+    # It should not have tried to complete the mpu
+    mock_aws_client.complete_multipart_upload.assert_not_called()
+    # It should succeed
+    assert response.status_code == 200
```
crungehottman, line 168: Could you please also verify the contents of the response (e.g., headers, body)? For backwards compatibility (and to hide this implementation detail from the client/user), I'd expect that the response would be the same as though the multipart upload completed successfully.

exodus_gw/routers/upload.py
negillett: There doesn't seem to be a way to do that, unfortunately, as `complete_multipart_upload` has unique metadata that doesn't seem available even from `get_object`, not that we'd really want to fetch a whole object anyway.
The intent here was just ensure a successful response, as this code path activating implies that a `complete_multipart_upload` response was already returned to the client.
I think an "abort response" is appropriate since that's what was announced as happening. However, if you prefer...we can do is return a _partial_ response _similar_ to that of `complete_multipart_upload`... My only concern is that this confuses/distracts clients/users from the proper `complete_multipart_upload` response which does hold all of the expected data.

tests/routers/upload/test_manage_mpu.py
negillett: Yeah, again, it's not possible to duplicate the response exactly. If you'd prefer, we can come up with another similar response to return instead; 
Rather than this:
```
xml_response(
        "CompleteMultipartUploadOutput",
        Location=response["Location"],
        Bucket=response["Bucket"],
        Key=response["Key"],
        ETag=response["ETag"],
    )
```
we could do something like this:
```
xml_response(
        "AbortedMultipartUploadOutput",
        Key=key,
        ETag=response.headers["ETag"],
    )
```
That doesn't hide anything though. And I'm not entirely sure why we should try to hide it. This probably should be a backwards-incompatible change anyway.
The end client/user will still see a proper response from the first `complete_multipart_upload`, and, in addition, may see "duplicate multipart upload detected, attempting to abort" and  "Abort" log messages. I worry that adding another xml response would just further confuse things, as it would be similar but still missing otherwise expected keys.

crungehottman: Good point, thanks for looking into it! OK, let's proceed with the backwards incompatible change.

crungehottman (approved):

PR #807, opened 2025-05-09
upload, multipart: abort if object already exists [RHELDST-30644]

negillett:
To help prevent duplicate upload operations, this commit adds a check for an s3 object's existence before attempting a multipart_put or complete_multipart_upload. If the object does already exist, an abort of the multipart upload is attempted to avoid overwriting the previous object and orphaned parts are cleaned up.

### DIFF
--- a/exodus_gw/routers/upload.py
+++ b/exodus_gw/routers/upload.py
@@ -71,6 +71,7 @@ bucket.upload_file('/tmp/hello.txt',
 import logging
 import textwrap
 
+from botocore.exceptions import ClientError
 from fastapi import (
     APIRouter,
     Depends,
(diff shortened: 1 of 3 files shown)


negillett:
From QE testing, it appears we can't expect AWS to produce the right error codes. Documentation shows that ["Error"]["Code"] should be something like "NoSuchKey", but what's seen in practice is the following;
```
[...] {'Error': {'Code': '404', 'Message': 'Not Found'}, 'ResponseMetadata': {'RequestId': 'C843D3[...]}
```
This commit has all the same good stuff but catches "404" code as well as the corresponding text version.

negillett:
@crungehottman, could you please take another look at this when you have the chance?

PR #898, opened 2026-04-06
Perform DBAPIError retry at the route level [RHALDST-37210]

mbooz-rh:
Remove DBAPIError from middleware, as call_next cannot be called after the receive stream closes upon error in starlette>=0.45.1. The retry mechanism was moved to the route level, since that is the lowest level calls can be retried at.


negillett (changes requested):
nice use of APIRoute, didn't know we could override all routes that way 👍🏻

negillett (dismissed):

crungehottman (changes requested):

exodus_gw/main.py
```diff
@@ -250,39 +245,61 @@ def new_db_session(engine):
...
+        async def retry_route_handler(request: Request):
+            max_tries = request.app.state.settings.db_session_max_tries
 
-        return response
+            @backoff.on_exception(
+                backoff.expo, DBAPIError, max_tries=max_tries
+            )
+            async def retry_wrapper():
+                try:
+                    return await original_route_handler(request)
```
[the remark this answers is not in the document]
mbooz-rh: request.state attributes can only be initiated in middleware or dependancies, so we can't eliminate the middleware entirely. But I've added logic to re-connect into the route handler.

negillett (approved):

exodus_gw/main.py
```diff
@@ -250,39 +245,73 @@ def new_db_session(engine):
...
+    response = await call_next(request)
+
+    if response.status_code >= 200 and response.status_code < 300:
+        await run_in_threadpool(request.state.db.commit)
+
+    broker.set_session(None)
+    await run_in_threadpool(request.state.db.close)
+    request.state.db = None
+
+    return response
```
crungehottman:
```suggestion
    try:
        response = await call_next(request)
        if response.status_code >= 200 and response.status_code < 300:
            await run_in_threadpool(request.state.db.commit)
    finally:
        # Check if RetryRoute has already cleaned up the session
        if request.state.db is not None:
            # If not, the session should be cleaned up.
            broker.set_session(None)
            await run_in_threadpool(request.state.db.close)
            request.state.db = None
    return response
```
The purpose of the `finally` block in the original code was to ensure that the cleanup was performed in all cases: successful requests, DBAPIError retries, and other exceptions (e.g., HTTP exceptions). Without it, any other raised exceptions would exit before the session cleanup could occur (potentially resulting in connection pool issues, memory leaks, db connection limits, db lock issues, etc).

crungehottman (changes requested):

negillett (dismissed):

exodus_gw/main.py
```diff
@@ -250,39 +245,75 @@ def new_db_session(engine):
 
 
 @app.middleware("http")
-def db_session(request: Request, call_next):
+async def db_session(request, call_next):
```
crungehottman:
```suggestion
async def db_session(request: Request, call_next):
```

Sorry, I noticed one more thing. It looks like the type hint was mistakenly removed. Let's keep the type hint to help catch errors / for devs using IDEs that leverage type hints.

crungehottman (changes requested):

exodus_gw/main.py
mbooz-rh: Had to add a couple of `# type: ignore` for `broker.set_session`

crungehottman: Looks good, thank you!

crungehottman (approved):

